PR #129, opened 2026-04-12
feat: NNG/mbedtls + async background handles

CGMossa:
## Summary

- Compile NNG 1.11.0 and mbedtls 3.6.5 from bundled C sources via R's Makevars
- Safe Rust wrappers for NNG sockets and messages (`feature = "nng"`)
- `#[miniextendr(background)]` proc-macro for non-blocking async execution (`feature = "async"`)

## NNG/mbedtls C Library Build

NNG and mbedtls source trees are bundled in `rpkg/src/nng/` and `rpkg/src/mbedtls/`. R's build system compiles them using per-directory pattern rules in `Makevars.in` with library-specific CPPFLAGS. The resulting static archives (`libnng.a`, `libmbedtls_all.a`) are passed to cargo via `-C link-arg=` in RUSTFLAGS for both the staticlib and cdylib builds.

Platform detection in `configure.ac` sets NNG defines for:
- POSIX/Darwin/Linux platform selection
- kqueue (macOS) / epoll (Linux) poll backends
- arc4random (macOS) / getrandom (Linux) entropy
- All protocols (bus, pair, pub/sub, push/pull, req/rep, surveyor/respondent)
- All transports (inproc, IPC, TCP, TLS, WebSocket, FDC)
- TLS engine (mbedtls) + HTTP + WebSocket supplemental

This follows the same approach as the nanonext R package but replaces cmake with R's native `$(CC)` compilation, keeping the build system simple and not requiring cmake at install time.

## NNG Safe Rust Wrappers

Feature-gated behind `nng = []` (no Cargo deps — NNG is linked by Makevars).

- **`NngSocket`**: RAII wrapper with `Drop` → `nng_close()`. Factory methods for all protocols (`req`, `rep`, `pair`, `pair1`, `pub_`, `sub`, `push`, `pull`, `bus`). Methods: `dial()`, `listen()`, `send()`, `recv()`, `send_msg()`, `recv_msg()`, `set_recv_timeout()`, `set_send_timeout()`.
- **`NngMsg`**: Owned message. `from_bytes()`, `body()`, `len()`, `append()`, `clear()`. `into_raw()` for send (NNG takes ownership, avoids double-free). RAII `Drop` → `nng_msg_free()`.
- **`NngError`/`NngResult`**: Error type wrapping `nng_strerror()`.
- **FFI**: `extern "C"` declarations for ~40 NNG C functions + `#[repr(C)]` structs.

Verified: PAIR echo + REQ/REP over `inproc://` transport.

## `#[miniextendr(background)]`

Feature-gated behind `async = []` (no deps — uses `std::thread` + `std::sync::mpsc`).

Annotate a function to run it on a background thread:

```rust
#[miniextendr(background)]
pub fn slow_compute(n: i32, scale: f64) -> f64 {
    std::thread::sleep(std::time::Duration::from_secs(n as u64));
    n as f64 * scale
}
```

R gets an async handle back immediately:

```r
handle <- slow_compute(2L, 3.14)
handle$is_resolved()  # FALSE (still computing)
Sys.sleep(3)
handle$is_resolved()  # TRUE
handle$value()        # 6.28
```

**Implementation:**
- `MxAsyncHandle`: `Arc<AtomicBool>` resolved flag + `mpsc::SyncChannel` one-shot receiver
- Proc-macro generates 3 C wrappers: `C_fn` (spawns thread, returns `ExternalPtr<MxAsyncHandle>`), `C_fn__is_resolved`, `C_fn__value` (blocks, downcasts, `IntoR`)
- R env-class wrapper with `$is_resolved()` and `$value()` methods
- `Result<T, E>`: Ok unwrapped in thread, Err propagated as R error
- Panics caught via `catch_unwind`, propagated as R errors

## Known Limitations

- **Thread-per-task**: Each `background` call spawns a new OS thread (fine for heavy work, wasteful for many small tasks)
- **No cancellation**: Background threads run to completion
- **No Shiny/promises integration** (v1): `$value()` blocks R — poll `$is_resolved()` for non-blocking
- **PUSH/PULL hangs**: Synchronous push/pull on inproc may hang if connection isn't established before send (needs NngAio)
- **No Windows NNG build**: Makevars only handles POSIX platforms
- **`configure.ac` dev mode**: Now writes `.cargo/config.toml` with vendor source replacement (instead of deleting it) to fix cargo workspace resolution in git worktrees

## Remaining Work (documented in `plans/nng-async-integration.md`)

1. **NNG Phase B**: `NngCtx` (context multiplexing), `NngAio` (async IO), TCP/IPC tests, TLS config
2. **mirai interop**: R serialization from Rust, mirai wire protocol, `MiraiClient`, `RustDaemon`
3. **NNG worker pool**: `#[miniextendr(async_dispatch)]` with serde/bincode, linkme function registry
4. **Async v2**: `as.promise.mx_async_handle` for Shiny, `print` method, thread pool backend

## Test plan

- [x] NNG version query from R
- [x] NNG message create/read roundtrip
- [x] NNG PAIR socket echo over inproc
- [x] NNG REQ/REP socket request-reply over inproc
- [x] `bg_add(3, 4)` → 7 (basic async)
- [x] `bg_greeting("World")` → "Hello, World!" (string return)
- [x] `bg_pi_approx(1000000)` → 3.141592 (heavy computation)
- [x] `bg_result_ok(5)` → 10 (Result::Ok path)
- [x] `bg_result_ok(-1)` → R error (Result::Err path)
- [x] `bg_panicking()` → R error (panic propagation)
- [ ] Formal `testthat` test suite
- [ ] `R CMD check` pass


CGMossa:
This is not a priority right now. We will punt it for when it becomes important to anyone

PR #196, opened 2026-04-17
fix(vendor-crates.R): reject git URLs in parse_tree_packages (#36)

### LINKED ISSUE #36: vendor-crates.R: parse_tree_packages treats git URLs as local paths; templates need CRAN rewriting for path deps
## Problem

`vendor-crates.R`'s `parse_tree_packages()` treats git source URLs as local filesystem paths. When a package uses `[patch.crates-io]` with git URLs (rather than `[patch."https://..."]` with local paths), `cargo tree --format {p}` outputs lines like:

```
miniextendr-api v0.1.0 (https://github.com/CGMossa/miniextendr#dcf727bc)
```

The parser's path-extraction regex matches any parenthesized string containing `/`, so it picks up the git URL as a "path". Downstream, `patch_packages_for_crate()` tries to run `cargo tree --manifest-path https://github.com/...` and fails.

### Reproduction

Any R package that depends on miniextendr via `[patch.crates-io]` with git URLs (the scaffolded default for non-monorepo packages) will hit this when running `vendor-crates.R pack`.

### Fix

Filter out URLs after path extraction in `parse_tree_packages()`:

```r
if (grepl("^https?://", path) || grepl("^git[+@]", path)) {
  path <- NA_character_
}
```

This is a one-line fix applied in dvs2 (a2-ai/dvs2). Should be backported to `rpkg/tools/vendor-crates.R` and the minirextendr templates.

---

## Missing: CRAN-mode Cargo.toml rewriting for monorepo path deps

The configure.ac templates (both rpkg and minirextendr scaffolded output) handle CRAN-mode rewriting for miniextendr's own git deps, but they have no mechanism for rewriting **user monorepo path deps** (e.g. `dvs = { path = "../../../dvs" }`).

When a user adds a monorepo crate dependency via `use_vendor_lib()` or manually, the CRAN build will fail because:

1. `cargo vendor` / `vendor-crates.R pack` correctly packages the local crate into `vendor/`
2. But configure.ac never rewrites `dvs = { path = "../../../dvs" }` to `dvs = { path = "../../vendor/dvs" }` for CRAN/offline builds
3. Cargo fails to find the crate at the monorepo-relative path

### What dvs2 needed to add to configure.ac (CRAN mode)

```sh
# Rewrite monorepo path dep
if test -d "$VENDOR_OUT/dvs"; then
  "$SED" 's|dvs = { path = "../../../dvs" }|dvs = { path = "../../vendor/dvs" }|' \
    src/rust/Cargo.toml > src/rust/Cargo.toml.tmp && \
    mv src/rust/Cargo.toml.tmp src/rust/Cargo.toml
fi
```

This is package-specific, but the scaffolding (`use_vendor_lib()`) should generate this rewriting block automatically. Currently it modifies Cargo.toml and adds a `vendor-lib-{crate}` AC_CONFIG_COMMANDS block, but the CRAN-mode rewriting in the main `cargo-vendor` block doesn't know about these crates.

### Suggestions

1. **Fix `parse_tree_packages()`** — filter git URLs from path extraction (one-liner, affects rpkg + templates)
2. **`use_vendor_lib()` should inject CRAN rewrite rules** into the `cargo-vendor` AC_CONFIG_COMMANDS block, not just a separate vendor-lib block
3. **Emit a configure-time warning** when Cargo.toml contains path deps that resolve outside the package root and no vendor rewrite is configured — this catches the "it builds in dev but silently breaks on CRAN" failure mode
4. **Consider a miniextendr-lint rule** that warns about unvendored path deps pointing above the R package directory

### Also: FORCE_VENDOR template uses raw `cargo vendor`

The `FORCE_VENDOR` path in configure.ac runs raw `cargo vendor` which cannot vendor path dependencies. It should delegate to `vendor-crates.R pack` instead:

```sh
# Before (broken for path deps):
$CARGO_CMD vendor --manifest-path src/rust/Cargo.toml "$VENDOR_OUT"

# After:
"$R_HOME/bin/Rscript" "$abs_rpkg_dir/tools/vendor-crates.R" \
  pack --path "$abs_rpkg_dir" 2>&1
```

This affects both rpkg's configure.ac and the minirextendr templates.

CGMossa:
Closes #36 (scope below).

## Summary

`cargo tree --format {p}` emits git and registry sources in the same `(…)` slot as local paths:

```
miniextendr-api v0.1.0 (https://github.com/CGMossa/miniextendr#dcf727bc)
serde v1.0.228 (git+ssh://git@gh.com/serde-rs/serde.git#abcdef)
```

`parse_tree_packages` was extracting the URL as a filesystem path, which then fed into `--manifest-path` downstream. The fix rejects anything matching `^https?://` or `^git[+@]` after the regex match, leaving `path = NA_character_` for those rows.

## Scope

The issue suggested applying the fix to `rpkg/tools/vendor-crates.R` and the minirextendr templates. Those files no longer exist — `cargo-revendor` (the standalone crate at `cargo-revendor/`) superseded the R-based vendor flow in the main package. The only remaining copies of `parse_tree_packages` live in the cross-package test fixtures (`tests/cross-package/{producer,consumer}.pkg/tools/vendor-crates.R`), so the fix lands there, synced between both.

The second half of #36 — CRAN-mode rewriting for monorepo path deps in `configure.ac` and `use_vendor_lib` injecting path rewrites — is a larger architectural change and should be tracked in its own issue/PR. I'll leave #36 open for that half; happy to adjust scope if preferred.

## Test plan

- [x] Hand-verified the regex with fixture lines (git https, git+ssh, git@, local path, no paren):

```
             name version              path
1 miniextendr-api   0.1.0              <NA>
2             dvs   0.1.0 /Users/me/dev/dvs
3           serde 1.0.228              <NA>
4           rayon  1.12.0              <NA>
5             foo   0.1.0              <NA>
```

- [x] No Rust code touched; clippy / fmt / rebuild / vendor not required.

### DIFF
--- a/tests/cross-package/consumer.pkg/tools/vendor-crates.R
+++ b/tests/cross-package/consumer.pkg/tools/vendor-crates.R
@@ -117,6 +117,13 @@ parse_tree_packages <- function(lines) {
     path <- NA_character_
     if (grepl(" \\([^()]*[/\\\\][^()]*\\)$", line)) {
       path <- sub("^.* \\(([^()]*[/\\\\][^()]*)\\)$", "\\1", line)
+      # cargo tree emits git/registry sources in the same (…) slot as local
+      # paths (e.g. "foo v0.1.0 (https://github.com/x/y#deadbeef)"). Those
+      # are not filesystem paths — reject them so downstream callers don't
+      # try to feed a URL to --manifest-path.
+      if (grepl("^https?://", path) || grepl("^git[+@]", path)) {
+        path <- NA_character_
+      }
     }
 
     data.frame(
--- a/tests/cross-package/producer.pkg/tools/vendor-crates.R
+++ b/tests/cross-package/producer.pkg/tools/vendor-crates.R
@@ -117,6 +117,13 @@ parse_tree_packages <- function(lines) {
     path <- NA_character_
     if (grepl(" \\([^()]*[/\\\\][^()]*\\)$", line)) {
       path <- sub("^.* \\(([^()]*[/\\\\][^()]*)\\)$", "\\1", line)
+      # cargo tree emits git/registry sources in the same (…) slot as local
+      # paths (e.g. "foo v0.1.0 (https://github.com/x/y#deadbeef)"). Those
+      # are not filesystem paths — reject them so downstream callers don't
+      # try to feed a URL to --manifest-path.
+      if (grepl("^https?://", path) || grepl("^git[+@]", path)) {
+        path <- NA_character_
+      }
     }
 
     data.frame(


CGMossa:
Closing — this fix is in dead code. `vendor-crates.R` was superseded by the `cargo-revendor` standalone crate in the main build flow; the only remaining copies live in `tests/cross-package/{producer,consumer}.pkg/tools/` as scaffolding fixtures that haven't been kept in sync with the current vendoring approach.

Patching the script is the wrong move. The correct action is to rescaffold both cross-package fixtures to use the current flow (cargo-revendor + the current `configure.ac` template). Filing a new issue to track that.

PR #199, opened 2026-04-17
fix(worker): cooperative shutdown via recv_timeout + atexit/R_unload (#103)

### LINKED ISSUE #103: Worker thread blocks on recv() — no clean shutdown on process exit
## Problem

The miniextendr worker thread (`worker-thread` feature) uses a blocking `mpsc::recv()` loop with the sender stored in a `OnceLock` static. When the R process exits, the static is never dropped, so the channel stays open and the thread blocks forever.

On Linux/macOS this doesn't cause problems because `ExitProcess` / `_exit` forcibly terminates all threads. On Windows under `system2()` pipe capture, it _can_ keep the process alive (though DataFusion's Tokio threads were the actual trigger for the R CMD check hang — see #101).

## Current state

The worker thread uses:
```rust
static JOB_TX: OnceLock<SyncSender<AnyJob>> = OnceLock::new();
// ...
while let Ok(job) = job_rx.recv() { job(); }
```

`OnceLock` has no `take()` method, so the sender can never be dropped to signal the thread.

## Approaches considered

- **`recv_timeout`** — thread exits eventually but sender is never dropped, so it loops on Timeout forever
- **`Mutex<Option<SyncSender>>` + atexit** — atexit handler drops sender, thread detects Disconnected. Works in theory but C `atexit()` doesn't reliably fire during R's exit on Windows
- **`reg.finalizer(onexit=TRUE)`** — R GC finalizers don't run during R CMD check test process exit
- **`.onUnload`** — not called when the test process exits (R doesn't unload packages on quit)

## Recommendation

Switch from `OnceLock<SyncSender>` to `Mutex<Option<SyncSender>>` and use `recv_timeout` in the worker loop. Register shutdown via multiple mechanisms (atexit + reg.finalizer + .onUnload) for defense-in-depth. This isn't urgent since DataFusion was the actual hang cause, but it's a latent issue.

## References

- PR #101 — investigation and DataFusion fix
- `reviews/windows-rcmdcheck-hang.md`

CGMossa:
Closes #103.

## Problem

The worker thread (`worker-thread` feature) blocks forever on `mpsc::recv()`. The static `JOB_TX: OnceLock<SyncSender<AnyJob>>` is never dropped, and `OnceLock` has no `take()` method, so `recv` never sees `Disconnected`. On process exit the worker stays pinned in a blocked syscall — harmless on Linux/macOS where `_exit` reaps threads, but a latent hang risk on Windows under `system2` pipe capture.

## Fix (defense-in-depth)

- Replace the blocking `while let Ok(job) = job_rx.recv()` loop with `recv_timeout(250 ms)` that polls a new `WORKER_SHOULD_STOP: AtomicBool`. Idle workers wake at most 4x/sec; in-flight jobs dispatch immediately.
- Add `miniextendr_runtime_shutdown()` — public, idempotent, a no-op without the `worker-thread` feature.
- `miniextendr_runtime_init` registers a libc `atexit` hook calling `miniextendr_runtime_shutdown`. atexit can be flaky on Windows but is additive — never hurts.
- `miniextendr_init!` now also generates `R_unload_<pkg>` which R calls on `detach(unload=TRUE)` / `dyn.unload()`.

## Test plan

- [x] `just clippy`
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (clippy_default)
- [x] `cargo clippy --workspace --all-targets --locked --features <strict set> -- -D warnings` (clippy_all)
- [x] `just fmt`
- [x] `just rcmdinstall` — builds & installs clean on worker-thread build
- [x] `just devtools-document` — `R_unload_miniextendr` doesn't surface as an R-level symbol, so NAMESPACE / wrappers unchanged from current
- [x] `just vendor` — `rpkg/inst/vendor.tar.xz` refreshed
- [x] `cargo test -p miniextendr-api --features worker-thread run_on_worker_reentry` — existing re-entry guard still green

## Notes

- The `recv_timeout` polling interval (250 ms) is the only new tuning knob — chosen so unloading a package is imperceptibly slow (~250 ms worst case) without the worker busy-polling the atomic flag when idle.
- We deliberately do NOT track a `JoinHandle` or call `thread::join`: the worker exits within ≤250 ms of the shutdown signal and the OS reaps the thread on process exit. Tracking a JoinHandle would add refcount machinery for marginal benefit.
- Capacity-0 rendezvous semantics on the job channel are unchanged, so in-flight job behavior is identical to before.


CGMossa:
We should consider depending on the rust crates ctor/dtor instead of relying on atexit. Althought this is a perfectly acceptable approach as is.

PR #279, opened 2026-04-19
fix(worker): synchronous shutdown + panic hook uninstall so DLL unload is safe on Windows (#277)

### LINKED ISSUE #277: Windows: panic runtime corrupted across DLL unload → 'failed to initiate panic, error 5'
# Windows: panic runtime corrupted across DLL unload → "failed to initiate panic, error 5"

## Symptom

On Windows-GNU (Rtools45 `x86_64-pc-windows-gnu`), `devtools::test("rpkg")`
aborts the R process partway through the suite with:

```
[Rust] Worker: about to panic
[Rust] Dropped: worker: boxed resource before panic
[Rust] Dropped: worker: resource before panic
fatal runtime error: failed to initiate panic, error 5, aborting
```

`error 5` is Win32 `ERROR_ACCESS_DENIED` surfacing from `RaiseException` —
Rust's panic runtime can't start unwinding, so `catch_unwind` /
`with_r_unwind_protect` never get to run and the process is aborted.

Destructors run before the abort ("Dropped: …" prints), so a panic *is*
being raised successfully; the failure is on the next `RaiseException`
attempt or a follow-up SEH operation.

## Why this is distinct from #190

PR #190 fixes a related but separate Windows issue: rustc emits
`-lgcc_eh -lgcc_s`, but Rtools45's `.static.posix` toolchain ships a
*unified* `libgcc.a` with no separate `libgcc_eh.a` / `libgcc_s.a`. The
PR's CI used empty `ar crs` archives for those names; with `ld.lld`
those empty archives linked fine but left `__gcc_personality_v0` /
`__gcc_personality_seh0` unresolved at runtime, so the *first* panic
aborted immediately. PR #190 (commit `0bfa85a4`) stages the real unified
`libgcc.a` under both names so the personality routines actually resolve.

With PR #190's libgcc fix in place, the first N panics work. This issue
is about what breaks later.

## Reproducer

Confirmed locally on Windows 11 + Rtools45 UCRT64 + rustc 1.94.1 +
R 4.5.3, with PR #190's libgcc fix applied:

- Fresh `library(miniextendr)` + repeat `.Call(C_worker_drop_on_panic)`
  4× in a row — all succeed, panic caught, converted to R error, clean.
- `Rscript -e 'for (f in files) testthat::test_file(f)'` across 40 test
  files including `test-errors-more.R` — all pass, no abort.
- `Rscript -e 'devtools::test("rpkg")'` — aborts with the error above
  when `unsafe_C_worker_drop_on_panic` runs (test-errors-more.R:46).

Differentiator: `devtools::test` uses `devtools::load_all(reset = TRUE)`
which triggers `unloadNamespace` → `library.dynam.unload` →
`R_unload_<pkg>` before the reload. A plain `library(miniextendr)`
never unloads.

## Root cause

Three pieces of the miniextendr runtime hold references into the DLL's
code that outlive `R_unload_<pkg>`:

1. **Worker thread.** `R_unload_<pkg>` calls
   `miniextendr_runtime_shutdown()` which calls
   `worker_channel::shutdown()` (`miniextendr-api/src/worker.rs:332`) —
   an atomic flag the worker polls every 250 ms in its `recv_timeout`
   loop. Shutdown is **signal-only, not join.** On Windows, R's
   `library.dynam.unload` unmaps the DLL's code pages synchronously
   once the unload returns. If the worker thread is still alive
   (mid-poll or about to wake), it resumes executing in freed memory,
   hits an access violation, and corrupts the process's SEH chain. The
   next `RaiseException` anywhere in the process fails.

2. **libc `atexit` handler.** `miniextendr_runtime_init`
   (`worker.rs:195-204`) registers an `atexit_shutdown()` function
   pointer whose address lies in the DLL. If the DLL is unloaded before
   libc's atexit registry fires (normal case for `dyn.unload`), process
   exit jumps to an unmapped address.

3. **Rust panic hook.** `miniextendr_panic_hook()` (`backtrace.rs:7-28`)
   captures a closure via `std::panic::set_hook`. The closure code
   lives in the DLL. After unload the global panic hook still points
   to it; any subsequent panic hook invocation jumps to unmapped
   memory. The project already knows this one — `init.rs:53-55` has an
   explicit *"on Windows, set_hook during DLL init can fail with
   'failed to initiate panic, error 5' because the panic infrastructure
   isn't fully available during DLL loading"* comment and skips
   `set_hook` in the wrapper-generation path.

## Suggested fix

All three hazards share the underlying shape *"DLL code outlives DLL
unload"* — fix them together. No timeouts, no polling, no sleeps.

### (1) Worker: tagged-message channel + blocking join

Replace the atomic-flag + `recv_timeout(250ms)` poll protocol with a
tagged message and a blocking `recv()`:

```rust
enum WorkerMsg {
    Job(AnyJob),
    Shutdown,
}
// JOB_TX: SyncSender<WorkerMsg> wrapped in a Mutex<Option<_>> so
// shutdown can take() it and drop after send, closing the channel.

fn worker_loop(rx: Receiver<WorkerMsg>) {
    while let Ok(msg) = rx.recv() {
        match msg {
            WorkerMsg::Job(job) => job(),
            WorkerMsg::Shutdown => break,
        }
    }
}

pub fn shutdown() {
    // idempotent; only runs the send/join on the first call
    let Some(tx) = JOB_TX.lock().unwrap().take() else { return };
    // Best effort — if the worker already exited the send errors, fine
    let _ = tx.send(WorkerMsg::Shutdown);
    drop(tx); // closing the channel is the other wake-up path
    if let Some(h) = WORKER_JOIN_HANDLE.lock().unwrap().take() {
        h.join().expect("worker thread panicked");
    }
}
```

Why it's rigid:

- No polling interval. No `sleep`. No `is_finished` race.
- `recv()` blocks until either a `Job`, a `Shutdown`, or
  sender-dropped — all three unblock the worker immediately.
- `h.join()` blocks exactly as long as the current in-flight job takes
  to drain plus one `match` — no jitter.
- `R_unload_<pkg>` does not return until the thread is genuinely gone,
  so `library.dynam.unload`'s subsequent `FreeLibrary` sees no live
  code references.
- Drop the `WORKER_SHOULD_STOP` atomic, the 250 ms
  `SHUTDOWN_POLL_INTERVAL`, and the `miniextendr_runtime_join_for_test`
  helper — the new design doesn't need them. The #204 test can assert
  via `miniextendr_runtime_shutdown()` directly; if it ever doesn't
  return, that's the bug to find.

Edge case the current design handles via the atomic and this one
handles via sender-ownership: callers of `dispatch_to_worker` after
shutdown. Now `JOB_TX.lock()` returns `None` and they return a
"worker shut down" error instead of `send` erroring — same outcome,
clearer error source.

### (2) atexit handler

Either:

- Drop the atexit registration entirely on Windows — `R_unload_<pkg>`
  is enough for every normal exit path, and `q("no")` etc. tear down
  the process without needing worker cleanup (the OS reaps the
  thread). The existing comment already notes *"atexit can be flaky on
  Windows"* as a caveat.
- Or register the handler in a host C file that's linked into libR,
  not the rpkg DLL — so the handler code outlives the DLL.

First option is smaller and aligns with the "rigid, no special cases"
spirit of the worker fix.

### (3) Panic hook

In `R_unload_<pkg>`, restore the previous panic hook so the global
hook no longer references DLL code:

```rust
// In miniextendr_panic_hook, save the current hook before replacing
// and expose an "uninstall" that reinstates it.
pub extern "C-unwind" fn miniextendr_panic_hook_uninstall() {
    let _ = std::panic::take_hook();
    // The default hook reinstated by take_hook is the process default,
    // not the closure we stored in DLL code — that's what we want.
}
```

Call it from `miniextendr_runtime_shutdown` after the worker has
joined (order matters: worker might itself panic as it exits and we
don't want the unmapped hook to fire).

## Non-fix: test-level skips

`test-errors-more.R:17` and `:25` already
`skip("thread panic propagation causes runtime errors in extern C-unwind")`
for two sibling tests. Adding one more skip for `worker_drop_on_panic`
on Windows would mask this specific failure but the underlying hazard
remains — any package user whose code panics after
`devtools::load_all(reset = TRUE)` hits it, and `atexit` / `set_hook`
remain time bombs at process exit. Skip is not enough; fix the
lifecycle.

## Scope

Separate from PR #190, which only addresses the libgcc personality-
symbol link-time issue. The libgcc fix is necessary on its own merit
(empty stubs → `_Unwind_RaiseException` can't find a personality even
on the first panic) and is unaffected by this issue's resolution.
Referencing #190 so the two can be read together.

CGMossa:
Closes #277.

## What's broken

On Windows, `R_unload_<pkg>` → `library.dynam.unload` unmaps the DLL's code pages synchronously once unload returns. Three pieces of the miniextendr runtime held references into that code and didn't drop them before the unmap, turning every reload cycle into a latent time bomb:

1. **Worker thread.** `miniextendr_runtime_shutdown` only set the `WORKER_SHOULD_STOP: AtomicBool` flag; the worker was asleep in `recv_timeout(250ms)` and would eventually wake — now inside freed memory — corrupting the process's SEH chain. `RaiseException` anywhere in the process then failed with `ERROR_ACCESS_DENIED` (5), which Rust's panic runtime surfaces as `fatal runtime error: failed to initiate panic, error 5, aborting`.
2. **Rust panic hook.** The closure installed by `miniextendr_panic_hook` lives in DLL code; `std::panic::set_hook` stored it in the process-global slot. Same hazard — the next panic would jump into unmapped memory.
3. **libc `atexit` handler.** Function pointer into DLL code, set up as "defense in depth" for `.onUnload`-missing exits. Same hazard at process shutdown. The original comment already called it "flaky on Windows".

`devtools::test` / `devtools::load_all(reset = TRUE)` hit (1) every run, which is why the test suite reliably aborts.

### Worker: tagged-message channel + blocking join

Replace the `AtomicBool` + `recv_timeout(250ms)` poll protocol with a proper `WorkerMsg::{Job, Shutdown}` channel. The worker uses blocking `recv()`; `shutdown()` sends `Shutdown`, drops the sender, then blocks on `JoinHandle::join()`. No timeout, no `sleep`, no polling interval. `R_unload_<pkg>` doesn't return until the thread is truly gone, so the DLL unmap sees no live code references.

`JOB_TX: OnceLock<SyncSender<_>>` becomes `WORKER: Mutex<Option<_>>` so shutdown can `.take()` both the sender and the join handle. `dispatch_to_worker` now returns a structured "worker not running" error after shutdown. `WORKER_SHOULD_STOP`, `SHUTDOWN_POLL_INTERVAL`, and `miniextendr_runtime_join_for_test` are gone.

### Panic hook: install/uninstall pair

`backtrace::miniextendr_panic_hook_uninstall()` added and called from `miniextendr_runtime_shutdown` after the worker has joined. Drops our closure via `std::panic::take_hook()` and resets the process slot to Rust's default. `INSTALLED: AtomicBool` replaces `std::sync::Once` so install/uninstall can pair within a DLL instance.

### atexit: removed

Dropped entirely. Normal path (package unload → `R_unload_<pkg>`) is covered. Abnormal exits (`q(\"no\")` etc.) rely on the OS reaping the worker, which it does.

## Testing

- `miniextendr-api/tests/worker_shutdown.rs` now asserts `miniextendr_runtime_shutdown()` itself returns in under 2s (no more separate `_join_for_test` helper). Passes locally in 1.4s.
- End-to-end Windows validation: `just rcmdinstall` completes cleanly (previously the wrapper-generation `dyn.unload` hung when I experimented with a polling-based join; the blocking join works first try). Full `devtools::test` validation is running locally and will be reported in a follow-up comment.

## CI

Expect Windows CI to fail on main's pre-existing vendoring bootstrap regression (cargo-revendor fails on `rpkg/src/rust/Cargo.toml`'s frozen path deps before vendor/ exists — the error was visible on PR #190's Windows job too, before it ever reached the panic path). Not caused by this PR. Merging this unblocks the real-test once that regression is fixed.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


CGMossa:
## Local validation — passes clean

End-to-end confirmation on Windows 11 + Rtools45 UCRT64 + rustc 1.94.1 + R 4.5.3:

```
[ FAIL 0 | WARN 0 | SKIP 39 | PASS 4403 ]
```

Zero `fatal runtime error` occurrences in the full `just devtools-test` run. The test that previously aborted the R process (`test-errors-more.R:46 → unsafe_C_worker_drop_on_panic`) is part of the 4403 passing count. The 39 skips are the pre-existing platform/feature gates (`skip_on_os("windows")`, `skip("feature not enabled: connections")`, the two explicit `skip("thread panic propagation causes runtime errors in extern C-unwind")` lines — those skips can be revisited in follow-up now that the underlying lifecycle hazard is fixed).

### What the run exercised

1. `just configure` → dev-monorepo vendor sync (cargo-revendor populates `rpkg/vendor/`).
2. `R CMD INSTALL rpkg` → staticlib + cdylib compile; `dyn.load(cdylib) → .Call(miniextendr_write_wrappers) → dyn.unload(cdylib)` for R-wrapper generation. **Critical path**: this is where the polling-based shutdown I tried earlier hung indefinitely — the new blocking-join shutdown completes first try.
3. `devtools::test("rpkg")` → `devtools::load_all(reset = TRUE)` (which unloads & reloads the package) → 94 test files, including the worker-panic tests.

Previously step 3 aborted the R process when the first `catch_unwind(panic!())` inside the worker ran. Now it returns a proper R error captured by `expect_error` like every other platform.

### Run time

`just devtools-test` took roughly 45 min end-to-end on my machine (most of that is cargo rebuild triggered by devtools::document; the test phase itself is ~5 min). That's unrelated to this PR — same as before.

### CI note

Windows / Linux / macOS CI jobs on this PR are failing for the same reason PRs #190 and #278 did: a pre-existing main-side vendoring bootstrap regression (cargo-revendor can't load `rpkg/src/rust/Cargo.toml`'s frozen vendor/ path deps when `vendor/` doesn't exist yet, and configure aborts before compilation). The jobs that don't run `just configure` (Rust lint, minirextendr, Version Check, Generated Files Check) pass. The worker fix in this PR is unaffected by that CI-side regression and will validate in CI once it lands.

PR #296, opened 2026-04-21
feat(minirextendr): add miniextendr_check_static() (no Rust compile)

CGMossa:
## Summary

Adds `minirextendr::miniextendr_check_static()` — a fast `R CMD check` that does **not** trigger a Rust compile. Targets the use case @CGMossa raised on the original PR thread: validating Rust-powered R packages **before** their dependencies are vendored, in sandboxes, or in any context where `cargo` cannot reach the network.

Wraps `rcmdcheck::rcmdcheck()` with `--install=fake`, which makes `R CMD INSTALL` set `use_configure <- FALSE` (`r-svn tools/R/install.R:2424`); `R CMD check` then auto-skips `do_examples` / `do_tests` / `do_vignettes` (`r-svn tools/R/check.R:7373`). No `./configure`, no `cargo`, no network.

### Catches
- DESCRIPTION fields, NAMESPACE consistency, R syntax
- Rd parsing / cross-refs / metadata, codetools usage
- subdir checks, top-level files, ASCII / encoding

### Does not catch
- Rust compile errors, `#[miniextendr]` macro expansion, ABI / linkage
- examples in Rd, `tests/testthat/*`, vignettes

For full validation pre-CRAN, run `miniextendr_check()` (which builds + vendors + runs cargo).

### Why a separate function (not a flag on `miniextendr_check()`)

Different *intent* (fast iteration vs full validation) and different *guarantee* (can't catch cargo-related failures). Hiding that behind a flag makes the diagnostic story muddier when "passes static" doesn't equal "passes CRAN". See `plans/check-static-no-compile.md` and `plans/revamp-minirextendr-check-static.md` for the full design rationale.

## Notes

- `roxygenize = TRUE` (default) opt-in: when `roxygen2` is installed, runs `roxygen2::roxygenize(path)` before checking — removes false-positive stale-doc warnings without touching cargo. Tests pass `roxygenize = FALSE` to keep them self-contained.
- Roxygen pin in `minirextendr/DESCRIPTION` bumped from `RoxygenNote: 7.3.3` to `Config/roxygen2/version: 8.0.0` because CRAN no longer serves 7.3.3 binaries; this aligns with `rpkg`'s existing 8.0.0 pin and produced minimal Rd churn (only `minirextendr-package.Rd` and the new `\seealso` block).
- Replaces the original implementation entirely (rebased off current `origin/main` rather than catching up 13 days of drift); preserves PR number + comment thread via force-push.

## Test plan

- [x] `just minirextendr-test` — 383/0/4 pre-existing warnings (unrelated `test-sync.R` autoconf-absent), 0 failures
- [x] `--filter check` — 8 PASS / 0 FAIL across the 3 new tests:
  - happy path on minimal R package
  - broken DESCRIPTION surfaces in `result\$errors` / `\$warnings`
  - **Rust-scaffolded fixture** (`_fixtures/static-check-rustpkg/`) confirms `--install=fake` skips configure even when `src/rust/Cargo.toml` + `configure` script are present; asserts `src/rust/target/` and `inst/vendor.tar.xz` are NOT created
- [x] roxygen regen produces bounded diff (3 files, 10 lines)

## References

- `r-svn tools/R/install.R:2424` — `use_configure <- FALSE` under `--fake`
- `r-svn tools/R/check.R:7371-7376` — examples/tests/vignettes auto-zeroed under fake install
- `r-svn tools/R/check.R:5929` — `--fake` passed to INSTALL_opts
- `plans/check-static-no-compile.md` — design doc with full rationale
- `plans/revamp-minirextendr-check-static.md` — revamp plan (6 design questions resolved)


CGMossa:
This is an idea really. We want a `R CMD check` that doesn't complain about rust powered R packages that do not vendor its dependencies. this will help development of rust powered r packages, without loosing complete compliance with `R CMD check`.

CGMossa:
While this is a good addition, I don't have an audience for it.

PR #297, opened 2026-04-21
miniextendr-api: add ColumnarDataFrame::with_column

CGMossa:
## Summary

- Adds `ColumnarDataFrame::with_column(self, name: &str, column: SEXP) -> Self` — replace a column by name with a pre-built SEXP.
- Matches existing `rename` / `drop` / `select` semantics: silent no-op on miss, builder-chain return.
- Caller is responsible for matching row length and a valid R vector; no validation.

## Why

Enables consumers to serialize a struct to a DataFrame via `vec_to_dataframe`, then swap individual columns to richer R types (POSIXct, Date, factor, ALTREP) without abandoning the serde path.

Motivating use case: downstream at A2-ai/dvs2 — presenting `add_time` as POSIXct in R by replacing the stringly-typed column the serializer produces from a `jiff::Timestamp`.

## Test plan

- [x] Unit: `test_columnar_with_column` — replaces an integer column with a character SEXP, verifies R type and values round-trip.
- [x] Unit: `test_columnar_with_column_noop` — nonexistent name is a silent no-op, other columns untouched.
- [x] `just clippy` — clean.
- [x] `just devtools-test` — 4561 PASS / 0 FAIL / 15 SKIP (skips match main).
- [x] R wrappers + NAMESPACE regenerated and committed.


CGMossa:
Alright, merging this. I think it's fine for now.

PR #308, opened 2026-04-24
chore: add RTK just shim for token-efficient recipe output

CGMossa:
## Summary

Adds `dev/just-bin/bash`, a thin shell wrapper that proxies every `just` recipe through [`rtk`](https://github.com/reachingforthejack/rtk) when it's installed, giving automatic token savings on all recipe output. Falls back to plain `bash` when rtk is absent — safe for all collaborators.

- `set shell` in all justfiles points to the shim instead of bare `bash`
- `[script("bash")]` blocks converted to `#!/usr/bin/env bash` (required: `[script]` bypasses `set shell`)
- `just_bin_dir` added to `PATH` so the shim is reachable by name from within recipes

## Files changed

- `dev/just-bin/bash` — shim (new)
- `justfile` — root just file
- `tests/cross-package/justfile`
- `minirextendr/inst/templates/rpkg/justfile` + `dev/just-bin/bash`
- `minirextendr/inst/templates/monorepo/justfile` + `dev/just-bin/bash`

## Test plan

- [ ] `just check` works with rtk installed (token savings visible in rtk gain)
- [ ] `just check` works without rtk installed (plain bash fallback)

### DIFF
--- a/dev/just-bin/bash
+++ b/dev/just-bin/bash
@@ -1,0 +1,42 @@
+#!/bin/sh
+set -eu
+
+self_dir=$(
+  CDPATH= cd -- "$(dirname -- "$0")" && pwd
+)
+
+strip_self_from_path() {
+  old_ifs=$IFS
+  IFS=:
+  new_path=
+  for entry in $1; do
+    if [ "$entry" = "$self_dir" ]; then
+      continue
+    fi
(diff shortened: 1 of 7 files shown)


CGMossa:
Rtk seems to be a nuisance when encountering issues. It doesn't vibe well with Claude.

PR #311, opened 2026-04-24
fix(serde): automatic NA type inference for all-None Option<T> columns

### LINKED ISSUE #307: ColumnarDataFrame: Option<T> field with all-None rows falls back to list column
## Problem

`miniextendr_api::serde::ColumnarDataFrame::from_rows` builds columnar R data.frames from a `Vec<T: Serialize>`. For a field `Option<U>` the docstring at `serde/columnar.rs:126` promises:

| Rust Type | R Column Type |
|---|---|
| `Option<T>` | Same type with NA for `None` |

This holds when **at least one** row has `Some(_)`. When every row has `None`, the column falls back to a list (`VECSXP`) with `NULL` entries, not an atomic vector with `NA`.

## Reproducer (sketch)

```rust
#[derive(Serialize)]
struct Row {
    stored: Option<u64>,
}

let rows = vec![Row { stored: None }, Row { stored: None }];
let df = ColumnarDataFrame::from_rows(&rows)?;
// Column `stored` is a list(NULL, NULL) in R, not c(NA_integer_, NA_integer_).
```

Seen in the wild in `A2-ai/dvs2` on `dvs_add(dry_run = TRUE)`, where `AddResult::stored_size: Option<u64>` is always `None` — R gets a `list(NULL)` column and callers have to flatten it themselves.

## Root cause

`serde/columnar.rs:568` `SchemaDiscoverer` probes runtime values to pick a `ColumnType`. For `Option<T>`, `serialize_none` (line 780) correctly keeps the current type so a later `Some` can upgrade it, but the default initial type is `ColumnType::Generic` (line 609). When the full row set emits only `None`, the type stays `Generic` → list column.

This is a fundamental limit of runtime serde probing: without a `Some` to dispatch on, there's no way to recover the Rust inner type of `Option<u64>` from a stream of `None`s.

## Proposed fix (attribute, minimum-breaking)

A field-level hint that sets `FieldInfo.col_type` up front and skips the probe:

```rust
#[derive(Serialize)]
pub enum AddDetail {
    Success {
        size: u64,
        #[miniextendr(column_type = "numeric")]
        stored_size: Option<u64>,
    },
    Error { error: String },
}
```

Valid values mirror `ColumnType`: `"logical"`, `"integer"`, `"numeric"`, `"character"`. Composes with `#[serde(flatten)]` and enum variants — the hint lives on the field, so `discover_schema_union` reads it from the `SchemaStructDiscoverer` side. Zero runtime cost.

## Proposed fix (trait, cleaner long-term)

A `ColumnKind` trait auto-implemented for the R-native scalar types and `Option<T: ColumnKind>`:

```rust
pub trait ColumnKind { const KIND: ColumnType; }
impl ColumnKind for u64     { const KIND: ColumnType = ColumnType::Real; }
impl ColumnKind for u32     { const KIND: ColumnType = ColumnType::Real; }
impl ColumnKind for i32     { const KIND: ColumnType = ColumnType::Integer; }
impl ColumnKind for f64     { const KIND: ColumnType = ColumnType::Real; }
impl ColumnKind for bool    { const KIND: ColumnType = ColumnType::Logical; }
impl ColumnKind for String  { const KIND: ColumnType = ColumnType::Character; }
impl<T: ColumnKind> ColumnKind for Option<T> { const KIND: ColumnType = T::KIND; }
```

`SchemaDiscoverer::process_field` consults `ColumnKind` first (type-level) and falls back to the value probe when the type isn't registered. Cleaner — `Option<u64>` just works without any attribute on the struct. Downside: it would be a breaking bound change on `ColumnarDataFrame::from_rows<T: Serialize>` unless gated behind a second entrypoint (e.g. `from_rows_typed`).

## Why not ALTREP

An ALTREP-backed column that wraps the `Vec<Option<u64>>` in place is possible, but the blast radius (C vtable code, GC dance, testing across R versions) is enormous for a case where the actual payload is a small numeric column. The atomic-with-NA representation is what users want; it's just a codec choice, not an abstraction-layer choice.

## Workaround today

On the dvs2 side I rebuild the column by hand and inject it via `.with_column`:

```rust
let stored_sizes: Vec<Option<u64>> = results.iter().map(|r| …).collect();
Ok(vec_to_dataframe(&results)?
    .with_column("stored_size", stored_sizes.into_sexp()))
```

This works because `Vec<Option<u64>>::into_sexp` (`into_r.rs::impl_vec_option_smart_i64_into_r!`) already does the right thing — picks `INTSXP` if all Some values fit i32, else `REALSXP`, with `NA` for `None`. It just isn't reachable from the serde columnar path when every row's value is `None`.

## Preference

The attribute fix if you want a small, non-breaking change now. The trait fix if you're OK with either a second entrypoint or a major-version bump. Happy to open a PR for whichever direction you prefer.

CGMossa:
## Summary

- Remove `ColumnarBuilder` and `hint()` API entirely — no user-facing annotations needed
- All-None `Option<f64/i32/bool/String>` columns now automatically produce the correct typed NA vector (`NA_real_`, `NA_integer_`, `NA`, `NA_character_`) using `std::any::type_name::<T>()` inside `process_field` when `TypeProbe` returns `Generic`
- `ColumnType` is crate-private again; only `ColumnarDataFrame` and `vec_to_dataframe` are re-exported
- Documentation updated to reflect automatic inference

## Test plan

- [ ] `just devtools-test` — 4664 pass, 0 fail (verified locally)
- [ ] `clippy_all` — clean (verified locally)
- [ ] All-None R tests: `test_columnar_all_none_real/int/bool/str` pass without any hint API
- [ ] Leading-None upgrade still works: `test_columnar_leading_none`

Closes #307


CGMossa:
the `.hint` is ugly. Will need to find an alternative plan for this.

CGMossa:
Closing this PR. The type-name string lookup approach is a dead end.

The root cause is a fundamental limitation of serde's protocol: `serialize_none()` carries zero inner-type information. Our fix used `std::any::type_name::<T>()` to infer the column type for all-None `Option<T>` columns — a closed string whitelist that doesn't compose with user-defined types or third-party structs.

The alternatives all have worse trade-offs:
- **TypeId**: requires `T: 'static`, but serde's `serialize_field` only guarantees `T: ?Sized + Serialize` — can't be added without breaking the protocol
- **New trait bound on `from_rows`**: forces `#[derive(ColumnarRow)]` on every row struct, and locks out third-party types entirely (orphan rules)
- **Separate derive macro**: composable, but a large feature that should be designed properly, not bolted onto this fix

The `.hint()` API removed by this PR was actually the most honest interface for this problem — it made the type erasure explicit. The "automatic" inference is an illusion that only works for a hardcoded set of standard-library types.

The right fix is a rethink of the columnar serialization strategy, not a patch on top of the current one.

PR #316, opened 2026-04-25
fix(macros): pass `.call = match.call()` from sidecar accessors

CGMossa:
## Summary

- **Bug fix**: sidecar `Type_get_field` / `Type_set_field` wrappers from `#[derive(ExternalPtr)]` were calling `.Call(C_..., x)` without `.call = match.call()`. The C wrappers take `__miniextendr_call: SEXP` as their first parameter, so omitting it shifted the user's `x` argument into the call slot and broke error attribution under panic. Twelve sites covered (getter+setter × Env/R6/S3/S4/S7/Vctrs).
- **DotCallBuilder unification**: routed every remaining hand-rolled `format!(".Call(..., .call = match.call(), ...)")` through `DotCallBuilder` — `MethodContext::{static_call, instance_call}`, `match_arg_prelude`, S4 instance methods, S7 validator/convert_from/convert_to, R6 finalizer/deep_clone/active-binding setter, and the standalone `match_arg` choices helper. Output is byte-identical for all routed sites; the only behavioural diff in `rpkg/R/miniextendr-wrappers.R` is exactly the 46 sidecar lines (verified).
- **Regression test**: `externalptr_derive::tests::sidecar_accessors_pass_match_call` exercises every `ClassSystem` variant of `generate_r_wrapper_for_slot` and asserts both getter and setter contain `.call = match.call()`.
- **Documentation**: new `docs/CALL_ATTRIBUTION.md` with a runnable fixture (`rpkg/src/rust/call_attribution_demo.rs`) showing two functions that raise the same error — one through the standard wrapper, one as `extern "C-unwind"` with no wrapper. Transcript captured from `Rscript` against the freshly installed package; auto-regenerated to `site/content/manual/call-attribution.md` via `just site-docs`.

## Why the sidecar gap mattered

`error_in_r_check_lines` reads `.val\$call %||% sys.call()`. The `.val\$call` field is populated by the C wrapper, which receives `__miniextendr_call` as the **first** SEXP and threads it through `with_r_unwind_protect_error_in_r`. When the R wrapper wrote `.Call(C_..._get_field, x)` instead of `.Call(C_..._get_field, .call = match.call(), x)`, the first positional SEXP arriving in C was `x` (the externalptr) — so the call slot held a pointer-y SEXP and the field accessor's first formal received nothing. Either way error attribution was wrong; under panic the diagnostic could be misleading or worse.

## Demo from the new fixture

```r
> outer_with    <- function(x) call_attr_with(x, x + 1L)
> outer_without <- function(x) miniextendr:::unsafe_C_call_attr_without(x, x + 1L)

> outer_with(5L)
Error in call_attr_with(left = x, right = x + 1L) :
  left + right is too risky

> outer_without(5L)
Error in miniextendr:::unsafe_C_call_attr_without(x, x + 1L) :
  left + right is too risky

> class(tryCatch(outer_with(5L), error = identity))
[1] "rust_error"  "simpleError" "error"       "condition"

> class(tryCatch(outer_without(5L), error = identity))
[1] "simpleError" "error"       "condition"
```

Three concrete benefits the wrapped path delivers: formal arg name matching, public symbol attribution (no `:::` leak of internal names), and structured `rust_error` class.

## Follow-ups (deferred to issues)

- #317 — Rust macros `condition!` / `error!` / `message!` / `warning!` for the full `simpleCondition` / `simpleError` / `simpleMessage` / `simpleWarning` set, layered with `rust_*` subclasses.
- #318 — The new demo fixture (`call_attr_with`, `unsafe_C_call_attr_without`) is exported into the public namespace; should be marked internal or moved into a test.
- #319 — `match.call()` in R6 finalizer / `deep_clone` / S7 property getter/setter/validator captures the internal R6/S7 dispatch frame, not the user's call. Decide between dropping `.call`, synthesizing a meaningful call expression, or walking back via `sys.call(-N)`.

## Test plan

- [x] `cargo ltest -p miniextendr-macros` — 285 passed (284 prior + new sidecar invariant test)
- [x] `just clippy` (workspace + cross-package + rpkg) — clean
- [x] `just configure && just rcmdinstall && just devtools-document` — green; wrapper diff is exactly 46 sidecar lines + the new demo wrappers, no unintended drift
- [x] Demo fixture run via `Rscript` against the installed package — error-attribution differences confirmed visually and via `tryCatch` class inspection
- [x] `zola check` on the regenerated site — 88 pages, 0 orphans, links clean


CGMossa:
Superseded by #344, which folds in the condition system unification (#317 + #319) plus a correction: the sidecar accessor C wrappers generated by `externalptr_derive.rs` actually take `(x)` / `(x, value)` with no `__miniextendr_call` slot — distinct from the `c_wrapper_builder.rs` C wrappers this PR was modeled on. The `.call = match.call()` added here caused arity errors at runtime (test-sidecar.R failures). #344 reverts that part and routes sidecar errors through the standard `%||% sys.call()` fallback, alongside the same fallback for the R6/S7 lambda contexts in #319. The `DotCallBuilder` unification, the call-attribution demo fixture, and `docs/CALL_ATTRIBUTION.md` are all preserved.

PR #334, opened 2026-04-29
fix(rpkg,minirextendr): cd into rust/ before invoking cargo

CGMossa:
## Summary

R CMD INSTALL runs `make` from `src/`, but cargo only walks up from the current working directory when discovering `.cargo/config.toml` — it does **not** walk up from `--manifest-path`. So the source-replacement config that `configure` writes at `src/rust/.cargo/config.toml` during tarball installs was never read, and `cargo build --offline` fell back to fetching git deps from the network and failed.

This was masked locally and in CI because the cargo cache (`~/.cargo/git/`) usually had `miniextendr` fetched already, so the offline mode found things in the per-user cache. Reproduces deterministically with empty `CARGO_HOME` + `CARGO_NET_OFFLINE=1`:

```
error: failed to get `miniextendr-api` as a dependency of package `dvs-rpkg`
Caused by: failed to load source for dependency `miniextendr-api`
Caused by: unable to update https://github.com/A2-ai/miniextendr#<sha>
Caused by: can't checkout from 'https://github.com/A2-ai/miniextendr': you are in the offline mode (--offline)
```

The fix adds `cd "$(ABS_RPKG_SRCDIR)/rust"` before each cargo invocation in both the staticlib and cdylib recipes. All paths passed to cargo (`--manifest-path`, `--target-dir`, link-args) are already absolute, so the cwd switch only affects config discovery.

Mirrored in `minirextendr/inst/templates/rpkg/Makevars.in` so freshly scaffolded packages get the fix too.

## Test plan

- [ ] `R CMD INSTALL <built-tarball>` with `CARGO_NET_OFFLINE=1` and an empty `CARGO_HOME` succeeds
- [ ] `just rcmdinstall` still green for the rpkg in this repo
- [ ] cross-package tests still green

### DIFF
--- a/minirextendr/inst/templates/monorepo/rpkg/Makevars.in
+++ b/minirextendr/inst/templates/monorepo/rpkg/Makevars.in
@@ -79,6 +79,7 @@ $(CARGO_AR): FORCE_CARGO $(OBJECTS)
 	LINK_ARGS=""; \
 	for obj in $(OBJECTS); do LINK_ARGS="$$LINK_ARGS -C link-arg=$(ABS_RPKG_SRCDIR)/$$obj"; done; \
 	if [ -n "$(CARGO_BUILD_TARGET)" ]; then TARGET_OPT="--target $(CARGO_BUILD_TARGET)"; fi; \
+	cd "$(ABS_RPKG_SRCDIR)/rust"; \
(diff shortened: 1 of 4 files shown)


CGMossa:
Failure on CI is unrelated to this PR. Merging.

PR #349, opened 2026-05-03
feat(#345): rust_* class layering for trait-ABI shims + ALTREP RUnwind callbacks

### LINKED ISSUE #345: Route trait-ABI and ALTREP `RUnwind` panics through the tagged-SEXP path for `rust_*` class layering
## Context

PR #344 unified the Rust→R condition pipeline so that `#[miniextendr]` functions raise structured conditions with layered `rust_*` classes (`c("rust_error", "simpleError", "error", "condition")` and analogous for warning / message / condition). This works because the default `error_in_r` mode returns errors as a tagged SEXP and lets the R wrapper raise `stop(structure(...))` with the layered class vector.

Two callsites still use the non-`error_in_r` path (`with_r_unwind_protect` → `Rf_errorcall`):

1. **Trait-ABI vtable shims** — `miniextendr-macros/src/miniextendr_trait.rs:808` and `miniextendr-macros/src/miniextendr_impl_trait/vtable.rs:489`. Cross-package C-ABI calls dispatched via `R_GetCCallable`.
2. **ALTREP `RUnwind` guard** — `miniextendr-api/src/ffi_guard.rs:71` → `with_r_unwind_protect_sourced`. Protects callbacks invoked from R's GC / vector dispatch.

A `panic!` from either of these surfaces as a plain `simpleError`, *losing* the `rust_error` class layering. `tryCatch(rust_error = h, …)` does **not** match it.

This is documented as a known limitation in `docs/CONDITIONS.md` (\"In trait-ABI shims and ALTREP callbacks…\") and `miniextendr-api/src/unwind_protect.rs` (line 271 doc comment, line 291 inline comment).

## Why it's hard

There's no R wrapper sitting between the user's R code and the C entry point in these contexts:

- **Trait shims** are reached via `R_GetCCallable` from another package — no `R/<wrapper>.R` file generated.
- **ALTREP callbacks** are invoked directly by R's runtime (vector dispatch / GC), not via `.Call`.

Without an R wrapper, there's no place to inspect a tagged SEXP and translate it into `stop(structure(...))`.

## Possible approaches

1. **Synthesize an R-side wrapper for trait-ABI calls.** Cross-package `#[miniextendr]` trait imports already generate Rust shims; extend the codegen to emit a tiny R closure that calls the shim and inspects the return for `rust_error_value`. Adds cost to every cross-package method invocation.

2. **Walk up the R stack from inside the shim** to find the nearest `error_in_r` frame and stash the class info out-of-band (e.g., in a thread-local picked up by the parent's `error_in_r_check_lines`). Fragile — relies on always being called from a `with_r_unwind_protect_error_in_r` context, which isn't guaranteed for ALTREP callbacks.

3. **Layer class info into the `Rf_errorcall` message itself** with a sentinel prefix that the *caller's* R wrapper recognises and rewrites. Brittle string-matching.

4. **Accept the limitation** and document it more prominently. Users who need `rust_error` matching from trait methods can wrap the call in a thin `#[miniextendr]` standalone function.

(1) is the most principled. (4) is the cheapest. (2) and (3) are likely dead ends.

## Acceptance criteria

- [ ] A `panic!()` from a cross-package trait method matches `tryCatch(rust_error = h, …)`.
- [ ] A `panic!()` from an ALTREP `RUnwind` guard matches `tryCatch(rust_error = h, …)` *when* the callback was invoked from a `#[miniextendr]` function (best effort — direct R-runtime invocation may not be feasible).
- [ ] No regression in non-panic call performance for trait method invocations and ALTREP callbacks.
- [ ] `docs/CONDITIONS.md` "Known limitations" section updated or removed.

## Out of scope

- `warning!()` / `message!()` / `condition!()` from these contexts. The non-error_in_r path can't suspend execution to call back into R, so non-fatal signals would still need to degrade. Document, don't fix.

CGMossa:
Closes #345.

## Summary

Two complementary fixes that close the last gap in the post-#344 condition
pipeline. Cross-package trait method panics and ALTREP `r_unwind` callback
panics now both produce R conditions that match `tryCatch(rust_error = h, ...)`
and `tryCatch(my_class = h, ...)` (when `error!(class = "my_class", ...)` is
used).

- **Trait-ABI shims** (Approach 1 from the plan): the vtable shim returns a
  tagged SEXP on panic instead of `Rf_errorcall`-ing. The generated View
  method wrapper inspects the result via `repanic_if_rust_error`, which
  reconstructs the `RCondition` via `RCondition::from_tagged_sexp` and
  re-panics. The consumer's outer `error_in_r` guard catches the re-panic
  and produces the tagged SEXP for the consumer's R wrapper. End-to-end
  behaviour is identical to a same-package call.

- **ALTREP `r_unwind` callbacks** (Approach 3 from the plan): the guard
  raises the R condition by `Rf_eval`-ing a hand-built `stop(structure(list(message, call), class = c(...)))`
  call. No R wrapper required — works directly in R's vector-dispatch
  context. The class vector is built directly with `set_class` (skipping
  the `structure()` wrapper call); `stop` is a cached interned symbol.

Plain Rust panics from the same contexts also pick up `rust_error` class
layering as a free side-effect — `tryCatch(rust_error = h, ...)` matches
panics from any source going through `with_r_unwind_protect_sourced`, not
just `error!()`-originated ones.

## Test results

- `just check` ✓
- `just clippy` ✓ (project recipe — iterates every workspace)
- `cargo clippy --workspace --all-targets --locked --features <full CI feature set> -- -D warnings` ✓ (CI clippy_all)
- `just test` ✓
- `just devtools-test`: `[ FAIL 0 | WARN 0 | SKIP 15 | PASS 4850 ]` (56 new tests in `test-altrep-conditions.R`)
- `just cross-test`: `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 262 ]` (5 new tests in `test-error-classes.R`)

## Remaining limitations (documented in `docs/CONDITIONS.md`)

- `warning!()` / `message!()` / `condition!()` from an ALTREP `r_unwind`
  callback degrade to a plain R error. R's vector-dispatch context can't
  cleanly suspend execution to deliver a non-fatal signal. This is an
  inherent constraint of ALTREP, not a regression.
- A trait `View::method()` called from Rust code that is not wrapped in
  `with_r_unwind_protect_error_in_r` falls back to the worker thread's
  `catch_unwind` boundary without `rust_*` layering. Every `#[miniextendr]`
  fn already provides the outer guard, so this only affects unusual
  callers (test harnesses, init callbacks).

## Notes for review

- `repanic_if_rust_error` wraps the read window in `OwnedProtect` defensively.
  The reads are non-allocating today, but R-devel's GC is aggressive enough
  that future regressions in the read path could turn that into a UB
  source — the guard is cheap insurance (see commit `aeb2ad2b`).
- `raise_rust_condition_via_stop` allocates STRSXP + VECSXP + names + call;
  PROTECT discipline mirrors `make_rust_condition_value` in
  `error_value.rs` (the established safe reference). The 5-allocation
  protect stack is rewound by R's longjmp machinery on `Rf_eval(stop(...))`.
- The plan document `plans/issue-345-trait-altrep-class-layering.md` ships
  with the PR for context. It documents three approaches; the rejected
  one (synthesized R closure for trait imports, "Approach 2") was
  evaluated separately and confirmed worse — see the plan's "Approaches"
  section for the architectural reason.

## Test plan

- [x] `just check`
- [x] `just clippy` + `clippy_all` features matrix
- [x] `just test`
- [x] `just devtools-test` (rpkg testthat — adds 56 tests in `test-altrep-conditions.R`)
- [x] `just cross-test` (cross-package trait-ABI — adds 5 tests in `test-error-classes.R`)
- [ ] CI green


CGMossa:
Pushed review fixes (commit `e9fd7f37`) addressing the items from the review:

**In-PR fixes:**
- `condition.rs:from_tagged_sexp` — added `len < 2` early-exit before `vector_elt(0)/(1)` reads (defense-in-depth).
- `condition.rs:from_tagged_sexp` — corrected the doc comment; code reconstructs Warning/Message/Condition variants per kind, doc had claimed they collapse to `Error`.
- `rpkg/DESCRIPTION` + `rpkg/man/*.Rd` — regenerated with roxygen2 **8.0.0** (CRAN current). Drops the `RoxygenNote: 7.3.2` stamp and ~70 .Rd files of incidental churn (R6 doc style switch, spurious `\docType{data}`/`\format{...}`/`\keyword{datasets}` blocks, missing Authors entry in `miniextendr-package.Rd`). Net diff vs main is now exactly the new `altrep_condition_tests.Rd` and 3 NAMESPACE export lines for the new fixtures.

**Follow-up issues filed** (deferred from review):
- #366 — route ALTREP `warning!`/`message!`/`condition!` degradation through `raise_rust_condition_via_stop` for consistency with the plain-panic upgrade
- #367 — `scripts/plans-to-json.sh` italic-strip regex mangles snake_case identifiers (`rust_error` → `rusterror`)

Sanity-checked locally: `cargo clippy -p miniextendr-api --all-targets -- -D warnings` clean.

Issue #319, opened 2026-04-25
`match.call()` in R6 GC / S7 dispatch lambdas captures internal frames, not user calls


CGMossa:
The current codegen emits `.call = match.call()` from several lambda contexts where `match.call()` does not capture the user's call frame at all — it captures the internal R6 / S7 dispatch frame that *invokes* the lambda. The result is harmless but semantically noisy: error messages from these paths attribute to internal machinery rather than to user code.

## Affected codegen sites

Each emits `.call = match.call()` today (via `DotCallBuilder` after the PR #316 unification):

| Site | Generator | Caller of the lambda |
|---|---|---|
| R6 finalizer | `miniextendr_impl/r6_class.rs:382-389` | R's GC, when the externalptr is finalized |
| R6 `deep_clone` | `miniextendr_impl/r6_class.rs:393-407` | R6's `\$clone(deep = TRUE)` machinery |
| S7 property getter | `miniextendr_impl/s7_class.rs` (via `MethodContext::instance_call(\"self@.ptr\")`) | S7 dispatch on `obj@prop` |
| S7 property setter | same | S7 dispatch on `obj@prop <- value` |
| S7 property validator | `miniextendr_impl/s7_class.rs:430-437` | S7 on object construction / property assignment |

In every case the lambda runs *inside* a frame that R6 or S7 set up to call it. `match.call()` evaluated in that frame returns something like `<R6 finalize call>` or the S7 dispatch closure's arguments — not the user's `obj\$x` access or `obj\$clone(deep = TRUE)` invocation.

## Why it is harmless today

`Rf_errorcall(call, msg)` accepts any SEXP for the call slot, including the irrelevant internal-frame call. R prints something at the user, just not the user's actual expression. None of these paths panic in normal operation, so the practical visibility is low.

## Why it is worth fixing

Every other `.Call()` site in the framework attributes to the user's expression. These five carve out an exception that surfaces as confusing error messages when the underlying Rust *does* panic — e.g., a finalizer panic shows `Error in finalize() : ...` with no information about which object's finalizer failed.

## Options

1. **Drop `.call = match.call()` from these sites** and pass `R_NilValue` (i.e., omit the named arg). On panic, R falls back to `sys.call()` of the wrapper frame, which is at least the externalptr accessor, not the lambda's internal-dispatch frame. Slightly less informative than ideal, but consistent with how `extern \"C-unwind\"` paths render.
[...]

Issue #317, opened 2026-04-25
Rust macros: `condition!` / `error!` / `message!` / `warning!` for structured R conditions


CGMossa:
## Goal

Give Rust code a `println!`-shaped surface for raising every flavour of R condition — `simpleCondition`, `simpleError`, `simpleMessage`, `simpleWarning` — instead of forcing every Rust failure through `panic!` (which always becomes `simpleError` via `Rf_errorcall`).

Each condition raised from Rust must also carry a `rust_*` subclass at the front of its class vector (`rust_error` already exists on the current panic path) so handlers can specifically catch Rust-originated conditions with `tryCatch(rust_warning = ...)` etc.

## Proposed API (Rust side)

Four format-string macros, each taking the same `format_args!` shape as `println!`:

```rust
// signalCondition(<rust_condition / simpleCondition / condition>(msg, call))
condition!("user is in {} mode", mode);

// stop(<rust_error / simpleError / error / condition>(msg, call))
error!("missing field: {}", name);

// message(<rust_message / simpleMessage / message / condition>(msg))
message!("loaded {} rows", n);

// warning(<rust_warning / simpleWarning / warning / condition>(msg, call))
warning!("deprecated arg `{}`", arg);
```

Each macro formats its args, wraps the resulting `String` in a tagged payload, and `panic_any`s it:

```rust
#[doc(hidden)]
pub enum RCondition {
    Condition(String),
    Error(String),
    Message(String),
    Warning(String),
}

// (sketch — full impl handles class-attr extension, structured fields, etc.)
macro_rules! error {
    ($($arg:tt)*) => {
        std::panic::panic_any($crate::RCondition::Error(format!($($arg)*)))
    };
}
```

`with_r_unwind_protect` already catches Rust panics; the new path is to recognise an `RCondition` payload before falling through to the generic panic-to-Rf_errorcall route, then dispatch:

| Variant | R primitive | Class vector | Notes |
|---|---|---|---|
| `Condition(msg)` | `signalCondition(structure(...))` | `c("rust_condition", "simpleCondition", "condition")` | Non-fatal — handlers can resume |
| `Error(msg)` | `stop(structure(...))` | `c("rust_error", "simpleError", "error", "condition")` | Equivalent to current `panic!` path (which already uses `rust_error`) |
| `Message(msg)` | `message(structure(...))` | `c("rust_message", "simpleMessage", "message", "condition")` | Goes to stderr, not via error system |
[...]

PR #381, opened 2026-05-05
just fmt

CGMossa:
- **plans: condition system unification (#316 + #319 + #317 + Rf_error doc cleanup)**
- **ran `just fmt`**


CGMossa:
This was not just a chore. Doing things by hand is messy 🤷

PR #391, opened 2026-05-05
WIP: WebR / WASM support — planning + design docs only

CGMossa:
**Status:** draft. Pure docs + plans, no code changes. Picking this up later — opening the PR so the design discussion is preserved as text rather than scrollback.

## What's in this branch

Four artefacts, fully cross-linked:

| File | Role |
|---|---|
| `plans/webr-support.md` | Index plan. Investigation summary of `.webr/` build flow + the source-side port checklist. Dispatches to the two detailed plans. |
| `plans/wasm-registry-codegen.md` | The linkme replacement. Host-time snapshot codegen → `wasm_registry.rs` consumed under `cfg(target_arch = "wasm32")`. |
| `plans/webr-dockerfile.md` | Build environment. `Dockerfile.webr` inheriting `ghcr.io/r-wasm/webr` digest-pinned + the layered tools + CI sequencing. |
| `docs/WEBR.md` | User-facing rationale (target triple, why nightly, why `-Z build-std=std,panic_abort`). Linked from `docs/README.md`. |

## TL;DR of the design

- **Target:** `wasm32-unknown-emscripten` (not `wasm32-unknown-unknown` — we need Emscripten's libc/pthread shims).
- **Toolchain:** nightly Rust + `-Z build-std=std,panic_abort` is mandatory. Two reasons: (1) rustup's precompiled `std` is built against whatever Emscripten was current when it was cut; webR pins its own Emscripten, and ABI mismatches cause link errors or runtime UB. (2) `panic = "abort"` has to apply to `std` itself; stable cargo can't relink the precompiled std.
- **linkme:** hard-errors on `wasm32-*` (verified in `linkme-impl-0.3.36/src/declaration.rs:48-51`). Of the 7 distributed slices, 4 are write-time only and can be empty on WASM; **3 are runtime-critical** (`MX_CALL_DEFS`, `MX_ALTREP_REGISTRATIONS`, `MX_TRAIT_DISPATCH`) and need a replacement registration mechanism.
[...]

PR #395, opened 2026-05-05
fix(macros): stabilise symbol names for ALTREP register fns + trait vtables (#392 #393 #394)

### LINKED ISSUE #392: ALTREP register fn missing pub + #[no_mangle] + extern "C"
**Found via audit during WASM-port planning ([#391](https://github.com/A2-ai/miniextendr/pull/391)).**

The macro that generates ALTREP registration functions emits a plain `fn` with no visibility, no `#[no_mangle]`, and no `extern "C"` ABI:

`miniextendr-macros/src/altrep.rs:67-74`:

```rust
let reg_fn_name = quote::format_ident!("__mx_altrep_reg_{}", ident.to_string().to_lowercase());
quote::quote! {
    #[::miniextendr_api::linkme::distributed_slice(::miniextendr_api::registry::MX_ALTREP_REGISTRATIONS)]
    #[linkme(crate = ::miniextendr_api::linkme)]
    #[doc(hidden)]
    fn #reg_fn_name() {
        <#ident as ::miniextendr_api::altrep_registration::RegisterAltrep>::get_or_init_class();
    }
}
```

This works today because linkme takes the function's address and the host linker resolves it intra-crate. But:

1. The function isn't externally addressable by a separate compilation unit — required for the WASM codegen path (`plans/wasm-registry-codegen.md`).
2. `__mx_altrep_reg_<lowercase>` collides for ALTREP types that differ only in case (`MyType` vs `MYType`). Vanishingly unlikely in practice, but a real bug.

**Fix:**

```rust
#[unsafe(no_mangle)]
pub extern "C" fn #reg_fn_name() {
    <#ident as ::miniextendr_api::altrep_registration::RegisterAltrep>::get_or_init_class();
}
```

Also drop `.to_lowercase()` to use the case-preserved ident (removes the case-collision footgun while we're here).

Single emission site; no other code emits `__mx_altrep_reg`. Standalone fix, no behaviour delta on host. Land before [#391](https://github.com/A2-ai/miniextendr/pull/391) needs to consume these symbols.

**Generic ALTREP types:** the macro deliberately emits an empty token stream when `generics.params.is_empty()` is false. Generic ALTREP types can't auto-register today either; not a regression, but worth documenting alongside this fix.

### LINKED ISSUE #393: Trait vtable static missing pub + #[no_mangle]
**Found via audit during WASM-port planning ([#391](https://github.com/A2-ai/miniextendr/pull/391)).**

The macro that emits per-`impl Trait for Type` vtable statics produces them with no visibility, no `#[no_mangle]`, and no `#[repr(C)]`. Two emission sites in `miniextendr-macros/src/miniextendr_impl_trait.rs`:

- `expand_miniextendr_impl_trait` around line 799
- `generate_tpie_invocation` around line 921-925

Pattern (from line 921-925):

```rust
static #vtable_static_name: #vtable_type_path =
    #builder_path::<#concrete_type>();
```

Name format: `__VTABLE_{TRAIT_UPPER}_FOR_{TYPE_UPPER}` (e.g. `__VTABLE_COUNTER_FOR_MYTYPE`).

This works today because linkme stores the vtable address as opaque `*const c_void` data and the host linker resolves intra-crate references. But the static isn't externally addressable by a separate compilation unit — required for the WASM codegen path (`plans/wasm-registry-codegen.md`).

**Fix:**

```rust
#[unsafe(no_mangle)]
pub static #vtable_static_name: #vtable_type_path =
    #builder_path::<#concrete_type>();
```

Apply at *both* emission sites. `#[repr(C)]` belongs on the `*VTable` *struct definition* (likely in `miniextendr-api` or generated by the trait-attribute macro), not on the static itself — verify it's already present for FFI; if not, add it as part of this fix.

Standalone, no behaviour delta on host. Land before [#391](https://github.com/A2-ai/miniextendr/pull/391) needs to consume these symbols.

**Note:** the related generic-collision bug is filed separately — see the companion issue. Both fixes will likely land in the same PR.

### LINKED ISSUE #394: Generic trait impls produce vtable name collision (`__VTABLE_TRAIT_FOR_TYPE`)
**Found via audit during WASM-port planning ([#391](https://github.com/A2-ai/miniextendr/pull/391)).** Pre-existing bug, not WASM-specific.

`miniextendr-macros/src/miniextendr_impl_trait.rs:type_to_uppercase_name` derives the vtable static's symbol name from the *last path segment's ident only*, uppercased. Generic args on the concrete type are stripped:

- `impl Trait for MyType<u32>` → `__VTABLE_TRAIT_FOR_MYTYPE`
- `impl Trait for MyType<f64>` → `__VTABLE_TRAIT_FOR_MYTYPE`

Two distinct monomorphisations, identical static name.

**Why it's hidden today:** linkme stores each vtable as opaque `*const c_void` data and gathers the entries by linker section. The two static items end up in the same section with the same name; the linker either deduplicates them (silently picking one — wrong vtable wins) or errors out depending on the section/visibility flags. Today's symbols have no visibility (see companion issue), so they're treated as local statics that don't conflict cross-translation-unit but still race within a crate.

**Why it matters now:** any non-linkme codegen path needs to reference these statics by name from a separate compilation unit. Both `MyType<u32>` and `MyType<f64>`'s vtables have the same name → the WASM codegen can't disambiguate them via `extern { static __VTABLE_TRAIT_FOR_MYTYPE: …; }`. It's a hard blocker for the linkme replacement in [#391](https://github.com/A2-ai/miniextendr/pull/391).

**Fix:**

Append a stable hash of the full generic args to the symbol name. Sketch:

```rust
let mangled = mangle_generics(&concrete_type);   // e.g. SipHash13 of token stream
let vtable_static_name = format_ident!("__VTABLE_{}_FOR_{}_{}",
    trait_name_upper, type_name_upper, mangled);
```

The hash needs to be deterministic across builds — use `proc_macro2::TokenStream::to_string()` of the canonical concrete-type tokens, hash it with a stable hasher.

**Backwards-compatibility:** unlikely but possible that user code references the unmangled `__VTABLE_*` name directly (debug tooling, custom dispatchers). Worth a CHANGELOG note when this lands. No CRAN concern (this is internal).

**Test:** add a regression — two `impl Trait for MyType<T>` blocks for distinct `T`s, assert both round-trip through trait dispatch. Today's behaviour is "one of them wins silently."

Companion to the visibility fixes for `__VTABLE_*` (see linked issue) — both should land together in step 1 of [#391](https://github.com/A2-ai/miniextendr/pull/391)'s implementation order.

CGMossa:
## Summary

Fixes three macro-emission bugs identified during WASM-port planning (#391). All three affect symbol addressability from external compilation units.

### #392 — ALTREP registration functions: `pub unsafe extern "C"` + `#[unsafe(no_mangle)]`

The `__mx_altrep_reg_<Ident>` functions emitted by `#[derive(AltrepInteger)]` and friends were private plain Rust `fn`s — not addressable from a separate compilation unit. Fixed by:
- Emitting `pub unsafe extern "C" fn` with `#[unsafe(no_mangle)]`
- Using the verbatim ident (dropped `.to_lowercase()` — case collision footgun between `MyType` vs `MYType`)
- Updating `MX_ALTREP_REGISTRATIONS: [unsafe extern "C" fn()]` in `registry.rs` and the call site

### #393 — Vtable statics: `pub` + `#[unsafe(no_mangle)]`

The `__VTABLE_{TRAIT}_FOR_{TYPE}` statics emitted at both sites in `miniextendr_impl_trait.rs` (`generate_vtable_static` and `generate_tpie_invocation`) were private. Fixed by adding `#[unsafe(no_mangle)]` + `pub` at both emission sites. `#[repr(C)]` was already present on the `*VTable` struct definition — no change needed.

### #394 — Generic type vtable name collision

`type_to_uppercase_name` stripped generic arguments, so `MyType<u32>` and `MyType<f64>` both produced `MYTYPE`, causing a silent vtable collision. Fixed by appending a 16-char FNV-1a-64 hex suffix of the full canonical token stream when the type has generic args: `MyType<u32>` → `MYTYPE_<hash>`. Plain non-generic `MyType` keeps its clean `MYTYPE` form.

The hash function is FNV-1a-64 — ~10 lines, no new dependencies, and fully deterministic across builds (unlike `DefaultHasher`/`RandomState`).

A regression test (`test_type_to_uppercase_name_generic_distinct`) was added to `miniextendr_impl_trait/tests.rs` asserting distinct 16-hex-char suffixes for distinct monomorphisations and determinism across calls.

## Test plan

- [x] `cargo test -p miniextendr-macros` — 289 passing (was 287; +2 new)
- [x] `just check` — all workspace manifests clean
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (clippy_default)
- [x] `cargo clippy --workspace --all-targets --locked --features rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,connections,nonapi,default-strict,default-coerce,default-r6,default-worker -- -D warnings` (clippy_all)
- [x] `just configure && just rcmdinstall && just devtools-document` — R package installs and documents successfully
- [x] Pre-existing R-integration SIGSEGV failures (externalptr, from_r, etc.) are pre-existing and unrelated to these changes — verified by reproducing without our changes

## Prerequisites / related

This is a prerequisite for the WASM registry-codegen port tracked in #391.

Closes #392
Closes #393
Closes #394


CGMossa:
Closing in favour of #NNN — same fix on a fresh branch (`fix/macro-symbol-stability`) cut from `main`, so the PR contains only the macro symbol-stability commit, not the two unrelated commits (`94ca0da9`, `1804bb18`) that were already on `refactor/raise-condition-helper` from prior work.

The new PR also has the `unsafe`-on-fn-signature tightening folded in (drops the `unsafe` from `__mx_altrep_reg_*` and reverts the `MX_ALTREP_REGISTRATIONS` slice element type back to `[extern "C" fn()]`, matching the convention used by `MX_CALL_DEFS`'s `extern "C-unwind" fn` entries).

`refactor/raise-condition-helper` has been force-pushed back to `1804bb18` (its pre-agent state — the user's two original commits are preserved). New PR link will be posted in a moment.

CGMossa:
Re-opened cleanly as #397.

PR #398, opened 2026-05-05
[codex] Fix generic trait vtable symbol names

### LINKED ISSUE #394: Generic trait impls produce vtable name collision (`__VTABLE_TRAIT_FOR_TYPE`)

CGMossa:
## Summary

Closes #394.

- add a deterministic hash suffix to generated trait vtable symbols when the trait path or concrete type has generic arguments
- share that naming helper across regular trait impl codegen and empty trait-impl expansion
- add regression coverage for generic concrete types, generic trait paths, and the non-generic compatibility case

## Validation

- cargo test -p miniextendr-macros miniextendr_impl_trait -- --nocapture
- just check
- cargo clippy --workspace --all-targets --locked -- -D warnings

## Notes

Non-generic impls keep the previous __VTABLE_{TRAIT}_FOR_{TYPE} spelling. Generic impls now append _H{stable_hash} to the type portion so distinct monomorphisations do not collapse to the same static symbol.


CGMossa:
Closing as a duplicate: open PR #397 already covers and closes #394.

PR #410, opened 2026-05-06
chore(rpkg): declare R (>= 4.4) dependency for %||% (#384)

CGMossa:
## Summary

- `rpkg/DESCRIPTION` had no minimum R version declared, but the generated `rpkg/R/miniextendr-wrappers.R` uses `%||%` (introduced in base R 4.4.0).
- Adds `Depends: R (>= 4.4)` after the `Version:` field, following standard CRAN placement convention.

## Plan

See `plans/issue-384-r-44-depends.md` for full audit plan.

## Audit findings

`%||%` is confirmed in use in `rpkg/R/miniextendr-wrappers.R` (line 18: `.call <- .val$call %||% .call_default`). The `Depends` declaration is necessary.

The cross-package test fixtures `tests/cross-package/consumer.pkg/R/consumer.pkg-wrappers.R` and `tests/cross-package/producer.pkg/R/producer.pkg-wrappers.R` also use `%||%` in their generated wrappers, but their `DESCRIPTION` files likewise lack an R minimum version. This is out of scope for this PR — tracked as a follow-up.

No template `DESCRIPTION` files were found under `minirextendr/inst/templates/` (none exist yet).

## Test plan

- [ ] `just r-cmd-build` — not run in this PR (compilation blocked in sandbox); CI will verify
- [ ] `just r-cmd-check` — not run in this PR; CI will verify
- [ ] Confirm `rpkg/DESCRIPTION` contains `Depends: R (>= 4.4)`

Closes #384

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #414, opened 2026-05-06
fix(macros): reject Self-returning constructors on vctrs-tagged impls (#248)

### LINKED ISSUE #248: vctrs: Self-returning ctor wraps ExternalPtr in vctrs::new_vctr (invalid .data)
## Problem

Surfaced while working on #208. The `DerivedCurrency` fixture in `rpkg/src/rust/vctrs_derive_example.rs` is a vctrs-tagged impl whose constructor returns `Self`:

\`\`\`rust
#[miniextendr(vctrs(kind = \"vctr\", base = \"double\", abbr = \"\$\"))]
impl DerivedCurrency {
    pub fn new(symbol: String, amounts: Vec<f64>) -> Self { ... }
}
\`\`\`

Generated R wrapper:

\`\`\`r
new_derivedcurrency <- function(symbol, amounts) {
  ...
  .val <- .Call(C_DerivedCurrency__new, .call = match.call(), symbol, amounts)
  ...
  data <- .val
  vctrs::new_vctr(data, class = \"DerivedCurrency\")
}
\`\`\`

\`.val\` here is the ExternalPtr handle to the Rust \`Self\`. Passing that to \`vctrs::new_vctr(data, ...)\` fails R-side with:

\`\`\`
Error in vctrs::new_vctr(data, class = \"...\"): \`.data\` must be a vector type.
\`\`\`

## Why it's latent

\`rpkg/tests/testthat/test-vctrs-derive.R\` only exercises \`DerivedCurrency\` through \`.Call\` directly — the generated \`new_derivedcurrency()\` R wrapper is never called. So the fixture has shipped for some time without the bug surfacing.

Verified during #208 work: calling \`new_derivedcurrency(\"\$\", c(1.0, 2.5))\` from R hits the \`.data must be a vector type\` error.

## Options

1. **Disallow \`impl ExternalPtr -> Self\` on vctrs-tagged impls**: lint or reject at macro expansion — vctrs ctors should return vector payload, not Self. This matches the fixture pattern adopted in #208/#241 (\`VctrsMatchArgScale::new(...) -> Vec<f64>\`).
2. **Auto-unwrap the ExternalPtr**: insert a codegen hook that extracts the underlying vector data from the Rust type before passing to \`new_vctr\`. Requires the type to have a well-defined \"data\" field/method. Ugly.
3. **Accept the status quo and document**: vctrs impls returning Self are meant to be used via \`.Call\` or custom R wrappers only; warn in docs.

Option 1 is cleanest. #208/#241 showed the canonical pattern: return the vector payload, let the codegen wrap it.

## Refs
- #208, #241 (vctrs match_arg fixture that surfaced this)
- \`rpkg/src/rust/vctrs_derive_example.rs:252-259\`
- \`rpkg/tests/testthat/test-vctrs-derive.R:381+\`

### LINKED ISSUE #416: test(rpkg): restore #[miniextendr(vctrs(format))] integration coverage after #248
## Context

PR #414 (closes #248) reshaped the `DerivedCurrency` fixture so its constructor returns `Vec<f64>` (the vector payload) rather than `Self`. As a side effect, the previous `#[miniextendr(vctrs(format))]` protocol-method-override demonstration was removed:

- before: `pub fn format_currency(&self) -> Vec<String>` annotated with `#[miniextendr(vctrs(format))]`
- after: `pub fn format_amounts(symbol, amounts) -> Vec<String>` (a plain static utility, no protocol annotation)

Macro-level coverage of the `vctrs(format)` codegen path is preserved in `miniextendr-macros/src/miniextendr_impl/tests.rs:814`. But `rpkg/tests/testthat/test-vctrs-derive.R` no longer exercises the protocol method end-to-end.

## Open question

It's unclear whether `vctrs(format)` with the new "constructor returns vector payload" pattern can have a sensible `&self` method at all — without a Rust `Self` instance backing the vector, `&self` doesn't have a meaningful runtime interpretation.

Two possible resolutions:
1. **If `vctrs(format)` is still usable** with the new pattern (e.g. by reconstructing state from the wrapped vector or via a static method form), add an rpkg fixture that exercises it end-to-end and asserts `format(x)` dispatches correctly.
2. **If `&self` on vctrs impls is fundamentally incompatible** with vector-payload constructors, extend MXL120 (or add MXL121) to reject `&self` / `&mut self` methods on `#[miniextendr(vctrs(...))]` impls, and document that vctrs protocol overrides must be expressed as static methods taking the unwrapped payload.

## Refs

- PR #414 (issue #248 fix)
- `miniextendr-macros/src/miniextendr_impl/tests.rs:814` — surviving macro-level coverage
- `rpkg/src/rust/vctrs_derive_example.rs` — fixture before/after #414

CGMossa:
## Summary

- Adds hard compile-time error **MXL120** when a `#[miniextendr(vctrs(...))]` impl constructor returns `Self`, the named type, `Box<Self>`, `Result<Self, _>`, or references thereof. The generated R wrapper calls `vctrs::new_vctr(.data, ...)` on the constructor result; an `EXTPTRSXP` (Rust `Self`) fails at runtime with _".data must be a vector type"_.
- Fixes the `DerivedCurrency` fixture by returning `Vec<f64>` from `new`; replaces the now-incompatible ExternalPtr-backed instance methods with a plain static `format_amounts`.
- Adds UI trybuild test `impl_vctrs_ctor_returns_self.rs` + `.stderr` snapshot. Updates all existing vctrs unit tests and insta snapshots so constructors return vector types.

## Plan

See [`plans/issue-248-vctrs-self-ctor-lint.md`](plans/issue-248-vctrs-self-ctor-lint.md).

## Test plan

- [x] `cargo test -p miniextendr-macros` — 289 passed (includes new UI compile-fail test)
- [x] `just check` — clean across all workspace members including `rpkg/src/rust`
- [x] `just clippy` — clean (collapsible_if fixed in new helpers)
- [x] `cargo clippy --workspace --all-targets --features "...full CI features..." -- -D warnings` — clean
- [x] `just devtools-test` — 4907 passed, 0 failed (new DerivedCurrency wrapper tests exercised end-to-end)
- [x] `Rscript -e 'library(miniextendr); cat("OK\n")'` — loads cleanly

## Deferred

A `miniextendr-lint` mirror of MXL120 is intentionally deferred — follow-up issue filed: closes #248.

---

## Rectification: &self method compatibility (closes #416)

**Investigation** (see `reviews/issue-248-self-method-compat.md`):

Code review on this PR surfaced issue #416: the `DerivedCurrency` fixture lost `vctrs(format)` integration coverage when `&self` methods were removed, raising the deeper question of whether instance methods can work at all on vector-payload vctrs impls.

**Finding — Option A (hard rejection)**:

`&self` / `&mut self` on a `#[miniextendr(vctrs(...))]` impl is fundamentally broken:

1. The generated R wrapper calls `.Call(C_Type__method, .call = match.call(), x)` where `x` is the vctrs S3 vector (a REALSXP/INTSXP base vector).
2. The C wrapper calls `ErasedExternalPtr::from_sexp(self_sexp)` then `.downcast_ref::<Type>().expect(...)`.
3. This always panics at runtime — the base vector has no `ExternalPtr` payload. There is no Rust `Self` stored anywhere.

No adapter pattern is feasible: for vector-payload vctrs, the R SEXP _is_ the data — there is no heap-allocated Rust struct to point at.

**Changes in rectification commits**:

- `miniextendr_impl.rs`: extends MXL120 to cover all instance-method receivers (`&self`, `&mut self`, `self`, `self: ExternalPtr<Self>`, etc.) with a clear diagnostic.
- `vctrs_class.rs`: extends the static-methods codegen to honour `vctrs_protocol` on static methods, emitting them as `format.<Class>` S3 methods rather than `<class>_<method>` utility functions. This restores the `vctrs(format)` use-case via the correct static-method pattern.
- `tests.rs`: removes all `&self` from vctrs fixtures; adds a `vctrs_protocol_method_override` assertion that static-method-with-`vctrs(format)` generates the correct `format.Currency` S3 method.
- `snapshots/`: regenerates `snapshot_vctrs_vctr` and `snapshot_vctrs_rcrd` with static helpers.
- `tests/ui/impl_vctrs_self_method_rejected.rs` + `.stderr`: new UI test for the new MXL120 case.
- `docs/VCTRS.md`: removes the outdated `-> Self` constructor + `&self` example; documents the static-only rule and the correct `vctrs(format)` usage pattern.


CGMossa:

**Verdict: looks good — ready for human merge review. One follow-up filed.**

### Proc-macro change

- `vctrs_ctor_returns_self_or_type` + recursive `ty_is_self_or_named` covers the requested cases: `Self`, named impl type, `&Self`, `&mut Self`, `Box<Self>`, `Result<Self, _>`, and arbitrary nesting (e.g. `Result<Box<Self>, _>` works through recursion).
- The `is_ctor` predicate in the lint matches the existing class-system-aware `is_method_constructor` on main (vctrs branch returns true for any `pub fn new()` with no receiver, regardless of return type). So the lint scope lines up exactly with what codegen will try to wrap with `vctrs::new_vctr()`. No false negatives.
- Error message references MXL120 by name and offers concrete fixes for each vctrs kind (`vctr` → `Vec<T>`, `rcrd` → named-list, `list_of` → `Vec<Vec<T>>`).

### UI test
`impl_vctrs_ctor_returns_self.rs` + `.stderr` snapshot are minimal and self-explanatory. Span points at `-> Self` correctly.

### Snapshot drift (`@export` on instance methods)
The two `.snap` updates adding `#' @export` to `numerator.Rational` / `value.Percent` / `scale.Percent` are pre-existing drift from commit `8dcaa652` ("fix(macros): emit @export on every S3 method codegen path"), which evidently shipped without re-running these particular vctrs full-snapshot tests. Absorbing them here is fine per CLAUDE.md's "fix warnings you see".

### Fixture rewrite — note
`DerivedCurrency` lost its previous demonstration of `#[miniextendr(vctrs(format))]` protocol-method override (the old `format_currency(&self)` annotation). Macro-level snapshot in `tests.rs:814` still covers expansion, so the codegen path isn't dark, but rpkg integration coverage of `vctrs(format)` is gone.

Filed as a follow-up: <will be filled in below>

The new tests exercise `vec_is`, `vec_size`, `vec_data`, `vec_ptype_abbr` end-to-end via `new_derivedcurrency()`, plus `derivedcurrency_format_amounts()` as a static utility. Net coverage is *better* for what we actually claim to test (the wrapper path), even with the protocol-override regression.

### Verification
Agent reports clean: `cargo test -p miniextendr-macros` (289 passed including new UI), `just check`, `just clippy` (with full CI feature set), `just devtools-test` (4907 passed). Cargo.lock churn is the same benign monorepo source-shape drift seen in #410/#412/#413.

### Deferred follow-ups
- #415 — miniextendr-lint mirror of MXL120
- (filing alongside this review) — restore `vctrs(format)` integration coverage in rpkg

CGMossa:
Follow-up filed: https://github.com/A2-ai/miniextendr/issues/416 (restore vctrs(format) integration coverage in rpkg).

CGMossa:
## Review (rectification — Claude Opus 4.7, automated)

**Verdict: rectification looks good — ready for human merge review.**

### Investigation depth

`reviews/issue-248-self-method-compat.md` is solid: traces the broken path through `vctrs_class.rs:300` → `.Call(C_Type__method, ..., x)` → `miniextendr_impl.rs:2426-2471` → `ErasedExternalPtr::from_sexp(self_sexp).downcast_ref::<T>().expect(...)`. Confirms it's a loud panic at runtime, not UB. Considers and rejects three potential adapter designs with reasons. Recommendation (Option A) is well-justified.

### MXL120 extension (`miniextendr_impl.rs:2170-2204`)

Covers all 6 instance receiver kinds via `ReceiverKind::is_instance()`. Error message includes a concrete before/after example for the user. Spans on the method ident, which renders cleanly in the UI test stderr.

### Static-method protocol codegen (`vctrs_class.rs:310-348`)

Real win — the agent went beyond just blocking the broken pattern and made `vctrs(format)` actually usable via the static-method form. A static method annotated with `#[miniextendr(vctrs(format))]` now emits as `format.Class <- function(...)` with `#' @method format Class` for proper roxygen S3 method registration.

The if/else has some duplication between protocol and non-protocol branches (both build `MethodDocBuilder` similarly); could be collapsed but it's fine for this PR.

### Test coverage

- New UI test `impl_vctrs_self_method_rejected.rs` + `.stderr` snapshot — minimal, clean diagnostic.
- All vctrs fixtures in `tests.rs` updated to the static-method shape.
- `vctrs_protocol_method_override` test now asserts `format.Currency <- function(amounts)` is generated (and that the non-protocol `symbol` method generates as `currency_symbol` regular function).
- Snapshots for `vctrs_vctr` and `vctrs_rcrd` regenerated to reflect the static-helper shape.

### Concern (filed as follow-up #418)

The generated `format.Currency <- function(amounts)` wrapper doesn't accept `...`. So `format(x, nsmall = 2)` will hit "unused argument (nsmall = 2)". That's a real ergonomic gap — users invoking format/print/etc. with extra args will get a confusing error. Not a blocker for this PR (the core `format(x)` case works), but worth tracking.

Filed as #418.

### Merge-order concern

`fix/issue-248-vctrs-self-ctor-rejected` (this PR) and `fix/issue-343-multi-paragraph-details` (PR #417) both touch the same vctrs snapshot files but in opposing ways:
- #417 adds `#' @export` to S3 method wrappers like `value.Percent` / `scale.Percent` / `numerator.Rational` (snapshot drift catch-up).
- This PR's rectification *removes* those S3 method wrappers entirely (replaced by static helpers `percent_value` / `percent_scale` / `rational_numerator`).

If #414 merges first, #417 will need to drop its snapshot delta on rebase (the wrappers being decorated no longer exist). If #417 merges first, this PR's snapshot regeneration absorbs the @export additions before deleting the methods — no extra work needed.

Recommend landing #414 first.

### Verification

Agent's report indicates clean run of `cargo test`, `just check`, `just clippy` (full CI features), and `just devtools-test`. CI on the new commits is re-running — Rust lint already green; R CMD check pending.

CGMossa:
Flagging the CI failure as flaky-looking: R tests on Linux ran 3 heap-checked iterations all clean (`FAIL 0 | WARN 0 | SKIP 15 | PASS 4907`) but the job hit a 33-minute cancellation, likely a timeout on the third iteration. The umbrella "CI Success" check failed because it counts "cancelled" as non-success. Recommend a retrigger before merge — no test regression introduced by the rectification commits.

PR #419, opened 2026-05-06
fix(codegen): deduplicate S7/S4 generic aliases via shared Rd block (#59)

CGMossa:
## Summary

R CMD check on main currently reports two non-#59 warnings (the originally-reported "duplicated alias" warnings for `get_value` / `s4_get_value` / `value` are no longer firing — they were silently mitigated by the class-qualified `@name` workaround in `s7_class.rs:619` and `s4_class.rs:138`, plus the `rpkg/R/generics.R` stub for S3 generics). Two real warnings remain:

1. **`S7Fahrenheit.Rd` — "Objects in \\usage without \\alias: 'convert'"**: the S7 `convert_from`/`convert_to` codegen emitted `@name convert-{From}-to-{To}` + `@rdname {ClassName}` but no `@aliases convert`, leaving the merged Rd file with `convert(...)` in `\\usage{}` but no matching `\\alias{convert}`.
2. **`S3Raiser.Rd` — "Undocumented arguments: 'id'"**: `S3Raiser::new(id: i32)` had no doc comment for `id`.

## Fixes

1. `miniextendr-macros/src/miniextendr_impl/s7_class.rs`: emit `#' @aliases convert` on both `convert_from` and `convert_to` codegen paths (8 lines added across the two emission sites).
2. `rpkg/src/rust/condition_class_system_tests.rs`: add `/// @param id Numeric identifier for this raiser instance.` to `S3Raiser::new`.
3. `rpkg/man/S7Fahrenheit.Rd`: surgical 1-line addition (`\\alias{convert}`) — manually applied rather than via `devtools::document` regen to avoid roxygen2 version-drift noise across 70 unrelated Rd files.
4. `rpkg/man/S3Raiser.Rd`: surgical 2-line addition (`\\item{id}{...}` + blank).
5. `rpkg/R/miniextendr-wrappers.R`: regenerated by the cdylib (not roxygen) to reflect the new source-side doc tags. 87-line diff is mostly line-number drift from the new `/// @param id` doc line.

## Plan / investigation

- `plans/issue-59-s7-s4-generic-alias-dedup.md`
- `reviews/issue-59-alias-dedup-investigation.md` — Phase 1 investigation showing the original warnings are no longer present.

## Note on the rectification commit

[...]

Issue #318, opened 2026-04-25
Demo fixture `call_attribution_demo` is exported into the public namespace


CGMossa:
## Goal

`rpkg/src/rust/call_attribution_demo.rs` defines two demo functions (`call_attr_with` and `C_call_attr_without`) that were added in PR #316 as runnable fixtures for `docs/CALL_ATTRIBUTION.md`. They carry `@keywords internal` but are **not** omitted from the NAMESPACE, causing `man/call_attribution_demo.Rd` to be installed and both R symbols to be discoverable via `getNamespaceExports`. The fix is **Option 1**: keep the Rust source as-is (the doc already uses `:::` throughout), add `#[miniextendr(no_export)]` (or equivalent) to suppress `@export` from the generated wrapper, then regen the generated files (`NAMESPACE`, `R/miniextendr-wrappers.R`, `man/`) so the functions are truly internal. A new testthat test in `rpkg/tests/testthat/test-call-attribution.R` verifies the `conditionCall` behaviour that the doc describes, so the transcript remains machine-verifiable.

> **Why Option 1 over Option 3?** `docs/CALL_ATTRIBUTION.md` already uses `miniextendr:::` throughout (lines 57, 61, 69, 74, 83–84). The doc transcript is already correctly written for internal-only access. Converting the fixtures to a proper testthat test file (Option 3) would duplicate logic but not eliminate the Rust source; Option 1 only requires adding `no_export` to the two `#[miniextendr]` attrs and regenerating — the smallest possible diff. Option 3 is additive (a new test file is still the right way to make the transcript verifiable), so the plan includes both: unexport via `no_export` AND add a new test file.

---

## Constraints (from CLAUDE.md and rpkg/CLAUDE.md)

- **No backwards-compat shims**: these fixtures have no external callers — remove the exports cleanly.
- **Fix warnings you see**: if `just devtools-document` emits any roxygen2 diagnostics, fix them; don't silence with `@noRd`/`@keywords internal` hacks.
- **No `mod.rs`**: don't introduce `call_attribution_demo/mod.rs` — file stays as-is.
- **No `// =====` section dividers**: use `// region:` / `// endregion` if new sections are added.
[...]

PR #421, opened 2026-05-06
feat(s7): emit @prop roxygen2 tags for S7 sidecar and impl-block properties

### LINKED ISSUE #379: S7 @prop emission: sidecar (r_data_accessors) properties are skipped (follow-up to #377)
## Background

PR #377 added `@prop` tag emission for non-constructor S7 properties in the impl block (getter/setter/validator-defined). It does not, however, cover **sidecar properties** that come from `#[miniextendr(s7, r_data_accessors)]` + the corresponding `#[derive(ExternalPtr)]` data accessors.

## Symptom

In `miniextendr-macros/src/miniextendr_impl/s7_class.rs`, the new emission loop iterates `properties.values()` (the impl-block-defined property map). Sidecar properties are appended later via:

```r
properties = c(list(...), .rdata_properties_<Type>)
```

i.e., they live in a separate runtime list built by the `ExternalPtr` derive macro, and S7 codegen has no static visibility into their names. So sidecar-only properties remain undocumented even though the proximate comment in PR #377 says they're handled.

## Why non-trivial

Sidecar property names come from a derived `#[derive(ExternalPtr)]` struct's `#[r_data_accessors]` fields. Wiring those into the S7 class roxygen block requires either:

1. Coordinating across two macro passes (the derive macro stamps a const list, and the S7 codegen reads it at write time — but that's a runtime-only signal, not compile-time).
2. Re-parsing the field list in the S7 codegen path (would mean S7 codegen has to inspect the struct's `#[derive(ExternalPtr)]` attribute) — duplicating logic.
3. Adding a new attribute on the struct for documentation strings, distinct from the `#[r_data_accessors]` setup.

## Proposed work

- Decide on the wiring model (probably option 3: a simple `#[miniextendr(s7(prop_doc = "..."))]` attribute on the struct fields, alongside the existing accessor-generation machinery).
- Add a fixture in `rpkg` exercising a sidecar property that would benefit from `@prop`.
- Verify the rendered Rd shows the sidecar property in the `\\section{Additional properties}` block.

## NITs from the same review (minor, lower priority)

- Constructor-formal parsing in `s7_class.rs` is string-fragile (`ctx.params.split(\", \")` then `split('=')`) — works for the documented edge cases (varargs, `.ptr`, simple defaults) but brittle for defaults containing `\", \"` substrings. Pre-existing pattern; not regressed by #377 but now also drives `@prop` vs `@param` classification. Could be replaced with a structured params iterator.
- The new `snapshot_s7_prop_tags` test only covers the happy path. Could add coverage for default-valued constructor params and varargs.

Reference: codex review on PR #377.

CGMossa:
## Summary

- Adds `#' @prop` roxygen2 tag emission for S7 class properties, covering both impl-block getter/setter pairs and `#[r_data]` sidecar fields (closes #379).
- Sidecar props use a write-time placeholder (`.__MX_S7_SIDECAR_PROP_DOCS_<Type>__`) substituted from a new `MX_S7_SIDECAR_PROPS` linkme distributed_slice, identical in pattern to `MX_MATCH_ARG_PARAM_DOCS`.
- Users add `#[r_data(prop_doc = "...")]` to document sidecar fields; omitting `prop_doc` emits `"(undocumented sidecar property)"`.

## Key design notes

- `generate_s7_r_wrapper` runs at proc-macro time so cannot do a live registry lookup; write-time substitution (the same pattern as `MX_MATCH_ARG_PARAM_DOCS`) is the correct approach.
- The `@prop` tag requires roxygen2 ≥ 8.0.0. roxygen2 7.3.3 warns `@prop is not a known tag` — expected and intentional. The DESCRIPTION already declares `Config/roxygen2/version: 8.0.0`.
- Doc text is extracted from getter doc comments by stripping the `@description ` / `@title ` prefix that `roxygen_tags_from_attrs_for_r6_method` adds automatically.

## Changes

| File | Change |
|---|---|
| `miniextendr-api/src/registry.rs` | `MX_S7_SIDECAR_PROPS` slice + `SidecarPropEntry` struct; write-time placeholder substitution |
| `miniextendr-macros/src/externalptr_derive.rs` | Parse `#[r_data(prop_doc = "...")]`; emit distributed_slice entries for S7 class systems |
| `miniextendr-macros/src/miniextendr_impl/s7_class.rs` | Emit `@prop` for impl-block getters; emit sidecar placeholder when `r_data_accessors` |
| `rpkg/src/rust/rdata_sidecar_tests.rs` | Add `prop_doc` attributes to `SidecarS7` fields |
| Snapshots (4 new, 3 updated) | `snapshot_s7_documented_props`, `snapshot_s7_sidecar_only_props`, `snapshot_s7_sidecar_and_impl_props`, `snapshot_s7_prop_tags_with_defaults_and_varargs`; fixed pre-existing vctrs missing `@export` |
| `rpkg/R/miniextendr-wrappers.R`, `rpkg/man/` | Regenerated — `@prop` lines added, plus standard roxygen2 7.3.3 `\docType{data}` churn on R6 classes |

## Before / after

**Before** (`SidecarS7` wrapper):
```r
#' @title SidecarS7 Sidecar Accessors
#' @param .ptr Internal pointer (used by static methods, not for direct use).
SidecarS7 <- S7::new_class("SidecarS7", ...)
```

**After**:
```r
#' @title SidecarS7 Sidecar Accessors
#' @param .ptr Internal pointer (used by static methods, not for direct use).
#' @prop prop_int An integer sidecar property.
#' @prop prop_flag A logical sidecar property.
#' @prop prop_name A character sidecar property.
SidecarS7 <- S7::new_class("SidecarS7", ...)
```

## Test plan

- [x] `just test` — all suites pass (293 passed, 66 ignored)
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean
- [x] `cargo clippy --workspace --all-targets --locked --features <all CI features> -- -D warnings` — clean
- [x] `just devtools-document` — exits 0 (roxygen2 7.3.3 warns `@prop is not a known tag` — expected for pre-8.0.0 versions)
- [x] 4 new snapshot tests for the `@prop` emission paths
- [x] `SidecarS7.Rd` contains prop accessor aliases and argument documentation


CGMossa:

**Verdict: rectified — ready for human merge review, with one integration-coverage caveat.**

### Rectification (commit `a...`)

The agent's initial commit bundled the source-side feature work with regenerated `.Rd` files across ~70 man pages. Those changes were pure roxygen2 7.3.3 vs 8.0.0 cosmetic drift — same trap as #318 / #419. `S7Fahrenheit.Rd` also regressed (lost the S7 `convert` method `\\usage{}` blocks) because 7.3.3 renders that differently from 8.0.0 and the project pins 8.0.0 via `Config/roxygen2/version`.

Restored all `rpkg/man/*.Rd` to main's content; restored `rpkg/src/rust/Cargo.lock` (benign source-shape drift the agent's run also produced).

### Source-side feature (preserved unchanged)

- `miniextendr-api/src/registry.rs` (+69): `SidecarPropEntry` struct + `MX_S7_SIDECAR_PROPS` distributed_slice + write-time placeholder substitution.
- `miniextendr-macros/src/externalptr_derive.rs` (+80): `parse_r_data_prop_doc` for `#[r_data(prop_doc = "...")]`. (Note: agent used `#[r_data]` not `#[r_data_accessors]` from the plan — that's correct, matches the project's actual attribute name.)
- `miniextendr-macros/src/miniextendr_impl/s7_class.rs` (+35): `@prop` emission for impl-block getters and sidecar fields.
- 6 macro snapshot tests covering: sidecar-only props, sidecar+impl props, default missing-doc text, defaults+varargs (the in-scope NIT), documented impl props, plus updates to existing vctrs snapshots.
- `rpkg/src/rust/rdata_sidecar_tests.rs`: `prop_doc` annotations on SidecarS7 fields.
- `rpkg/R/miniextendr-wrappers.R` (+10): cdylib regen for the new prop_doc data — legitimate.

### Integration coverage caveat

The agent flagged this themselves: `@prop` is roxygen2 8.0.0-only. The agent's local roxygen2 was 7.3.3, which silently drops unknown tags. So the regenerated `SidecarS7.Rd` doesn't actually demonstrate the new feature working end-to-end (the `@prop` lines are produced by codegen but stripped by 7.3.3 before reaching `.Rd`). Macro snapshot tests verify the codegen produces the right roxygen tags; rendering them through 8.0.0 is downstream of installing the matching toolchain.

This is acceptable for merging the codegen change — the snapshots are the contract — but the integration verification (`?SidecarS7` showing the sidecar props in R) requires roxygen2 8.0.0 to run.

### Net diff vs main

15 files, 554 insertions / 3 deletions. No `.Rd` churn. No `DESCRIPTION` change. No `Cargo.lock` noise.

EOF

PR #423, opened 2026-05-06
fix(macros): vctrs protocol-method R wrappers accept ... for S3 dispatch (#418)

### LINKED ISSUE #418: feat(vctrs): protocol-method R wrappers should accept ... for S3 dispatch passthrough
## Context

PR #414's rectification (commit 650c4338) added codegen for `#[miniextendr(vctrs(<protocol>))]` on static methods, emitting them as `<protocol>.<Class>` S3 methods. Example: `#[miniextendr(vctrs(format))]` on `pub fn format_currency(amounts: Vec<f64>) -> Vec<String>` generates:

```r
format.Currency <- function(amounts) {
  .val <- .Call(C_Currency__format_currency, .call = match.call(), amounts)
  ...
  .val
}
```

## Concern

When R calls `format(x, ...)` (e.g. `format(x, nsmall = 2)`), S3 dispatch evaluates `format.Currency(x, nsmall = 2)`. The generated wrapper has signature `function(amounts)` — no `...`. R will error:

```
Error in format.Currency(x, nsmall = 2) : unused argument (nsmall = 2)
```

This breaks compatibility with idiomatic R format/print/etc. usage, where extra args are conventional.

## Possible fixes

1. **Append `...` to protocol-method R wrappers** — codegen appends `, ...` to the R formals and silently discards extras. Lossy but compatible with S3 conventions.
2. **Pass `...` through to the Rust call** — would need the Rust function to accept variadic args (not natural in Rust).
3. **Generate a wrapper layer** — `format.Currency <- function(x, ...) { invisible(...); .Call(C_..., x) }` — explicit drop, no surprise.

Option 3 is probably best: dropped args are explicit and documented; the method behaves correctly for the args the Rust side actually understands.

## Refs

- PR #414, commit 650c4338
- `miniextendr-macros/src/miniextendr_impl/vctrs_class.rs:312-345` — the protocol-method codegen branch

CGMossa:
## Summary

- Appends `, ...` to the R formals of protocol-method wrappers generated by `#[miniextendr(vctrs(<protocol>))]` on static methods in `#[miniextendr(vctrs(...))] impl` blocks.
- Non-protocol static helpers keep their existing fixed-formals shape (no change).
- `...` is silently dropped — the underlying Rust function has a fixed signature.

## Problem

When `format.Currency <- function(amounts)` is dispatched via `format(x, nsmall = 2)`, R errors with "unused argument (nsmall = 2)". This is a standard S3 dispatch pattern — all format methods must accept `...`.

## Fix

In `vctrs_class.rs`, the static-method codegen now emits:

```r
format.Currency <- function(amounts, ...) {
  .val <- .Call(C_Currency__format_currency, .call = match.call(), amounts)
  ...
}
```

## Tests

- Updated `vctrs_protocol_method_override` assertion: expects `function(amounts, ...)`.
- Added negative assertion: `currency_symbol <- function(amounts, ...)` must NOT appear (non-protocol helpers keep fixed formals).
- All 289 `miniextendr-macros` tests pass.

## Dependency on PR #414

This branch is based on `fix/issue-248-vctrs-self-ctor-rejected` (PR #414), which introduced the `is_protocol` branch for static-method codegen. The commit `650c4338` from PR #414 is the direct parent. When #414 merges to main, this branch should be rebased onto main before merging.

## Test plan

- [x] `cargo test -p miniextendr-macros` — 289 passed.
- [x] `just check` — clean.
- [x] `just clippy` — clean.
- [ ] After #414 merges: rebase onto main, re-run tests, then merge.

Closes #418.

### DIFF
--- a/miniextendr-macros/src/miniextendr_impl/tests.rs
+++ b/miniextendr-macros/src/miniextendr_impl/tests.rs
@@ -828,13 +828,16 @@ fn vctrs_protocol_method_override() {
 
     // format_currency method should be generated as format.Currency, not format_currency.Currency
     assert!(wrapper.contains("#' @method format Currency"));
-    assert!(wrapper.contains("format.Currency <- function(amounts)"));
+    // Protocol methods get a trailing `...` so `format(x, nsmall = 2)` doesn't error
+    // with "unused argument (nsmall = 2)" when R dispatches to format.Currency.
+    assert!(wrapper.contains("format.Currency <- function(amounts, ...)"));
 
     // Should NOT create a new S3 generic for "format" (it's a base R function)
     assert!(!wrapper.contains("format <- function(x, ...) UseMethod(\"format\")"));
 
-    // symbol static helper should be generated as regular function currency_symbol(amounts)
+    // symbol static helper (non-protocol) should keep fixed formals — no trailing `...`
     assert!(wrapper.contains("currency_symbol <- function(amounts)"));
+    assert!(!wrapper.contains("currency_symbol <- function(amounts, ...)"));
 }
 // endregion
 
--- a/miniextendr-macros/src/miniextendr_impl/vctrs_class.rs
+++ b/miniextendr-macros/src/miniextendr_impl/vctrs_class.rs
@@ -348,7 +348,15 @@ pub fn generate_vctrs_r_wrapper(parsed_impl: &ParsedImpl) -> String {
             lines.extend(method_doc.build());
         }
 
-        lines.push(format!("{} <- function({}) {{", fn_name, ctx.params));
+        // Protocol methods accept `...` so `format(x, nsmall = 2)` and similar
+        // S3 dispatch calls with extra arguments don't error with "unused argument".
+        // The `...` is silently dropped; the underlying Rust function has a fixed signature.
+        let formals = if is_protocol {
+            format!("{}, ...", ctx.params)
+        } else {
+            ctx.params.to_string()
+        };
+        lines.push(format!("{} <- function({}) {{", fn_name, formals));
 
         // Inject r_entry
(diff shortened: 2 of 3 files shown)


CGMossa:

**Verdict: looks good — ready for human merge review (after #414).**

### Source diff
9-line change in `vctrs_class.rs`: when `is_protocol` is true (the static method has `#[miniextendr(vctrs(<protocol>))]`), append `, ...` to the R formals; otherwise keep the existing fixed-formals shape. Clean conditional, good comment explaining why ("The `...` is silently dropped; the underlying Rust function has a fixed signature").

### Test coverage
`vctrs_protocol_method_override`:
- Updated positive assertion to expect `format.Currency <- function(amounts, ...)`.
- Existing assertion that no `format` S3 generic gets created (since `format` is base R) is preserved.
- Updated assertion that the non-protocol `currency_symbol` still has `function(amounts)` (no `...`).
- **Added negative assertion** that `currency_symbol` is NOT `function(amounts, ...)` — locks in the protocol-only behavior.

That's exactly the contract: protocol methods get `...`, regular static helpers don't. Snapshot tests pass per the agent's run (`cargo test -p miniextendr-macros` 289 passing).

### Base branch (just fixed)

I retargeted the PR base from `main` to `fix/issue-248-vctrs-self-ctor-rejected` so the diff renders as the 60-line change this PR introduces, not as #414's full content + this change. After #414 merges to main, this PR will need a rebase onto main.

### Skipped
`just rcmdinstall` failed in the agent's worktree due to missing R packages (cli, lifecycle, R6, S7, vctrs) in the rv library — environment issue, unrelated to the change. The macro snapshot tests directly exercise the codegen and confirm the fix; rendered `R/miniextendr-wrappers.R` regen is downstream.

Closes #418.

PR #425, opened 2026-05-07
test(encoding): exhaustive coverage for package_init UTF-8 locale gate

CGMossa:
## Summary

Adds tests for the load-time UTF-8 locale assertion in `package_init`
(`miniextendr-api/src/init.rs` → `encoding::miniextendr_assert_utf8_locale`).
Today the assertion has zero coverage; `tests/testthat/test-encoding.R`
only checked that `encoding_info_available()` returned a logical.

- New `#[miniextendr]` fixture `assert_utf8_locale_now()` lets R-side tests
  re-invoke the assertion after `Sys.setlocale()` flips `LC_CTYPE`.
- `tests/testthat/test-encoding.R` now covers both the correct and
  incorrect locale paths.

## Coverage

**Correct locale (success):**
- baseline UTF-8 session — `assert_utf8_locale_now()` returns `TRUE`
- idempotency (5x)
- `library(miniextendr)` in subprocess under `LC_ALL=C.UTF-8` / `en_US.UTF-8`
- `library(miniextendr)` under `LANG=C` + `LC_CTYPE=en_US.UTF-8`
  (verifies `LC_CTYPE` wins over `LANG`)

**Incorrect locale (must error):**
- mid-session `Sys.setlocale(\"LC_CTYPE\", \"C\")` → assertion errors
- mid-session `Sys.setlocale(\"LC_CTYPE\", \"<latin-1 candidate>\")` → errors
- restore `LC_CTYPE` to UTF-8 → assertion succeeds again
- subprocess `library(miniextendr)` under `LC_ALL=C` → load fails with
  \"UTF-8 locale\" message
- subprocess `library(miniextendr)` under `LC_ALL=POSIX` → load fails

Each non-UTF-8 test guards on `l10n_info()[['UTF-8']]` actually flipping
in the spawned/in-process state, so platforms that refuse the locale
(or where glibc keeps a UTF-8 default under `LC_ALL=C`) skip the
assertion cleanly instead of false-failing.

Subprocess tests `skip_on_os(\"windows\")` to match the existing
cross-session ALTREP test policy (Windows `processx` orphan handles).

## Local results

`just devtools-test encoding` → `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 19 ]`
on macOS arm64, R 4.5.

## Why draft

Opened as draft for review of the coverage shape before extending the
same pattern to other `package_init` steps (panic hook install, ALTREP
registration, mx_abi). Not for merge as-is — please weigh in on whether
this is the right place / shape for these tests.

## Test plan

- [ ] CI green on Linux R 4.3 / 4.4 / 4.5 / 4.6 + macOS R 4.6
- [ ] Confirm subprocess tests don't hang under processx on the runner
- [ ] Verify a Latin-1 locale is actually present on at least one CI image
      (otherwise the Latin-1 test always skips, which is fine but worth
      knowing)


CGMossa:
this is tests so we do not have to worry about the alignment between R and the encoding state.

PR #438, opened 2026-05-08
fix(configure): place CARGO_TARGET_DIR outside the package source tree

CGMossa:
## Summary

Closes the same problem #436 was solving (forbidden-symbol WARNING from Rust runtime in `R CMD check --as-cran`) without needing a `strip` step at install time.

R CMD check's compiled-code phase calls `tools:::.shlib_objects_symbol_tables`, which from `src/` does:

```r
list.files("..", recursive = TRUE, pattern = "[.]a$")
```

and saves each scanned `.a`'s symbol table to `libs/symbols.rds`. At check time, `compare()` (`sotools.R:861`, comment: *"Drop the DLL symbols not in any object"*) keeps a `.so` symbol on the bad list only if it ALSO appears in some scanned object.

Default `CARGO_TARGET_DIR` was `<rpkg>/rust-target`, so `<rpkg>/rust-target/release/libminiextendr.a` sat inside `..`'s scan tree. Rust's panic / alloc / unwind runtime references libc's `_exit`, `abort`, `exit`; those names landed in `symbols.rds` attributed to the staticlib and intersected with the `.so`'s symtab → `compiled code … WARNING`.

## Fix

Move `CARGO_TARGET_DIR` to `$abs_top_srcdir/../rust-target-$PACKAGE_NAME`. The staticlib now lives outside the scan tree.

- **Tarball install** in `$TMPDIR/<pkg>-<ver>/` → target at `$TMPDIR/rust-target-<pkg>/`. The `$PACKAGE_NAME` suffix prevents parallel CRAN-style installs from clashing on shared `$TMPDIR`.
- **Monorepo dev** → target at `<repo>/rust-target-miniextendr/`, alongside (not inside) `rpkg/`. Cargo's incremental cache survives across `just rcmdinstall` runs — no slowdown.
- Existing tarball-mode cleanup (`Makevars.in:54-63`) still removes `$(CARGO_TARGET_DIR)` to keep the installed package small. Source-mode dev preserves the cache.

Mirrored to `minirextendr/inst/templates/rpkg/configure.ac`; `patches/templates.patch` refreshed via `just templates-approve`. `.gitignore` learns `rust-target-*/`.

## Test plan

- [x] `just rcmdinstall` — builds cleanly with new path; `rust-target-miniextendr/` populated at monorepo root.
- [x] `just devtools-check` (Apple clang 21.0.0 / R 4.5.2): `* checking compiled code ... OK` (was WARNING).
- [ ] CI on Linux/Windows. The fix is platform-agnostic — Linux is where the WARNING actually fires; macOS already had it `OK`.
- [ ] Verify CRAN-style `R CMD check --as-cran <tarball>` on Linux still reports `compiled code … OK`.

## Supersedes

Obsoletes #436 — no `strip --strip-symbol` step needed. Closing #436 with a pointer here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@ rpkg/src/rust/.cargo/
 
 # Cargo build artifacts (workspaces in repo root and subdirectories)
 target/
+# Per-package CARGO_TARGET_DIR set by rpkg/configure.ac (lives outside the
+# package source tree so R CMD check's compiled-code phase doesn't scan it).
+rust-target-*/
 
 # R package build artifacts
 *.Rcheck/
--- a/minirextendr/inst/templates/rpkg/configure.ac
+++ b/minirextendr/inst/templates/rpkg/configure.ac
@@ -146,15 +146,32 @@ dnl Convenience: cargo command that includes optional toolchain selector
 CARGO_CMD="$CARGO $RUST_TOOLCHAIN"
 
 dnl ---- Cargo target/profile configuration ----
-dnl IMPORTANT: target directory MUST be outside src/ to avoid pkgbuild scanning
-dnl generated .rs files in target/build/.../out/ during needs_compile() checks.
-AC_ARG_VAR([CARGO_TARGET_DIR], [Cargo target directory (default: rust-target, outside src/)])
-: ${CARGO_TARGET_DIR="$abs_top_srcdir/rust-target"}
+dnl IMPORTANT: target directory MUST be outside the package source tree.
+dnl Two reasons:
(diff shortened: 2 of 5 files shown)


CGMossa:
Walking this back — the relocation is unnecessary. The existing tarball-mode cleanup (Makevars.in:54-63) already deletes `` before `make symbols.rds` runs, which makes `file.exists(pkglibs)` return FALSE in `tools:::.shlib_objects_symbol_tables`, so the staticlib never reaches symbols.rds and `compiled code ... OK` ships without any strip step. `<rpkg>/rust-target/` is the right location. PR #436 is also unnecessary for the same reason.

PR #464, opened 2026-05-09
feat(api): IntoR for Vec<Option<C>> — fixes opaque container enum align (#461)

### LINKED ISSUE #461: DataFrameRow enum align: opaque container fields fail to compile (no IntoR for Vec<Option<C>>)
## Symptom

Deriving `DataFrameRow` on an enum where any variant has an opaque container field — `Vec<T>`, `HashSet<T>`, `BTreeSet<T>` (and presumably `HashMap`/`BTreeMap` once #457 lands) — fails to compile:

```text
error[E0277]: the trait bound `Vec<Option<Vec<i32>>>: miniextendr_api::IntoR` is not satisfied
   --> dataframe_enum_payload_matrix.rs:27:24
    |
27  | #[derive(Clone, Debug, DataFrameRow)]
```

Reproduction: any enum variant with `field: Vec<T>` / `HashSet<T>` / `BTreeSet<T>` *without* `#[dataframe(expand)]` or `#[dataframe(width = N)]`. With `expand` / `width`, codegen emits per-element scalar columns and skips opaque `Vec<Option<C>>`.

## Root cause

`miniextendr-macros/src/dataframe_derive/enum_expansion.rs` builds a column registry where every column is `Vec<Option<T>>` so non-payload variants can NA-fill. Combined with opaque container element types, this requires `IntoR` for `Vec<Option<Vec<T>>>`, `Vec<Option<HashSet<T>>>`, etc. — impls that don't exist in `miniextendr-api/src/into_r.rs` (only `Vec<C>` exists, not `Vec<Option<C>>`).

## Options

1. Add blanket `IntoR for Vec<Option<C: IntoR>>` mapping `Some(c)` → R list element, `None` → `R_NilValue`. Mirrors how scalar `Vec<Option<T>>` already lands as a typed R vector with NAs.
2. Have the enum derive special-case opaque-container columns: emit a list-column where empty/absent rows are `NULL`. This is what `expand`/`width` already do for scalar element types, but the opaque path needs distinct codegen.
3. Reject opaque containers in enum variants at derive time with an MXLxxx pointing the user at `#[dataframe(expand)]` or `#[dataframe(as_list)]`.

Option 1 is the minimum-friction fix.

## Acceptance

- Opaque-container variant fields compile.
- Both `to_dataframe` (NA-fill) and `to_dataframe_split` (per-variant) round-trip on a fixture covering `Vec<i32>`, `HashSet<String>`, `BTreeSet<i32>` enum variants.
- Add the blocked fixtures in `rpkg/src/rust/dataframe_enum_payload_matrix.rs` once unblocked (the Vec opaque / HashSet / BTreeSet sections were dropped from PR for `tests/dataframe-enum-payload-matrix` for this reason).

CGMossa:
## Summary

- Fixes #461: `Vec<Option<C>>` for container types was missing, causing opaque container fields (`Vec<T>`, `HashSet<T>`, `BTreeSet<T>`) to fail compilation on the `align` path of `#[derive(DataFrameRow)]` enums.
- Adds `vec_option_of_into_r_to_list` helper in `miniextendr-api/src/into_r.rs` and 8 concrete `IntoR` impls: `Vec<Option<Vec<T>>>`, `Vec<Option<Vec<String>>>`, `Vec<Option<HashSet<T>>>`, `Vec<Option<HashSet<String>>>`, `Vec<Option<BTreeSet<T>>>`, `Vec<Option<BTreeSet<String>>>`, `Vec<Option<HashMap<String,V>>>`, `Vec<Option<BTreeMap<String,V>>>`.
- Re-adds the three previously-removed fixture sections in `rpkg/src/rust/dataframe_enum_payload_matrix.rs`: Vec opaque (`Vec<i32>`), `HashSet<String>`, and `BTreeSet<i32>` — each with both `align` and `split` fixtures.
- Adds corresponding R tests in `test-dataframe-enum-payload-matrix.R` covering the 5-cell matrix (NvNr align + NvNr split for each shape).
- Adds `gc_stress_vec_option_collection()` no-arg fixture in `gc_stress_fixtures.rs` to exercise the new `OwnedProtect` allocation path under GC pressure.

References plan file: `plans/issue-461-vec-option-intor.md`

## Test plan

- [x] `just check` — clean
- [x] `just clippy` (clippy_default) — clean
- [x] CI clippy_all (full feature flags) — clean
- [x] `just configure && just rcmdinstall && just devtools-document` — clean
- [x] `just devtools-test` — 5217 passed, 0 failed
- [x] New opaque Vec, HashSet, BTreeSet fixtures compile and pass
- [x] Existing scalar `Vec<Option<T>>` impls unaffected (all prior tests still pass)


CGMossa (commented):
## Review of PR #464 — `IntoR for Vec<Option<C>>`

**Verdict: REQUEST_CHANGES** — one real gap (incomplete cardinality matrix in new fixture sections, inconsistent with the file's own doc contract). Everything else is correct.

---

### Findings

**Finding 1 — Incomplete cardinality matrix in new fixture sections (correctness gap)**

`rpkg/src/rust/dataframe_enum_payload_matrix.rs` (lines 26–154) and the test file (`rpkg/tests/testthat/test-dataframe-enum-payload-matrix.R`) both declare in their module/file doc comments that each payload shape exercises four cardinality cells: `1v1r`, `1vNr`, `Nv1r`, `NvNr`.  Every pre-existing shape (vec_width, vec_expand, array, boxed_slice) has all four split variants plus the NvNr align test.

The three new sections (0a `VecOpaqueEvent`, 0b `HashSetEvent`, 0c `BTreeSetEvent`) only expose `_align_nvnr` and `_split_nvnr` functions — no `_split_1v1r`, `_split_1vnr`, `_split_nv1r`. This violates the stated matrix contract in the doc comment and creates a test-coverage gap for the 1-variant bare-data.frame return path and the short/empty-row edge cases.

The plan's flat work list (#2, #3) only specifies NvNr for the new sections, so the implementation is plan-faithful — but the plan itself was underspecified here. The fix needs both the plan and the implementation updated.

**Fix**: Add `_split_1v1r`, `_split_1vnr`, and `_split_nv1r` fixture functions to each of the three new enum sections in `dataframe_enum_payload_matrix.rs`, matching the pattern at lines 178–207 (vec_width). Add corresponding test blocks in the R test file. Update the plan's flat work list items #2 and #3 to match.

If the intent is to intentionally scope the new sections to NvNr only (e.g., to keep the PR small), that's fine — but the module doc comment must be updated to say "NvNr only for sections 0a/0b/0c" and a follow-up issue should be filed.

---

### Not findings (confirmed correct)

**GC discipline — LGTM.**  `vec_option_of_into_r_to_list` protects the outer VECSXP via `OwnedProtect` (calls `Rf_protect` immediately after `Rf_allocVector`). Inner `elt` values from `v.into_sexp()` are passed directly to `SET_VECTOR_ELT` with no intervening allocation between the return and the store. `SET_VECTOR_ELT` does a write barrier (`CHECK_OLD_TO_NEW`) but does not allocate and cannot trigger GC. Pattern is identical to `vec_of_into_r_to_list`. Safe under R-devel's aggressive GC.

**GC stress fixture — present and correct.** `gc_stress_vec_option_collection` in `rpkg/src/rust/gc_stress_fixtures.rs` exercises `Vec<Option<Vec<i32>>>`, `Vec<Option<HashSet<String>>>`, and `Vec<Option<BTreeSet<i32>>>` as no-arg exported functions. Per CLAUDE.md convention.

**Coherence — no overlap.** The new impls are `Vec<Option<Vec<T>>>`, `Vec<Option<HashSet<T>>>`, etc. — none overlap with the existing scalar impls (`Vec<Option<i32>>`, `Vec<Option<f64>>`, `Vec<Option<String>>`, etc.) which are all `Vec<Option<ScalarType>>` not `Vec<Option<CollectionType>>`. Rust resolves them unambiguously.

**All 8 impls present.** The plan specified 8 concrete impls; all 8 are present in the `// region: Vec<Option<Collection>> conversions` block.

**R-side semantics correct.**
- HashSet tests use `expect_setequal` (unordered) — correct.
- BTreeSet tests use `expect_equal` with sorted values — correct (input `[3,1,2]` → `[1,2,3]`).
- Absent-variant rows produce NULL, not NA or empty vector — correct.

**Generated files in sync.** `rpkg/R/miniextendr-wrappers.R`, `NAMESPACE`, and `man/dataframe_enum_payload_matrix.Rd` all include the new fixture exports. Pre-commit hook would have caught mismatches.

**Code style — LGTM.** `// region:` / `// endregion` markers used, no `mod.rs` pattern, `Infallible` error type, `From`/`TryFrom` not `as` casts.

---

### Style nits (non-blocking)

- `rpkg/src/rust/dataframe_enum_payload_matrix.rs` line 1: the module doc comment still says "For each currently-supported payload shape, exercise four cardinality cells" — if Finding 1 is resolved by filing an issue and deferring 1v1r/1vNr/Nv1r for sections 0a/0b/0c, update this sentence.
- The three new enum sections use `#[allow(dead_code)]` at file level. The file-level allow is already present so this is a no-op, but the struct fields `label`, `items`, `tags`, `cats` etc. are used by `DataFrameRow` derive, not dead — no change needed.

CGMossa:
Applied review finding: added missing 1v1r/1vNr/Nv1r cardinality cells for vec_opaque / hashset / btreeset to match the existing matrix coverage. Local CI green (check, clippy_default, clippy_all, devtools-test: 5292 passed / 0 failed).

Issue #470, opened 2026-05-10
WebR/WASM support — port miniextendr to wasm32-unknown-emscripten


CGMossa:
Umbrella tracking issue for porting miniextendr to webR (R compiled to WebAssembly via Emscripten).

**Status (2026-05-29): core complete and CI-validated.** Steps 1–8 below all shipped; tier-1/2/3 CI is green. The remaining genuinely-open item under this umbrella is cross-crate trait dispatch (#495); see the follow-up list at the bottom.

Design now lives in **[`docs/WEBR.md`](../blob/main/docs/WEBR.md)** (toolchain rationale, the wasm_registry snapshot mechanism, the emcc side-module link, the two-bundle Node gotcha, and the 3-tier CI). The original `plans/webr-support.md` / `plans/wasm-registry-codegen.md` / `plans/webr-dockerfile.md` design notes were folded into `docs/WEBR.md` and into the issues linked below when `plans/` was retired wholesale (#512) — those paths no longer exist.

## Why

`linkme` (`compile_error!`s on `wasm32-*`) is load-bearing for runtime registration of three slices: `MX_CALL_DEFS`, `MX_ALTREP_REGISTRATIONS`, `MX_TRAIT_DISPATCH`. We need a different gather mechanism on WASM.

## Steps

- [x] **Step 1 — Stabilise symbol names so a separate compilation unit can reference them by name.**
    - [x] #392 ALTREP register fn `pub extern "C"` + `#[unsafe(no_mangle)]`
    - [x] #393 trait vtable `pub` + `#[unsafe(no_mangle)]`
    - [x] #394 generic-type vtable name collision (append stable hash of generics)
- [x] **Step 2 — `cfg_attr`-gate every `#[distributed_slice(...)]` emission in `miniextendr-macros`.** (#471)
- [x] **Step 3 — Add `MX_*_SYMBOLS` parallel `&'static str` slices** populated by macros so the cdylib write step has the symbol names without reverse-symbol lookup. (#472)
- [x] **Step 4 — Cdylib snapshot writer.** `wasm_registry_writer` formats `<crate>/src/rust/wasm_registry.rs` from the runtime slices + symbol slices, with a generator-version + content-hash header. (#473)
- [x] **Step 5 — `registry.rs` cfg branching.** Drop `linkme::distributed_slice` under `cfg(target_arch = "wasm32")`; read from a `OnceLock` populated at `R_init`; `build.rs` validates the snapshot's generator-version. (#474, #475)
[...]

PR #480, opened 2026-05-10
ci: webR wasm32 cargo-check workflow (#470)

CGMossa:
## Summary

- Adds `.github/workflows/webr.yml`: a paths-gated CI job that runs `cargo check --target wasm32-unknown-emscripten` against `miniextendr-api` and `rpkg/src/rust`
- Catches cfg-gating regressions and macro-emission bugs on the wasm32 target without requiring a full webR container or emscripten link step
- Refs #470 (tier 1 of the webR CI follow-ups)

## Tier scope

This PR implements **tier 1 only**: `cargo check` on `wasm32-unknown-emscripten`.

- **Tier 1 (this PR)**: `cargo check` — catch cfg-gate and macro regressions; no linking, no runtime
- **Tier 2 (deferred)**: full wasm32 link via emscripten/webR Docker image; verifies no missing symbols
- **Tier 3 (deferred)**: runtime webR smoke tests

Tiers 2 and 3 are deferred per `plans/webr-ci.md` — they require a webR container, substantially more CI time, and are blocked on resolving the emscripten toolchain integration.

## Paths filter rationale

The workflow triggers on changes to:
- `miniextendr-api/**` — primary wasm32 surface; cfg gates live here
- `miniextendr-macros/**` — emits the cfg-gated code
- `miniextendr-engine/**` — codegen engine; affects what macros emit
- `miniextendr-lint/**` — build-time lint runs as build-dep on wasm32 builds
- `rpkg/**` — rpkg src + `wasm_registry.rs`
- `tests/cross-package/**` — cross-package crates use miniextendr-api
- `Cargo.toml` / `Cargo.lock` — workspace member / dep changes
- `Dockerfile.webr` — webR base digest changes (signals tier 2/3 environment shift)
- `.github/workflows/webr.yml` — re-trigger on workflow edits

Excluded: `miniextendr-bench/**` (native perf only), `miniextendr-cli/**` (tooling, not built into wasm rpkg), docs/site/markdown.

## Open question: preflight status

**Preflight skipped locally — relying on the PR's first CI run as verification.**

The local environment does not have `wasm32-unknown-emscripten` installed, so `cargo check --target wasm32-unknown-emscripten` was not run locally before opening this PR.

Additionally, `bash ./rpkg/configure` calls `R RHOME` early. If the CI runner lacks R, configure may error before writing `.cargo/config.toml`, causing the rpkg cargo check to resolve `miniextendr-api` from git instead of the local workspace siblings.

**Fallback plan if first CI run fails due to missing R:**
1. Add a `r-lib/actions/setup-r@v2` step before "Configure rpkg" (~45 s overhead), **or**
2. Drop the rpkg cargo check from tier 1 and run only `cargo check -p miniextendr-api` until a configure-without-R solution is in place

The reviewer should check the first CI run and advise which fallback to apply.

## Acceptance checklist

- [x] Single new file only (`.github/workflows/webr.yml`)
- [x] Action versions match other workflows (`actions/checkout@v6`, `dtolnay/rust-toolchain@stable`, `Swatinem/rust-cache@v2`)
- [x] Env vars match `ci.yml` global env (`CARGO_TERM_COLOR`, `RUST_BACKTRACE`, `CARGO_INCREMENTAL`)
- [x] Concurrency block matches `ci.yml` pattern
- [x] Paths filter has all 10 entries in both `pull_request.paths` and `push.paths`
- [x] `dtolnay/rust-toolchain@stable` has `targets: wasm32-unknown-emscripten`
- [x] "Configure rpkg" step precedes "cargo check (rpkg)"
- [ ] At least one green `wasm32 check / cargo check` status visible (pending first CI run)

### DIFF
--- a/.github/workflows/webr.yml
+++ b/.github/workflows/webr.yml
@@ -48,6 +48,12 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      # build.rs:64 unconditionally runs `R RHOME`; $RUNNER_TEMP is always a
+      # real directory so Path::new(&r_home).is_dir() passes. cargo check never
+      # links, so the dummy value is sufficient. Same pattern as rust-lint in ci.yml.
+      - name: Set dummy R_HOME
+        run: echo "R_HOME=$RUNNER_TEMP" >> "$GITHUB_ENV"
+
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
@@ -62,9 +68,8 @@ jobs:
       - name: cargo check (miniextendr-api)
         run: cargo check -p miniextendr-api --target wasm32-unknown-emscripten
 
-      - name: Configure rpkg
-        run: bash ./rpkg/configure
-
-      - name: cargo check (rpkg)
-        working-directory: rpkg/src/rust
-        run: cargo check --target wasm32-unknown-emscripten
+      # rpkg/configure invokes ${R_HOME}/bin/Rscript directly for feature
+      # detection (configure.ac:133, :275, :473). The dummy R_HOME has no
+      # Rscript binary, so configure fails. Deferred to a follow-up that either
+      # adds r-lib/actions/setup-r@v2 or waits for the build.rs wasm32 guard
+      # (see #482). Only miniextendr-api is checked in tier 1.
--- a/.github/workflows/webr.yml
+++ b/.github/workflows/webr.yml
@@ -1,0 +1,70 @@
+name: webR / wasm32
+
+on:
+  pull_request:
+    paths:
+      - 'miniextendr-api/**'
+      - 'miniextendr-macros/**'
+      - 'miniextendr-engine/**'
+      - 'miniextendr-lint/**'
+      - 'rpkg/**'
+      - 'tests/cross-package/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'Dockerfile.webr'
+      - '.github/workflows/webr.yml'
+  push:
+    branches: [main]
+    paths:
+      - 'miniextendr-api/**'
+      - 'miniextendr-macros/**'
+      - 'miniextendr-engine/**'
+      - 'miniextendr-lint/**'
+      - 'rpkg/**'
+      - 'tests/cross-package/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'Dockerfile.webr'
+      - '.github/workflows/webr.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  CARGO_INCREMENTAL: "0"
+
+jobs:
+  cargo-check:
+    name: wasm32 check
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-emscripten
+
+      - name: Cache Cargo
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: wasm32-check
+          cache-on-failure: true
+
+      - name: cargo check (miniextendr-api)
+        run: cargo check -p miniextendr-api --target wasm32-unknown-emscripten
+
+      - name: Configure rpkg
+        run: bash ./rpkg/configure
+
+      - name: cargo check (rpkg)
+        working-directory: rpkg/src/rust
+        run: cargo check --target wasm32-unknown-emscripten


CGMossa (commented):
## Review: ci: webR wasm32 cargo-check workflow

### Acceptance checklist

- **Only `.github/workflows/webr.yml` added, nothing else modified** — PASS. Single file, 70 additions, 0 deletions. `git diff --name-only` shows exactly one file.
- **Action versions match other workflows** — PASS. `actions/checkout@v6`, `dtolnay/rust-toolchain@stable`, `Swatinem/rust-cache@v2` all match `ci.yml`. No new versions introduced.
- **Env vars match `ci.yml` global env** — PASS. `CARGO_TERM_COLOR: always`, `RUST_BACKTRACE: 1`, `CARGO_INCREMENTAL: "0"` are all present and match exactly.
- **Concurrency block correct** — PASS. `group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}` with `cancel-in-progress: true` — identical to `ci.yml`.
- **Paths filter has all 10 entries in both `pull_request.paths` and `push.paths`** — PASS. Both blocks are identical and contain all 10 entries from the brief (including `miniextendr-engine/**` and `miniextendr-lint/**`).
- **Toolchain has `targets: wasm32-unknown-emscripten`** — PASS.
- **`Configure rpkg` step precedes `cargo check (rpkg)`** — PASS. Step order is correct.
- **At least one green `wasm32 check / cargo check` CI run visible** — FAIL. First CI run is red.

---

### Blocker (must fix before merge)

**Issue 1: `miniextendr-api/build.rs` calls `R RHOME` unconditionally, breaking `cargo check` without R installed**

The workflow fails at the very first step — `cargo check -p miniextendr-api` — because `build.rs:64` runs `R RHOME` regardless of the target architecture. Error:

```
thread 'main' panicked at miniextendr-api/build.rs:64:14:
Failed to run `R RHOME` (set R_HOME or put `R` on PATH): Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

The `configure` step was never even reached — this is an `R`-not-on-`PATH` failure in `build.rs`, not in the shell configure script.

**The fix already exists in this repo.** The `rust-lint` job in `ci.yml` (line 184–185) uses exactly this workaround:

```yaml
- name: Set dummy R_HOME
  run: echo "R_HOME=$RUNNER_TEMP" >> "$GITHUB_ENV"
```

`$RUNNER_TEMP` is always a real directory on GitHub Actions runners, which satisfies `build.rs`'s `Path::new(&r_home).is_dir()` check. Since `cargo check` never links, it never needs the actual libR — the dummy directory is sufficient for type-checking.

**Required change to `.github/workflows/webr.yml`:** add the `Set dummy R_HOME` step immediately after `actions/checkout@v6`, before the toolchain install (matching the pattern in `rust-lint`):

```yaml
      - uses: actions/checkout@v6

      - name: Set dummy R_HOME
        run: echo "R_HOME=$RUNNER_TEMP" >> "$GITHUB_ENV"

      - name: Install Rust toolchain
        ...
```

**Why not fix `build.rs` instead?** A `CARGO_CFG_TARGET_ARCH == "wasm32"` guard in `build.rs` would be the more correct long-term fix — anyone running `cargo check --target wasm32-unknown-emscripten` locally without R installed will hit the same error. But that is a separate change to `miniextendr-api/build.rs` and is out of scope for this single-file workflow PR. Add a `gh issue create` for the build.rs improvement and reference it in the PR body. The dummy-R_HOME workaround is the established repo pattern and is sufficient to unblock this PR.

---

### Issues to fix (numbered)

1. **[`.github/workflows/webr.yml`]** Add `Set dummy R_HOME` step (same as `ci.yml` lines 184–185) after `actions/checkout@v6` and before `Install Rust toolchain`. See above for exact YAML.

2. **[PR body]** Open a follow-up issue for `build.rs` wasm32 guard and reference it in the PR body. Suggested: `miniextendr-api/build.rs: skip link_to_r() when CARGO_CFG_TARGET_ARCH == "wasm32"`. This is the real fix that makes `cargo check --target wasm32-unknown-emscripten` work without a workaround for any contributor running it locally.

---

### Nits / optional

- The `Configure rpkg` step (`bash ./rpkg/configure`) will now also need R to be absent — BUT since `R_HOME` is set to `$RUNNER_TEMP` (a real directory), `configure` should now proceed past the R_HOME check. However, `configure` does more than just set R_HOME — it tries to run `Rscript` for feature detection. This _may_ still fail. If the rpkg cargo check step fails after the miniextendr-api step is fixed, the right fallback is to drop the rpkg `cargo check` from tier 1 (run only `cargo check -p miniextendr-api`) until step 8 (configure.ac + build.rs integration) lands. Document the decision if this happens.

- No `permissions:` inheritance issue — `contents: read` is present. Good.

- `timeout-minutes: 20` is appropriate for a cold cache run. No change needed.

---

### Verdict: needs fixes

One blocker: add the `Set dummy R_HOME` step (2-line change). Everything else is correct. After that single change, push, and verify the CI run goes green before requesting re-review.

CGMossa:
Fixed. Two changes landed in the latest push:

1. **Added `Set dummy R_HOME` step** (verbatim from `ci.yml:184-185`) immediately after `actions/checkout@v6` and before `Install Rust toolchain`. `R_HOME=$RUNNER_TEMP` satisfies `build.rs`'s `Path::new(&r_home).is_dir()` check without an actual R installation.

2. **Dropped `Configure rpkg` and `cargo check (rpkg)` from tier 1.** `rpkg/configure` invokes `${R_HOME}/bin/Rscript` directly at configure.ac:133, :275, and :473 — the dummy `RUNNER_TEMP` value has no Rscript binary, so configure would fail. The step is deferred to a follow-up: #482 tracks adding a `CARGO_CFG_TARGET_ARCH == "wasm32"` guard to `build.rs` (the real fix) and reopening the rpkg check once that or `setup-r` is in place.

Only `cargo check -p miniextendr-api --target wasm32-unknown-emscripten` runs in tier 1 for now.

PR #481, opened 2026-05-10
feat(rpkg): wasm32 detection in configure.ac + build.rs snapshot validation (#470)

CGMossa:
## Summary

- **Part A (`configure.ac` + `Makevars.in`)**: Detects `CC=emcc` (webR's emscripten toolchain) during `R CMD INSTALL`. When detected: sets `IS_WASM_INSTALL=true`, overrides `CARGO_BUILD_TARGET=wasm32-unknown-emscripten`, forces `RUST_TOOLCHAIN=+nightly` if not already pinned, sets `CARGO_BUILD_STD_FLAG="-Z build-std=std,panic_abort"`, and fails fast if `src/rust/wasm_registry.rs` is absent. Makevars.in adds the two `@substitution@` vars, threads `$(CARGO_BUILD_STD_FLAG)` into the `$(CARGO_AR)` cargo build line, and wraps `$(WRAPPERS_R)` in a GNU make `ifeq` so the wrapper-gen step is skipped (touch no-op) on wasm32 installs.

- **Part B (`build.rs`)**: On wasm32 targets, validates that `wasm_registry.rs` exists and carries a `// generator-version: N` header matching `EXPECTED_VERSION = 1` (mirroring `wasm_registry_writer.rs`). Panics with a clear regeneration message on any mismatch. The `rerun-if-changed` directive is placed before the early-return guard so native incremental builds also track the file.

Both template trees (`minirextendr/inst/templates/rpkg/` and `monorepo/rpkg/`) mirror every change. `patches/templates.patch` regenerated via `just templates-approve`.

Closes step 8 of #470.

## Known unknowns

1. **RUSTFLAGS for emscripten SIDE_MODULE link not verified.** The plan proposes `-C relocation-model=pic -C link-args=-s SIDE_MODULE=1`, but exact values depend on webR's pinned emscripten version. This PR does not set them — user-supplied `ENV_RUSTFLAGS` is the right knob. A `# TODO(#470): verify exact RUSTFLAGS for SIDE_MODULE link against webR's pinned emscripten` comment is placed in the Makevars.in wasm no-op branch as a marker.

2. **Nightly availability not validated at configure time.** When `CC=emcc` is detected and `RUST_TOOLCHAIN` is not pinned, configure forces `+nightly`. If nightly is absent, cargo will error at build time with a clear message. This is acceptable (no silent fallback to stable).

3. **`@touch $(WRAPPERS_R)` assumes `R/<pkg>-wrappers.R` already exists.** The configure check only validates `wasm_registry.rs`. If `R/*-wrappers.R` is also missing, touch creates an empty file, resulting in a broken (zero-export) package. The expected workflow is: run a host install first (which generates both files), then cross-compile. Not adding an extra configure check to avoid scope creep; tracked in #470.

4. **`ifeq` inside Make recipe is GNU make-specific.** R CMD INSTALL uses GNU make on Linux/macOS/Windows-Rtools — safe in practice.

5. **Explicit `CARGO_BUILD_TARGET=wasm32-unknown-emscripten` without `CC=emcc` does NOT trigger wasm mode** (`IS_WASM_INSTALL` stays false, `CARGO_BUILD_STD_FLAG` is empty). Documented as non-goal; the emcc CC signal is the canonical webR indicator.

## Verification done

- `just configure && just check` — both exit 0 on native (macOS arm64, rustc 1.95.0)
- `bash ./configure 2>&1 | grep -E "wasm |emcc"` → `checking whether CC is emscripten (webR/wasm32 install)... no` (correct, no false positive on native)
- `just templates-approve && just templates-check` — both exit 0
- Acceptance checklist items manually verified:
  - `grep -n "IS_WASM_INSTALL\|CARGO_BUILD_STD_FLAG" rpkg/src/Makevars.in` → both present (lines 21-22)
  - `grep "ifeq" rpkg/src/Makevars.in` → 1 match at line 110
  - `grep "EXPECTED_VERSION" rpkg/src/rust/build.rs` → `const EXPECTED_VERSION: u32 = 1`
  - Both template `configure.ac` files have the wasm block
  - `rpkg/configure` regenerated and committed (both in one commit with `configure.ac`)
- wasm32-unknown-emscripten target not installed locally; cross-compile verification skipped (noted as optional in brief)


CGMossa (commented):
## Review of feat/webr-step8

### Acceptance checklist

- ✅ `just configure && just check` passes on native — reported by implementer (macOS arm64, rustc 1.95.0); CI is green on all completed jobs.
- ✅ `IS_WASM_INSTALL` and `CARGO_BUILD_STD_FLAG` in Makevars.in header block — confirmed at lines 21–22 in `rpkg/src/Makevars.in`.
- ✅ `CARGO_BUILD_STD_FLAG` threaded into the `$(CARGO_AR)` cargo build line — confirmed at line 91.
- ✅ Exactly one `ifeq ($(IS_WASM_INSTALL),true)` in `rpkg/src/Makevars.in` — confirmed at line 110.
- ✅ `EXPECTED_VERSION: u32 = 1` in `build.rs`, mirrors `wasm_registry_writer.rs`'s `GENERATOR_VERSION = 1` — confirmed.
- ✅ Both template `configure.ac` files (`rpkg/` and `monorepo/rpkg/`) have the wasm detection block — confirmed; `just templates-check` reported passing.
- ✅ `rpkg/configure` regenerated and committed — `IS_WASM_INSTALL` and `CARGO_BUILD_STD_FLAG` appear in `ac_subst_vars` list (lines 609–610 of generated file).
- ✅ `rerun-if-changed=wasm_registry.rs` is before the early `return` in `build.rs` — confirmed; it's the second statement in `main()`.
- ✅ PR body explicitly flags unverified RUSTFLAGS with `# TODO(#470)` comment reference — confirmed in Makevars.in line 112 and in PR body known-unknowns section.

---

### Issues to fix

**1. BLOCKER — `$(WRAPPERS_R)` still depends on `$(CARGO_CDYLIB)` even on the wasm path**

`rpkg/src/Makevars.in` line 109:
```
$(WRAPPERS_R): $(CARGO_CDYLIB)
ifeq ($(IS_WASM_INSTALL),true)
    @touch $(WRAPPERS_R)
```

The `ifeq` only gates the recipe body. The prerequisite `$(CARGO_CDYLIB)` is unconditional — Make always builds it first. On a wasm32 install, this triggers `cargo rustc --target wasm32-unknown-emscripten --crate-type cdylib`, which produces a `.wasm` side-module (not a native `.so`). The cdylib build either fails with linker errors (missing SIDE_MODULE flags) or produces an unusable artifact. The wasm no-op path is never reached cleanly.

Fix: split the rule on `IS_WASM_INSTALL` so the prerequisite itself is conditional:

```makefile
ifeq ($(IS_WASM_INSTALL),true)
$(WRAPPERS_R):
	@echo "wasm32 install: skipping wrapper-gen (pre-generated R/$(PACKAGE_NAME)-wrappers.R + src/rust/wasm_registry.rs expected)"
	@touch $(WRAPPERS_R)
else
$(WRAPPERS_R): $(CARGO_CDYLIB)
	@echo "Generating R wrappers + wasm_registry.rs..."
	@_cdylib='$(CARGO_CDYLIB)'; \
	... existing recipe ...
endif
```

Same fix needed in both template Makevars.in files (`minirextendr/inst/templates/rpkg/Makevars.in` and `minirextendr/inst/templates/monorepo/rpkg/Makevars.in`). Run `just templates-approve` after.

**2. MINOR — Tab-prefixed `# TODO` comment inside a Make recipe**

`rpkg/src/Makevars.in` line 112 (and both template equivalents):
```
	# TODO(#470): verify exact RUSTFLAGS for SIDE_MODULE link against webR's pinned emscripten
```

The leading tab means Make passes this to the shell, which interprets it as a shell comment. It is harmless but adds noise to `make -n` output and is confusing to readers. Move it above the `$(WRAPPERS_R):` rule line as a plain make comment (no tab), or remove it entirely since the PR body already documents the known unknown. (The comment's informational purpose is served by the PR body and the `#470` issue.)

---

### Nits / optional

- **Nit**: The `all` target comment at Makevars.in line 50 still says "Cdylib builds second (`$(WRAPPERS_R)` needs `$(CARGO_CDYLIB)`)". After the fix above, this comment will be stale for wasm installs. Worth updating to mention the conditional.
- **Nit**: On a system where the default toolchain is old (< 1.85) but nightly is installed, the `rustc --version` min-version check fires before the wasm detection block sets `RUST_TOOLCHAIN=+nightly`, so configure fails with "rustc 1.85+ required" even though nightly would satisfy it. This is an edge case unlikely to affect the webR Docker image (which ships a recent nightly), so it's low priority — but worth noting for the future.

---

### Verdict

**Ready to mark non-draft after fixing issue #1** (the `$(CARGO_CDYLIB)` unconditional prerequisite). That's a one-file structural fix in Makevars.in (and its two template mirrors + `just templates-approve`). Issue #2 (tab-prefixed comment) should be fixed in the same commit since it's trivial. Everything else — configure.ac logic, build.rs, generated configure, template ports — is correct.

CGMossa:
Fixed both issues from the review:

**Issue #1 (BLOCKER):** Split `$(WRAPPERS_R)` into two separate target rules gated on `IS_WASM_INSTALL`. The cdylib prerequisite is now only present on the native (non-wasm) branch:

```makefile
ifeq ($(IS_WASM_INSTALL),true)
$(WRAPPERS_R):
    @echo "wasm32 install: skipping wrapper-gen ..."
    @touch $(WRAPPERS_R)
else
$(WRAPPERS_R): $(CARGO_CDYLIB)
    @echo "Generating R wrappers + wasm_registry.rs..."
    ... existing recipe ...
endif
```

Same fix applied to both template files (`minirextendr/inst/templates/rpkg/Makevars.in` and `minirextendr/inst/templates/monorepo/rpkg/Makevars.in`). `just templates-approve` + `just templates-check` both pass.

**Issue #2 (MINOR):** Removed the tab-prefixed `# TODO(#470):` comment from inside the wasm recipe (it was being passed to the shell). Replaced with a bare Make comment block above the `ifeq` that explains the structural reason for the split.

Also updated the `all` target comment to note the conditional wasm32 behavior (nit from the review).

`just configure && just check` passes on native (macOS arm64) — the native path is unchanged.

PR #483, opened 2026-05-10
feat(tests): webr-smoke.sh + just docker-webr-smoke (#470)

CGMossa:
## Summary

Adds a local-only smoke test (`tests/webr-smoke.sh`) and a `just docker-webr-smoke` recipe that exercise the complete wasm32 install path for `rpkg`. The script drives three stages inside the `miniextendr-webr-dev:latest` container: (1) a native `R CMD INSTALL` to regenerate `wasm_registry.rs` from the live cdylib slice contents, (2) a `CC=emcc`-triggered wasm32 `R CMD INSTALL` to produce the side-module, and (3) a webR Node.js session that mounts the wasm library, loads `miniextendr`, optionally installs `testthat`, and runs the full testthat suite. The script exits 0 as long as `library(miniextendr)` succeeds — testthat failures are expected and tolerated (many tests assume worker thread / fork / threading that doesn't exist under wasm).

## Refs

- Refs #470 (webR support tracking issue)

## Stacked on #481

This PR is stacked on `feat/webr-step8` (PR #481) and **cannot merge until #481 lands**. The `CC=emcc` detection in `rpkg/configure.ac` that drives Phase 2's `IS_WASM_INSTALL=true` branch lives entirely in that PR. After #481 merges to `main`, this PR will be retargeted to `main`.

## Verification done

- `bash -n tests/webr-smoke.sh` — syntax check passes locally.
- `git ls-files --stage tests/webr-smoke.sh` confirms mode `100755` (executable bit captured in git).
- `just docker-webr-build` and `just docker-webr-smoke` end-to-end: **not run** — first cold run on amd64-emulated-on-Rosetta (Apple Silicon) takes 1–2 hours for Cargo + wasm32 compilation. Script is self-contained and the reviewer or CI can run it in a native amd64 environment where it will be faster.

## Known limitations

- **Many testthat tests will fail under wasm.** This is expected and by design: tests that exercise the worker thread, `fork()`, OS-level threading, or anything requiring a real shared-library `dlopen` will fail. The script counts and reports pass/fail/skip but does not exit non-zero on test failures — only on infra failures (docker unavailable, build error, `library(miniextendr)` crash).
- **testthat requires HTTPS to `repo.r-wasm.org`.** If the container's outbound network is blocked or the fetch fails, the script skips the testthat suite and checks `library()` only. This is handled gracefully with a clear "skipping testthat suite" message.
- **`wasm_registry.rs` stub check.** After Phase 1, the script greps for the `content-hash` field. If it is still `0000000000000000` (the stub value), the script aborts with a clear diagnostic.
- **amd64 emulation on Apple Silicon.** Docker on macOS arm64 emulates via Rosetta 2. The first cold run (full `cargo build --target wasm32-unknown-emscripten`) is slow — 1–2 hours. Subsequent runs with a warm `rust-target/` are much faster.
- **Makevars restoration.** Phase 2 leaves `rpkg/src/Makevars` in wasm-mode. The `cleanup` trap always runs `bash rpkg/configure` (native, no `CC=`) inside the container to restore the file before exit, even on failure paths.
- **`--no-staged-install` is mandatory** in Phase 2. A staged install would attempt to `dyn.load` the wasm side-module on the host R, which fails. The script always passes this flag.


CGMossa (commented):
## Review of `tests/webr-smoke.sh` + `just docker-webr-smoke`

Brief note on the diff view: `gh pr diff 483` shows the full merge-base diff against `main` (since the branch is stacked on `feat/webr-step8` which itself is ahead of `main`). The actual changes introduced by this PR are only in `tests/webr-smoke.sh` and `justfile` — confirmed by looking at `git log origin/feat/webr-step8..origin/feat/webr-smoke` (single commit). The rest of the `gh pr diff` output belongs to the `feat/webr-step8` layer.

---

### Acceptance checklist

1. **Executable mode 100755** ✅
   `git ls-tree origin/feat/webr-smoke tests/webr-smoke.sh` → `100755`. Git has the bit.

2. **`set -euo pipefail`, `--help`, cleanup trap** ✅
   Line 21: `set -euo pipefail`. `usage()` at line 46 greps the header comment, matches `demanding-smoke.sh` house style. `trap cleanup EXIT` at line 126.

3. **Exit-code policy: 0 on load success, 1 on infra failure** ✅
   `library(miniextendr)` failure sets `process.exitCode = 1` and returns (no `throw`, so `docker_run` propagates the exit code back). testthat failures don't exit non-zero — they're counted and reported, then `main()` returns normally → `process.exitCode` stays 0 → overall exit 0. The `main().catch()` handler at the bottom is an additional safety net for unexpected runtime errors.

4. **Phase 1 uses `/opt/R/current/bin/R` (native); Phase 2 uses `/opt/webr/host/R-4.5.1/bin/R`** ✅
   Lines 172-179 use `${R_NATIVE_EXE}` = `/opt/R/current/bin/R`. Lines 204-215 use `${R_HOST_EXE}` = `/opt/webr/host/R-4.5.1/bin/R`. Both constants are set correctly at the top.

5. **Phase 2 passes `WASM_TOOLS`, `R_SOURCE`, `R_MAKEVARS_USER=webr-vars.mk`, `--no-staged-install`** ✅
   Lines 207-215. All three env vars exported as per `.webr/packages/Makefile` pattern. `--no-staged-install` present. `--no-docs` and `--no-test-load` also present.

6. **NODEFS mount before `library()`** ✅
   In `smoke-runner.mjs`: `webR.FS.mkdir` then `webR.FS.mount('NODEFS', ...)` then `.libPaths(c("/wasm-rlib", ...))` then `library(miniextendr)`. Correct sequence. `webR.FS` (uppercase) matches the actual property name in `webr-main.ts`.

7. **Node imports webR from `file:///opt/webr/src`; baseUrl `file:///opt/webr/dist/`** ✅
   `package.json`: `"webr": "file:///opt/webr/src"`. The webR source `package.json` exports `"./dist/webr.mjs"` for the `"import"` condition, so `import { WebR } from "webr"` resolves correctly. `new WebR({ baseUrl: "file:///opt/webr/dist/" })` — correct path (the `make`-built output, not `src/dist`).

8. **`just docker-webr-smoke` recipe placement, dependency, args passthrough** ✅
   Lines 1189-1190 in justfile, immediately after `docker-webr-test`. Has `docker-webr-build` dep. `*args` passed through.

9. **No `Dockerfile.webr` / `rpkg/` / `.github/workflows/` changes** ✅
   This PR's single commit touches only `tests/webr-smoke.sh` and `justfile`. Confirmed via `git show origin/feat/webr-smoke --stat`.

10. **Stub-detection check after Phase 1** ✅
    Lines 183-193: greps `content-hash:` from `wasm_registry.rs`, extracts with `awk '{print $NF}'` (correctly escaped as `\$NF` in the double-quoted `docker_run` argument), compares against `0000000000000000`, and exits 1 with a clear diagnostic if it matches.

---

### Issues to fix

**Issue 1 (lines 399-403) — Missing timeout on `node smoke-runner.mjs`**

The Node session has no timeout. If the webR WASM runtime hangs (e.g. during wasm instantiation, or R blocks waiting for an event that never fires), `node smoke-runner.mjs` runs forever and the script hangs indefinitely. The brief explicitly asked for `timeout 1800`. Fix:

```bash
    log "Running Node.js smoke runner..."
    docker_run "
        set -euo pipefail
        cd ${SMOKE_TMP}
        timeout 1800 node smoke-runner.mjs
    "
```

`timeout` is available in the container (it's a GNU coreutils command, present on Debian/Ubuntu). This is the same container that already has `just`, `autoconf`, `cargo` etc. — `timeout` is baseline. If the 30-minute cap is too tight for cold testthat runs, bump to 3600, but a cap is essential.

**Issue 2 (lines 237-250, 256-378) — `SMOKE_TMP` hardcoded inside single-quoted heredocs**

The outer variable is `SMOKE_TMP="/tmp/webr-smoke"` but the `write_runner_files` function hardcodes `/tmp/webr-smoke` in both heredoc bodies (package.json path at line 240, JS constant at line 262, R `sink()` path at line 322). Since the heredocs use single-quoted markers (`<<'OUTER_EOF'` / `<<'EOF'`), bash can't expand `$SMOKE_TMP` inside them — the hardcoding is the right technique. But the comment at line 233 says "where we want literal `$SMOKE_TMP` to be already expanded because we pass it as a shell variable" — this is misleading. `$SMOKE_TMP` is never expanded here; the literal `/tmp/webr-smoke` string is used. The comment should say the opposite: "single-quoted heredocs prevent expansion — we hardcode the path directly since it must match `SMOKE_TMP`." Not a bug, but the misleading comment is a trap for future editors who might try to rename `SMOKE_TMP`.

Suggested fix: update the comment, and add a guard at the top of `write_runner_files`:

```bash
write_runner_files() {
    # Hardcoded paths inside single-quoted heredocs must match SMOKE_TMP.
    if [[ "${SMOKE_TMP}" != "/tmp/webr-smoke" ]]; then
        fail "SMOKE_TMP changed but write_runner_files hardcodes /tmp/webr-smoke — update both."
        exit 1
    fi
```

This self-documents the coupling and catches future drift.

---

### Nits / optional

- **`bash -n` as verification.** The implementer correctly notes it only catches syntax errors. Semantic bugs (wrong env var names, path typos that only manifest inside the container, JS runtime errors) won't surface until the actual docker run. That's expected — noted in the PR body. The review is accordingly weighted toward code-reading correctness, not runtime proof.

- **`npm install --prefer-offline`** — This is a perf hint for caches (uses local cache before network). Since the dependency is `file:///opt/webr/src` (a local symlink), npm never hits the network anyway. `--prefer-offline` is harmless but also a no-op here; could be `--ignore-scripts` to avoid postinstall surprises, but webR's `package.json` has no `postinstall`. Leave as-is.

- **The auto-build path in preflight** (lines 149-154) builds the image when it's not found, even when `--rebuild-image` wasn't passed. This is user-friendly but could surprise a maintainer who wants to intentionally use a stale image for debugging. Low-stakes; leave as-is.

- **`docker info &>/dev/null`** at line 138 does a full daemon status check, which is slightly slow on macOS. Fine for a smoke script.

---

### Verdict

**Request changes on Issue 1 (timeout); Issue 2 is a fix-or-comment-update at author's discretion.**

Issue 1 is the only real functional gap: without a timeout, a hanging webR session blocks the script indefinitely. Everything else in the acceptance checklist passes. The heredoc quoting is correct — the `<<'OUTER_EOF'`/`<<'EOF'` nesting works as intended, JS `${...}` template literals pass through verbatim, and the package.json `"webr": "file:///opt/webr/src"` is syntactically correct. The toJs() unwrapping for both the load check and the testthat result map matches the actual webR API shape.

Once the timeout is added (and ideally the misleading comment in `write_runner_files` is corrected), this is ready to merge after #481 lands.

CGMossa:
Both issues addressed in the latest commit:

**Issue 1 (timeout):** `node smoke-runner.mjs` is now wrapped with `timeout 1800` — a 30-minute cap that covers the slowest Rosetta/cold-WASM scenario while ensuring CI/agent contexts can't hang indefinitely.

**Issue 2 (SMOKE_TMP comment + guard):** The misleading comment in `write_runner_files` has been corrected — it now accurately says single-quoted heredocs do no expansion and that `/tmp/webr-smoke` is hardcoded. Added a guard at the top of `write_runner_files` that asserts `SMOKE_TMP == "/tmp/webr-smoke"` and fails loudly if they diverge, so future refactors that rename the variable don't silently write runner files to the wrong container path.

`bash -n tests/webr-smoke.sh` passes cleanly.

CGMossa:
Closing due to head/base ref corruption from a botched base retarget after #481 merged. Reopening from feat/webr-smoke against main.

PR #535, opened 2026-05-14
fix(ci): wire test-bootstrap-vendor regression into CI (#452)

### LINKED ISSUE #452: ci: wire test-bootstrap-vendor into CI workflow (#441 follow-up)
## Context

PR #441 (fix/441-vendor-latch-hygiene) adds `just test-bootstrap-vendor` as a regression test that verifies `bootstrap.R` produces `inst/vendor.tar.xz` from a clean source tree. This catches the #439/#440 regression pattern.

## What's needed

Wire `just test-bootstrap-vendor` into `.github/workflows/ci.yml` as a dedicated CI job (or as a step in an existing R-build job). The job needs:
- `cargo-revendor` on PATH
- `pkgbuild` installed in R

This was explicitly deferred from the PR fixing #441 to keep the change scope tight. The test script gracefully skips if either prerequisite is missing, so adding it to CI without these prerequisites would silently no-op.

## Related

- Follow-up from the PR closing #441
- See `tests/bootstrap-produces-vendor.sh` and `just test-bootstrap-vendor` recipe

CGMossa:
## Summary

- Adds a dedicated `bootstrap-vendor-test` CI job that runs `just test-bootstrap-vendor` (i.e., `tests/bootstrap-produces-vendor.sh`) on every push/PR when R or Rust files change.
- Ensures `bootstrap.R`'s auto-vendor path produces `inst/vendor.tar.xz` from a clean tree, catching the #439/#440 regression pattern automatically.
- Registers the new job in `ci-success` so branch protection gates on it.

## Setup

The job mirrors `sync-checks`: R release + Rust stable + cargo-revendor.
`pkgbuild` is installed directly (no full `setup-r-dependencies` needed — only `pkgbuild::build` is called, not `R CMD check`).
No `autoconf`, no `just configure` — the test must exercise the cold bootstrap path.

## Test plan

- [ ] "Bootstrap Vendor Test" appears as a named check on this PR
- [ ] The job passes (green)
- [ ] `ci-success` still passes

Closes #452

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1060,6 +1060,57 @@ jobs:
           retention-days: 7
 
   # ===========================================================================
+  # Bootstrap vendor regression test (#441 latch)
+  # ===========================================================================
+  bootstrap-vendor-test:
+    name: Bootstrap Vendor Test
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    needs: changes
+    if: needs.changes.outputs.r == 'true'
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      SCCACHE_GHA_ENABLED: "true"
+      RUSTC_WRAPPER: "sccache"
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          lfs: true
+
+      - name: Install just
+        uses: taiki-e/install-action@just
+
+      - name: Setup R
+        uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: release
+          use-public-rspm: true
+
+      - name: Install pkgbuild
+        run: Rscript -e 'install.packages("pkgbuild", repos = "https://cloud.r-project.org")'
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - uses: mozilla-actions/sccache-action@v0.0.9
+
+      - name: Cache Cargo
+        uses: Swatinem/rust-cache@v2
+        with:
+          # bootstrap.R calls cargo-revendor which calls `cargo metadata` on the
+          # rpkg workspace — cache that workspace's target dir.
+          workspaces: rpkg/src/rust -> target
+          shared-key: bootstrap-vendor-test
+          cache-on-failure: true
+
+      - name: Install cargo-revendor
+        run: cargo install --path cargo-revendor
+
+      - name: Run bootstrap-vendor regression test
+        run: just test-bootstrap-vendor
+
+  # ===========================================================================
   # Summary job (for branch protection)
   # ===========================================================================
   ci-success:
@@ -1077,6 +1128,7 @@ jobs:
       - cross-package-tests
       - minirextendr
       - cran-check
+      - bootstrap-vendor-test
     runs-on: ubuntu-latest
     if: always()
     steps:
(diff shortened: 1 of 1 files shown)


CGMossa (commented):
## Review: bootstrap-vendor-test CI job

Overall: **APPROVE-after-fixes** — one minor issue worth fixing before merge, everything else is solid.

---

### What's correct

- Both `ci-success.needs` and the `results` array are updated (lines 1128 and 1147 in patched file). No yaml-array gotcha here.
- `actions/checkout@v6` matches every other job in the file. ✓
- `mozilla-actions/sccache-action@v0.0.9` matches all seven existing uses. ✓
- `pkgbuild` is explicitly installed — the script would SKIP (exit 0) without it, so the job would appear green while not actually testing anything. Installing it is required and correct.
- `cargo-revendor` installed via `cargo install --path cargo-revendor` — matches `sync-checks`. ✓
- `timeout-minutes: 15` matches plan. `sync-checks` has no timeout at all so this is actually stricter, which is fine.
- `if: needs.changes.outputs.r == 'true'` — correct gate; same as `r-check-linux`, `r-tests`, `cran-check`, etc. The bootstrap path is exercised when rpkg/R files or Rust sources touching rpkg change; those paths are all under the `r` filter.
- No `just configure`, no `autoconf` — cold bootstrap path preserved. ✓
- `Rust stable` (not pinned version) consistent with the rest. ✓
- Placement: new job is inserted between `cran-check` and `ci-success`. Plan said "after `cran-check` and before `ci-success`" — matches. ✓

---

### Issue 1 (minor, fixable): `sync-checks` has no `timeout-minutes`; this job does but is inconsistent with it

Not a blocker for this PR but worth noting: `sync-checks` (the job this one was modelled after) has no `timeout-minutes`. The new job has `timeout-minutes: 15`, which is the right call per the plan. No action needed — this is strictly better than the baseline.

---

### Issue 2 (minor): sccache on a `cargo install` step — low yield

`.github/workflows/ci.yml:line ~1102` in the new job: sccache is configured but `cargo install --path cargo-revendor` is not a workspace build — it compiles `cargo-revendor` from its own standalone manifest. The `Swatinem/rust-cache@v2` `workspaces: rpkg/src/rust -> target` key targets the rpkg Rust workspace, not the `cargo-revendor` binary compilation. The `cargo-revendor` binary itself isn't cached by either the `rust-cache` or `sccache` layers in a meaningful way here. This is the same pattern as `sync-checks` (which also does `cargo install --path cargo-revendor` without a separate cache for it), so it's not a regression — just an accepted cost (plan item 7).

Not a blocker.

---

### Issue 3 (blocker candidate): `cargo-revendor` `install` does not cache across runs efficiently

The `Swatinem/rust-cache@v2` `workspaces` key points at `rpkg/src/rust -> target`. That is `cargo-revendor`'s dependency tree will not be in that cache. Every cold runner will re-compile `cargo-revendor` from scratch (the sccache wrapper helps here, but only if GHA sccache has a warm entry). This is the same tradeoff as `sync-checks` and is explicitly accepted in plan section 7 — so not a blocker, just noting it matches the plan's accepted risk.

---

### Issue 4 (real concern): Missing `timeout-minutes` on `sync-checks` but not a bug in this PR

Not introduced here.

---

### Verdict

The implementation is a faithful, pixel-perfect copy of the plan. The only non-cosmetic gap I can identify is that **`cargo-revendor` re-compilation is not cached**, but this is explicitly called out and accepted in the plan (section 7, "First run will be slow (~2 min to compile cargo-revendor); subsequent runs ~30 s"). The `sccache` wrapper does cover the object-file level, so subsequent runs should be fast.

**No blocking issues.** LGTM with the note that if the first CI run shows the job taking >10 min, the fallback is to add a `rust-cache` entry keyed to the `cargo-revendor` standalone workspace.

PR #536, opened 2026-05-14
fix(minirextendr): cran-prep trap-cleans inst/vendor.tar.xz on failure (#454)

### LINKED ISSUE #454: minirextendr template: cran-prep should trap-clean inst/vendor.tar.xz on failure
## Leak path

`minirextendr/inst/templates/rpkg/justfile`'s `cran-prep` recipe runs:

```just
cran-prep: vendor rbuild rcheck
    rm -f inst/vendor.tar.xz
```

The `rm -f inst/vendor.tar.xz` is in the recipe body, not wrapped in a `trap … EXIT`. If `rbuild` or `rcheck` fails mid-run, `just` exits the recipe early and the `rm -f` line never executes. The tarball remains in the source tree, flipping every subsequent `configure` into tarball mode — silently freezing out cargo-resolved dependency edits.

## Expected fix

Port the trap-cleanup pattern used in the main justfile's `r-cmd-build` / `r-cmd-check` recipes:

```just
[script("bash")]
cran-prep: vendor
    set -euo pipefail
    trap 'rm -f inst/vendor.tar.xz' EXIT
    just rbuild
    just rcheck
```

(Or keep the recipe dependencies on `rbuild` / `rcheck` and use a wrapper `[script("bash")]` + trap.)

## Notes

- Pre-existing in the template — not introduced by PR #453.
- The main justfile's equivalent recipes already use `trap` correctly.
- Tracked here so the template stays in sync with the fix.

CGMossa:
## Summary

- Fixes the latch-leak bug in `cran-prep` in both template justfiles: if
  `rbuild` or `rcheck` fails mid-run, the `rm -f inst/vendor.tar.xz` line
  was never reached, leaving the tarball in the source tree and silently
  switching subsequent `configure` runs into tarball mode.
- Ports the `[script("bash")]` + `trap 'rm -f …' EXIT` pattern already used
  by the root justfile's `r-cmd-build` and `r-cmd-check` recipes.
- Moves `rbuild` and `rcheck` from recipe deps to the body, so the trap
  covers them (deps run before the body; a trap registered in the body cannot
  catch dep failures).
- Regenerates `patches/templates.patch` to keep CI's sync-check passing.

**Files changed:**
- `minirextendr/inst/templates/rpkg/justfile` — standalone template
- `minirextendr/inst/templates/monorepo/justfile` — monorepo template
  (same root cause; included per the "no branch scoping" project rule)
- `patches/templates.patch` — regenerated via `just templates-approve`

## Test plan

- [ ] `just templates-check` exits 0
- [ ] Failure simulation: inject `false` before `just rcheck` → verify
      `inst/vendor.tar.xz` is absent after `cran-prep` fails
- [ ] CI sync-check passes (templates.patch up to date)

Closes #454

### DIFF
--- a/minirextendr/inst/templates/monorepo/justfile
+++ b/minirextendr/inst/templates/monorepo/justfile
@@ -142,8 +142,12 @@ vendor-lib crate dev_path:
 # `just install` / `devtools::install("{{rpkg}}")` silently switch to tarball
 # mode (configure's only signal is `[ -f inst/vendor.tar.xz ]`), which freezes
 # out monorepo workspace-crate edits via `[patch."git+url"]`.
-cran-prep: vendor rbuild rcheck
-    rm -f {{rpkg}}/inst/vendor.tar.xz
+[script("bash")]
+cran-prep: vendor
+    set -euo pipefail
+    trap 'rm -f {{rpkg}}/inst/vendor.tar.xz' EXIT
+    just rbuild
+    just rcheck
 
 # ============================================================================
 # Version management
--- a/minirextendr/inst/templates/rpkg/justfile
+++ b/minirextendr/inst/templates/rpkg/justfile
@@ -143,14 +143,17 @@ vendor:
 
 # Prepare for CRAN submission: vendor deps, build, check
 #
-# Cleanup: rpkg/inst/vendor.tar.xz must be removed after the build/check
+# Cleanup: inst/vendor.tar.xz must be removed after the build/check
 # pipeline. Otherwise the leftover in the source tree makes the next
 # `just install` / `devtools::install(".")` silently switch to tarball mode
 # (configure's only signal is `[ -f inst/vendor.tar.xz ]`), which freezes out
 # any cargo-resolved dependency edits.
-cran-prep: vendor rbuild rcheck
-    # TODO(#454): trap-clean on failure — see issue #454
-    rm -f inst/vendor.tar.xz
+[script("bash")]
+cran-prep: vendor
+    set -euo pipefail
+    trap 'rm -f inst/vendor.tar.xz' EXIT
+    just rbuild
+    just rcheck
 
 # ============================================================================
 # Version management
--- a/patches/templates.patch
+++ b/patches/templates.patch
@@ -1,6 +1,6 @@
 diff -ruN -U2 a/monorepo/rpkg/bootstrap.R b/monorepo/rpkg/bootstrap.R
---- a/monorepo/rpkg/bootstrap.R	2026-05-14 09:18:35
-+++ b/monorepo/rpkg/bootstrap.R	2026-04-28 10:42:41
+--- a/monorepo/rpkg/bootstrap.R	2026-05-14 10:07:47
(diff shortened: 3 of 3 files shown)


CGMossa (commented):
## Review

**Verdict: APPROVE-after-fixes** — one real issue, two minor items.

---

### Issue 1 (minor, style): Stale comment prefix corrected, but verify

`minirextendr/inst/templates/rpkg/justfile` line 146 (pre-PR) said:
```
# Cleanup: rpkg/inst/vendor.tar.xz must be removed after the build/check
```
The PR changes this to `inst/vendor.tar.xz` — that's correct. The standalone rpkg template's justfile lives inside the rpkg root, so there is no `rpkg/` prefix. Good catch.

---

### Issue 2 (correctness): `patches/templates.patch` — `---` line was missing before; now fixed

The pre-PR patch had a malformed diff header (only `+++`, no `---` line at offset 2). The PR's `just templates-approve` regeneration fixes this as a side effect. Fine as-is; nothing to change, just noting the patch fix is real, not spurious.

---

### Issue 3 (potential CI concern): `patches/templates.patch` does NOT record the justfile changes

`just templates-sources` only tracks files derived from `rpkg/` (bootstrap.R, configure.ac, build.rs, Makevars.in, etc.) — **the justfiles are not in that list** because they are templates-only, not derived from `rpkg/`. So the patch correctly does not capture justfile deltas. However, the PR test-plan says "CI sync-check passes" — if CI runs `just templates-check` and the justfile is not tracked, CI will not catch a future justfile drift. This is a pre-existing architectural gap, not introduced by this PR. Worth noting with a `gh issue create` per the concessions-as-issues policy.

---

### All checklist items pass

- `[script("bash")]` correctly placed above the recipe header line in both files.
- `vendor` retained as the only dep; `rbuild`/`rcheck` moved into the body — correct.
- `trap 'rm -f inst/vendor.tar.xz' EXIT` in rpkg template — correct relative path.
- `trap 'rm -f {{rpkg}}/inst/vendor.tar.xz' EXIT` in monorepo template — correct path (workspace root relative, matching `vendor` recipe).
- `TODO(#454)` comment removed.
- `set -euo pipefail` in body of `[script("bash")]` recipe — matches reference pattern.
- `trap … EXIT` fires on all exits including SIGINT and `set -e` aborts — correct.
- Plan section 3a's self-correction (keeping only `vendor` as dep) is implemented correctly.
- PR body explains why trap-on-success is intentional (tarball is transient).

---

### Action required before merge

File a `gh issue create` for the architectural gap: justfiles in both templates are not tracked by `templates-sources` / `templates-check`, so the sync-check CI job provides no coverage for justfile drift. The issue body should sketch adding `rpkg/justfile` and `monorepo/justfile` to `templates-sources` (with the caveat that monorepo justfile has no corresponding `rpkg/justfile` upstream source, so it may need a separate tracking mechanism).

Other than that issue being filed, this is ready to merge.

CGMossa:
Follow-up filed at #538 (templates-check coverage gap for justfiles) per the review note. Merging.

Issue #523, opened 2026-05-11
vendor: replace miniextendr_vendor() with shell-out to cargo-revendor (single source of truth)


CGMossa:
Follow-up from PR #520 (and §12.3 of `analysis/build-system-investigation-2026-05-11.md`).

## Context

`miniextendr_vendor()` is the R-side path for producing `inst/vendor.tar.xz` from a scaffolded package. PR #520 removed the broken post-processing (clearing `.cargo-checksum.json` to `{\"files\":{}}` and stripping `checksum =` from Cargo.lock) so the R path now agrees with `just vendor` byte-for-byte. But two paths still exist and could drift again.

## Proposal

Refactor `miniextendr_vendor()` to be a thin wrapper around `cargo-revendor`:

```r
miniextendr_vendor <- function(path = \".\") {
  # Same lock-regen dance as just vendor (already in bootstrap.R after #520)
  # Then a single system2(\"cargo\", c(\"revendor\", ...))
  # Followed by the compress step (same options as just vendor)
}
```

Remove all CRAN-trim logic from R (`strip_vendored_dir()` etc.) — cargo-revendor already does it via `--strip`. The R wrapper becomes maybe 30 lines of glue + error handling.

## Trade-off

- Pro: single source of truth for vendor-tree shape.
- Con: scaffolded packages need cargo-revendor on PATH for `miniextendr_vendor()` to work — they currently fall back to a pure-R path if cargo-revendor is absent. Mitigate by clear error message + docs note on installation.

## Acceptance

- `just vendor` and `Rscript -e 'minirextendr::miniextendr_vendor(\"rpkg\")'` produce byte-identical `vendor/` trees (current state since PR #520).
- After refactor: `miniextendr_vendor()` source is < 60 lines.
- Clear error if cargo-revendor is missing.

CGMossa:
**Current state (2026-05-14):** PR #526 (`8b0577b4`) removed the broken post-processing (checksum-zeroing + Cargo.lock checksum stripping), so `miniextendr_vendor()` and `just vendor` now produce matching vendor trees.

[...]

PR #537, opened 2026-05-14
fix(vendor): miniextendr_vendor() regenerates Cargo.lock before vendoring (#523 partial)

CGMossa:
## Summary

- `miniextendr_vendor()` was missing the lock-regeneration dance that
  `just vendor` and `bootstrap.R` both perform before invoking `cargo revendor`.
- A developer calling it from a monorepo checkout could silently produce a
  tarball whose `Cargo.lock` contains `source = "path+file:///..."` entries for
  `miniextendr-{api,macros,lint,engine}`, causing offline CRAN install to fail.
- Ports the `safe_rename()` + `cargo generate-lockfile` block from `bootstrap.R`
  (added in PR #526) into `miniextendr_vendor()`.
- Factors `safe_rename()` into a shared `utils.R` internal helper with a comment
  pointing to the inline bootstrap.R copy.

Addresses #523. The full shell-out refactor (`miniextendr_vendor()` → thin wrapper
around `cargo-revendor`, < 60 lines) remains as the main scope of #523.

## Test plan
- [ ] `miniextendr_vendor("../rpkg")` from monorepo: verify zero `path+file://` in
  `Cargo.lock` after run
- [ ] `just minirextendr-test` passes
- [ ] `just templates-check` clean
- [ ] `unlink("rpkg/inst/vendor.tar.xz")` after test to avoid latch leak

### DIFF
--- a/minirextendr/R/utils.R
+++ b/minirextendr/R/utils.R
@@ -625,3 +625,28 @@ is_miniextendr_package <- function() {
 bullet_created <- function(path, verb = "Created") {
   cli::cli_alert_success("{verb} {.path {path}}")
 }
+
+#' Rename a file, hard-failing on any error
+#'
+#' `file.rename()` returns `FALSE` on failure without throwing — this wrapper
+#' converts any failure into a hard error so callers never silently swallow a
+#' rename that did not happen (important on Windows where renames are
+#' non-atomic and can fail when the target is locked).
+#'
+#' NOTE: `bootstrap.R` carries an identical inline copy of this helper because
+#' it is a standalone script (not a package), not a package function. See #523.
+#'
+#' @param from Source path.
+#' @param to Destination path.
+#' @param context Human-readable context string included in the error message.
+#' @return Invisibly.
+#' @noRd
+safe_rename <- function(from, to, context) {
+  if (!isTRUE(file.rename(from, to))) {
+    cli::cli_abort(c(
+      "Failed to rename {.path {from}} -> {.path {to}}",
+      "x" = context
+    ))
+  }
+  invisible()
+}
--- a/minirextendr/R/workflow.R
+++ b/minirextendr/R/workflow.R
@@ -169,7 +169,7 @@ miniextendr_vendor <- function(path = ".") {
 
   rust_dir      <- usethis::proj_path("src", "rust")
   cargo_cfg     <- fs::path(rust_dir, ".cargo", "config.toml")
-  cargo_cfg_bak <- paste0(cargo_cfg, ".tmp_vendor")
+  cargo_cfg_bak <- paste0(cargo_cfg, ".tmp_bootstrap_vendor")
   cargo_lock    <- fs::path(rust_dir, "Cargo.lock")
   cargo_toml    <- fs::path(rust_dir, "Cargo.toml")
 
--- a/minirextendr/R/workflow.R
+++ b/minirextendr/R/workflow.R
@@ -156,8 +156,67 @@ miniextendr_vendor <- function(path = ".") {
   with_project(path)
   cli::cli_h1("miniextendr vendor workflow")
 
-  # Step 1: cargo revendor + CRAN-trim (delegates to vendor_crates_io)
-  cli::cli_h2("Step 1: vendor all dependencies")
+  # Step 1: regenerate Cargo.lock in tarball-shape.
+  #
+  # In a monorepo dev checkout .cargo/config.toml carries [patch."git+url"]
+  # overrides that redirect miniextendr-{api,lint,macros} to local paths.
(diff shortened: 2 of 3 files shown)


CGMossa (commented):
## Review of PR #537 — `miniextendr_vendor()` lock-regen

**Overall**: nearly correct port; one naming inconsistency needs fixing, one test is brittle and should be replaced. Everything else is solid.

---

### Issue 1 — Backup suffix mismatch (functional defect) 🔴

`workflow.R` uses `.tmp_vendor` as the backup suffix; `bootstrap.R` uses `.tmp_bootstrap_vendor`. While each function uses its own suffix consistently (move-aside and restore both match within the same function), this diverges from the plan and from the stated goal of "consistency for recovery."

More importantly: if a user has a stale `.tmp_bootstrap_vendor` file lying around from a partially-interrupted `bootstrap.R` run, and then calls `miniextendr_vendor()`, the restore step silently leaves `.tmp_vendor` alongside the stale `.tmp_bootstrap_vendor` with no clear connection. A consistent suffix across all three tools (plus `repair-lock.R` uses `.tmp_repair_lock`) makes leftover temps diagnosable.

The plan explicitly required `.tmp_bootstrap_vendor` for consistency. The PR uses `.tmp_vendor`.

**`minirextendr/R/workflow.R`, line with `cargo_cfg_bak <- paste0(cargo_cfg, ".tmp_vendor")`**:
```r
# Should be:
cargo_cfg_bak <- paste0(cargo_cfg, ".tmp_bootstrap_vendor")
```

---

### Issue 2 — Deparse-body test is brittle (per project memory) 🟡

`tests/testthat/test-regressions.R` lines ~280–299 use `deparse(body(miniextendr_vendor))` to assert ordering. Per the project's standing memory rule:

> *deparse(body) tests are brittle — minirextendr regression tests grep function source for literal strings; inlining to satisfy them is implementation theater, not a real fix.*

The plan itself flagged this: "Note: per the project feedback memory, `deparse(body)` tests are flagged as brittle — use this only if the testthat harness already has examples of this pattern for workflow functions. If not, skip the structural test."

The existing harness in `test-repair-lock.R` already has a `run_with_logging` mock pattern. The right test here is behavioural: mock `system2` (record invocations), call `miniextendr_vendor()`, and assert (a) `generate-lockfile` was invoked, (b) it was invoked before `revendor`.

Alternatively, drop the structural test entirely and rely on CI's functional smoke test (`just minirextendr-test`). Either is better than the deparse grep.

**Recommendation**: replace with a mock-based test or drop entirely.

---

### Confirmed correct ✅

- `safe_rename()` semantics match `bootstrap.R`: `!isTRUE(file.rename(...))`, includes `from`/`to`/context in the error.
- Lock-regen block is before `vendor_crates_io()` (Step 1 → Step 2 → Step 3 renumbering is internally consistent; no orphaned "Step 2: compress" comment remains — diff shows all three steps correctly renumbered).
- `cargo_toml` path uses `usethis::proj_path("src", "rust")` → `fs::path(rust_dir, "Cargo.toml")`, matching `vendor_crates_io()`'s own pattern.
- Restore happens before the `lock_status != 0` error check — workspace cannot be left bricked.
- `.cargo/config.toml` move-aside is guarded by `fs::file_exists(cargo_cfg)` — safe when `configure` hasn't been run.
- PR body correctly says "Addresses #523" (not "Closes").
- `safe_rename()` no-ops (i.e., the move-aside block is skipped entirely) when the source is absent.

---

### Summary

Two items to fix before merge: (1) rename `.tmp_vendor` → `.tmp_bootstrap_vendor`, (2) replace or drop the deparse-body test. Issue 1 is a naming correctness defect per the plan; issue 2 is a test quality issue. The core lock-regen logic is correct.

CGMossa:
Addressed both reviewer-flagged items:

**Fix 1 — Backup suffix** (`workflow.R` line 172): renamed `.tmp_vendor` → `.tmp_bootstrap_vendor`. This matches `bootstrap.R`'s suffix so stale temp files from a partially-interrupted run of either function are diagnosable by name.

**Fix 2 — Brittle deparse test** (`test-regressions.R`): dropped the `deparse(body(miniextendr_vendor))` test block entirely. Decision rationale: `mockery` is not in `minirextendr/DESCRIPTION` Suggests/Imports, so adding a mock-based replacement would require a new dependency just for this one test. Per the project memory rule ("deparse(body) tests are brittle — implementation theater"), dropping is cleaner. A comment at the section header records the rationale. Behavioural coverage remains via CI's `just minirextendr-test` functional smoke test.

Diff: `workflow.R` 1-line change (suffix string); `test-regressions.R` net -12 lines (15 removed, 3-line comment added).

PR #539, opened 2026-05-14
feat(configure): accept CARGO_HOME override (#516)

### LINKED ISSUE #516: configure.ac: accept CARGO_HOME override for restricted environments
**Context**: analysis/build-system-investigation-2026-05-11.md §10.2, §12.9.

## Problem

`configure.ac` does not accept or document a `CARGO_HOME` override. End users
on systems with locked-down or read-only `\$HOME` (containers, multi-user
research clusters, air-gapped builds with a pre-populated cargo home elsewhere)
have no standard way to redirect the cargo registry cache from `~/.cargo`.

`CARGO_HOME` is inherited from the parent process environment if set, so a
workaround exists (`CARGO_HOME=/path R CMD INSTALL ...`). But it's
undocumented, and the user has to know to export it before R starts.

## Proposed change

Add an `AC_ARG_VAR([CARGO_HOME], [cargo registry cache location; defaults to user environment])` in
`rpkg/configure.ac` and propagate to the make environment so it reaches
the cargo invocations in Makevars.

Mirror in templates:
- `minirextendr/inst/templates/rpkg/configure.ac`
- `minirextendr/inst/templates/monorepo/rpkg/configure.ac`

Document the override in `docs/` (somewhere under building / install topics)
and in the relevant CLAUDE.md sections.

## Open question

Should configure also try a sensible per-package fallback (e.g.
`tools::R_user_dir('miniextendr', 'cache')`) when `HOME` is unwritable?
Probably not — feature creep.

## Acceptance

- `CARGO_HOME=/tmp/test-cargo R CMD INSTALL rpkg` uses `/tmp/test-cargo` for
  registry cache.
- Default behavior unchanged (no `CARGO_HOME` set → `~/.cargo`).
- Both project structures (monorepo, standalone) work the same way.

CGMossa:
## Summary

- Adds `AC_ARG_VAR([CARGO_HOME], [...])` to `rpkg/configure.ac` (and both
  minirextendr templates) so the variable is documented in `./configure --help`,
  cached in `config.status`, and forwarded to cargo invocations.
- Adds `CARGO_HOME = @CARGO_HOME@` + `export CARGO_HOME` to `Makevars.in`
  (and template copies) so the value reaches cargo at build time.
- Documents `CARGO_HOME` in `docs/ENVIRONMENT_VARIABLES.md`.
- Regenerates `rpkg/configure` via `autoconf`.
- Regenerates `patches/templates.patch` via `just templates-approve`.

Default behavior is unchanged: when `CARGO_HOME` is unset, cargo uses `~/.cargo`
as before.

## Test plan

- [ ] `./configure --help | grep CARGO_HOME` shows the description.
- [ ] Unset `CARGO_HOME`: `grep CARGO_HOME rpkg/src/Makevars` → empty value, build unaffected.
- [ ] Set `CARGO_HOME=/tmp/test-cargo-home just rcmdinstall` → cargo populates `/tmp/test-cargo-home/`.
- [ ] `just templates-check` passes.

Closes #516

### DIFF
--- a/docs/ENVIRONMENT_VARIABLES.md
+++ b/docs/ENVIRONMENT_VARIABLES.md
@@ -23,6 +23,7 @@ single signal:
 | `CARGO_PROFILE` | Build profile: `dev` or `release` | `release` |
 | `CARGO_TARGET_DIR` | Artifact directory (must be outside `src/`) | `${abs_top_srcdir}/rust-target` |
 | `CARGO_BUILD_TARGET` | Rust target triple for cross-compilation | Empty (native); auto-detected from autoconf host |
+| `CARGO_HOME` | Cargo registry/cache directory for offline or restricted environments | Empty (cargo's default, `~/.cargo`) |
 | `RUST_TOOLCHAIN` | Toolchain selector (e.g., `+stable`, `+nightly`) | Empty (system default) |
 | `ENV_RUSTFLAGS` | Rust compiler flags, passed as `RUSTFLAGS` to cargo | Value of `RUSTFLAGS` |
 
--- a/minirextendr/inst/templates/monorepo/rpkg/Makevars.in
+++ b/minirextendr/inst/templates/monorepo/rpkg/Makevars.in
@@ -7,6 +7,7 @@ CARGO_TARGET_DIR      = @CARGO_TARGET_DIR@
 CARGO_PROFILE         = @CARGO_PROFILE@
 CARGO_STATICLIB_NAME  = @CARGO_STATICLIB_NAME@
 ENV_RUSTFLAGS         = @ENV_RUSTFLAGS@
+CARGO_HOME            = @CARGO_HOME@
 CARGO_BUILD_TARGET    = @CARGO_BUILD_TARGET@
 ABS_RPKG_SRC_CARGO    = @ABS_RPKG_SRC_CARGO@
 CARGO_LIBDIR          = @CARGO_LIBDIR@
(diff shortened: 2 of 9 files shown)


CGMossa (commented):
Reviewed against plan `/tmp/plan-516.md` and cross-checked all three configure.ac / Makevars.in files.

## Verdict: APPROVE-after-nit (one cosmetic issue, no blockers)

### The primary concern from the review brief — is `export CARGO_HOME` with an empty value dangerous?

**No.** I traced this to `home-0.5.12/src/env.rs:68`:

```rust
match env.var_os("CARGO_HOME").filter(|h| !h.is_empty()) {
```

The `home` crate (used by cargo) explicitly filters out empty strings before using the value, so `export CARGO_HOME` with `@CARGO_HOME@` expanding to an empty string is equivalent to not setting the variable at all. Cargo falls back to `~/.cargo`. The unconditional `export CARGO_HOME` is safe.

### All three configure.ac files — consistent ✓

`rpkg/configure.ac:254`, `minirextendr/inst/templates/rpkg/configure.ac:270`, `minirextendr/inst/templates/monorepo/rpkg/configure.ac:218` all have identical four-line blocks in the right slot (after `AC_SUBST([ENV_RUSTFLAGS])`, before `AC_ARG_VAR([CARGO_BUILD_TARGET])`).

### All three Makevars.in files — consistent ✓

`CARGO_HOME = @CARGO_HOME@` at line 10 and `export CARGO_HOME` are present in all three.

### `rpkg/configure` regenerated ✓

`CARGO_HOME` appears in `ac_subst_vars` (lines 604 and 682), in `--help` output (line 1313), and the `: ${CARGO_HOME=""}` + conditional notice block is at line 2501–2507.

### `patches/templates.patch` ✓

The 18-line churn is entirely timestamps and hunk-line-number shifts (caused by the 6-line CARGO_HOME block being added identically to both sides of the diff). No spurious content delta.

### `docs/ENVIRONMENT_VARIABLES.md` ✓

Row added at the correct position — between `CARGO_BUILD_TARGET` (line 25) and `RUST_TOOLCHAIN` (line 27). Fits the existing three-column schema.

### Plan alignment ✓

The plan listed `export CARGO_HOME` at file level as the preferred option. The PR uses exactly that.

---

### Nit (cosmetic, non-blocking)

`rpkg/src/Makevars.in` (and both template copies): `export CARGO_HOME` lands on line 43, immediately *after* the `## ------` closing separator of the "filled by configure" block. It would read more cleanly just above that separator, adjacent to `ENV_RUSTFLAGS = @ENV_RUSTFLAGS@`. No functional impact.

---

No deferred items or follow-up issues needed — the change is self-contained and additive.

Issue #243, opened 2026-04-18
refactor(macros): retarget standalone-fn codegen at CWrapperContext


CGMossa:
## Context

`miniextendr-macros/src/lib.rs::build_fn_c_wrapper` hand-rolls three
variants of the `extern "C-unwind"` wrapper (main-thread × `{rng, no-rng}`,
worker-thread with error_in_r fused). `miniextendr-macros/src/c_wrapper_builder::CWrapperContext`
already does the same thing for impl methods via a builder.

Both paths use matching concepts (main vs worker thread strategy, rng
wrapping, unwind-protect selection, return-handling variants), but the
fn path was written first and diverged enough that unifying them during
the 2026-04-18 refactor sweep wasn't mechanical.

Unifying these two paths would:

- delete ~200 lines from `lib.rs` (the whole `build_fn_c_wrapper` helper
  + its `FnCWrapperInputs` struct),
- eliminate silent-divergence risk — today any fix in
  `CWrapperContext`'s wrapper emission has to be mirrored by hand in
  `build_fn_c_wrapper` (or vice versa), and the pre-sweep duplication
  survey found 47 shared quoted tokens between the two sites.

## Blockers / deltas to resolve

Route-through isn't free; each of these needs an explicit decision:

### 1. C-wrapper parameter name mangling

`CWrapperContext::build_c_params` renames SEXP parameters to
`arg_0, arg_1, …` in the C wrapper signature. The standalone-fn path
preserves the user's original identifiers, which are **visible in
rustdoc** (the C wrapper is `#[doc = ...]`-annotated and linked from
the Rust fn). Switching would be a user-observable rename.

Options:

- (a) Extend `CWrapperContextBuilder` with `preserve_param_names: bool`
  and plumb the preference through `build_c_params`.
- (b) Accept the rename (breaks links that reference `C_myfn(x: SEXP, y: SEXP)`
  by name in downstream docs).
- (c) Emit both names via a `#[doc(alias = \"…\")]` shim.

(a) is the safest; (b) is the cleanest if we're willing to absorb the
rustdoc churn.

### 2. Visibility forwarding

The fn path honors `#vis` from the user's source (so a `pub(crate) fn`
produces a `pub(crate) extern "C-unwind" fn C_myfn(…)`). `CWrapperContext`
always emits wrappers with the default (no vis) visibility. The builder
needs a `with_vis(syn::Visibility)` hook — new field + `.with_vis(vis)`
[...]

PR #540, opened 2026-05-14
feat(jiff): adapter-trait coverage for RDate/RZoned/RTimestamp methods + laziness assertion (#305)

CGMossa:
## Summary

Adds adapter-trait test coverage for the jiff phase 6+7 surface — methods that existed but had zero rpkg fixtures and zero testthat assertions — plus a laziness proof for `Vec<Timestamp>` ALTREP.

### Coverage added

- `RDate`: `weekday`, `day_of_year`, `first_of_month`, `last_of_month`, `tomorrow`, `yesterday`
- `RZoned`: `start_of_day`, `strftime`
- `RTimestamp`: `strftime`, `as_millisecond`
- Laziness: `OnceLock<AtomicUsize>` materialization counter exposed via a debug fixture; R-side asserts count is 0 on construction, 0 on length query, ≥1 only after first element access.

### Note on the wasm_registry.rs change (+11k lines)

`rpkg/src/rust/wasm_registry.rs` was committed as a stub when wasm32 support landed (`ca2b2e2d`). No PR since then ran `just rcmdinstall`, so it was never populated. This PR is the first run that regenerates it; the diff captures the entire existing rpkg registry, not just the 12 new jiff entries. The file is tracked, generated by `miniextendr_write_wasm_registry()` from the cdylib pass during `just rcmdinstall`, and a wasm32 build with the stub would register zero routines — so this is a correct fix-side-effect, not noise.

### Implementation notes

- Trait methods dispatched via `<Type as Trait>::method(&val)` because `jiff::civil::Date::weekday()` etc. exist as inherent methods that would shadow trait dispatch.
- `OnceLock<AtomicUsize>` for the counter (stable since Rust 1.70, project MSRV is recent).
- Generated files committed in sync: `R/miniextendr-wrappers.R`, `NAMESPACE`, `man/*.Rd`.

## Test plan

- [x] `just devtools-test` — 121 PASS, 0 FAIL, 0 WARN, 0 SKIP
- [x] Generated wrappers + NAMESPACE + man committed in sync (pre-commit hook satisfied)
- [ ] CI green

Closes #305.

🤖 Generated with Claude Code (Sonnet implementer, Opus planner/reviewer pipeline).

PR #541, opened 2026-05-14
feat(macros): forward prefer = "list"/"externalptr"/"native" to C wrapper codegen (#358)

### LINKED ISSUE #358: feat: forward ReturnPref::List/ExternalPtr/Native to CWrapperContext
## Background

`#[miniextendr(return = "list")]`, `return = "externalptr"`, and `return = "native"` parse into `ReturnPref::List/ExternalPtr/Native` (in `miniextendr_fn.rs`) but the new `CWrapperContext`-based standalone-fn codegen path introduced in #243 ignores these variants.

The old `analyze_return_type` path wrapped the return value in `AsList(v)`, `AsExternalPtr(v)`, or `AsRNative(v)` before calling `into_sexp`. The new `CWrapperContext` path uses `ReturnHandling::IntoR` which calls `into_sexp` directly.

## Impact

Only affects standalone `#[miniextendr]` functions that explicitly specify `return = "list"`, `"externalptr"`, or `"native"`. **No rpkg usage exists** — no regressions in existing tests.

## Fix

Add a `ReturnHandling` variant (or a builder option) to `CWrapperContext` that wraps the call result in `AsList`/`AsExternalPtr`/`AsRNative` before calling `into_sexp`. Route `return_pref` from `MiniextendrFnAttrs` into the builder call in `lib.rs`.

## Deferred from

PR #243 (issue #243) — see `plans/issue-243-cwrapper-unification.md`.

CGMossa:
## Summary

- Removes the `return_pref: _` deferral at `lib.rs:720` (deferred from PR #243).
- Adds `ReturnHandling::AsListOf`, `AsExternalPtrOf`, `AsNativeOf` variants to
  `c_wrapper_builder.rs`; each wraps `__result` in `AsList`/`AsExternalPtr`/`AsRNative`
  before calling `IntoR::into_sexp`. Both main-thread and worker-thread codegen
  paths handle the three new variants.
- Routes `ReturnPref` from `MiniextendrFnAttrs` through a new `apply_return_pref`
  helper that substitutes `IntoR` with the appropriate variant when `prefer =` is set.
  `Option*`/`Result*`/`Unit`/`RawSexp`/`ExternalPtr` variants are left unchanged —
  wrapping would be semantically wrong or there is no plain value to wrap.
- Removes `#[allow(dead_code)]` on `return_pref` field in `MiniextendrFnAttrs`.
- Adds three attribute-form fixture functions to `rpkg/src/rust/convert_pref_tests.rs`
  (`attr_prefer_list`, `attr_prefer_externalptr`, `attr_prefer_native`) and three
  corresponding testthat assertions validating they match the explicit wrapper forms.
- Regenerates `R/miniextendr-wrappers.R`, `NAMESPACE`, `man/convert_pref_tests.Rd`,
  and `wasm_registry.rs`.

Only standalone `#[miniextendr]` functions are affected. No change to impl-method or
trait-ABI codegen. `prefer = "auto"` (default) is bit-identical to previous behaviour.

Closes #358.

## Test plan

- [ ] `cargo check --workspace --all-targets` — clean
- [ ] `cargo test --workspace` — clean
- [ ] `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean
- [ ] `cargo clippy ... --features <CI feature list> -- -D warnings` — clean
- [ ] `R CMD INSTALL rpkg` — compiles, generates wrappers, installs OK
- [ ] `devtools::test('rpkg', filter='convert-pref')` — 17 tests pass (14 pre-existing + 3 new)
- [ ] Full `devtools::test('rpkg')` — all tests pass (pre-existing altrep-serialization-cross-session failures are environment-only, unrelated to this change)


CGMossa (commented):
## Code Review — PR #541 (`prefer = "list"/"externalptr"/"native"` codegen)

Overall the implementation is clean and the core happy-path is correct. A few issues to address:

### Issues

**1. `OptionIntoR`/`OptionIntoRUnwrap` mapping deviates from plan (semantic concern)**

`apply_return_pref` in `lib.rs:1563-1580` maps `OptionIntoR | OptionIntoRUnwrap` to the new `As*Of` variants. The plan explicitly says "the simpler and correct approach for the first pass: **only apply wrapping to `ReturnHandling::IntoR`**; leave `Option*` variants unchanged." For standalone functions, `detect_return_handling_standalone_fn` always produces `OptionIntoR` for `-> Option<T>`. So `#[miniextendr(prefer = "list")] fn foo() -> Option<Hybrid>` generates `AsList(Option<Hybrid>)`, which requires `Option<Hybrid>: IntoList` — almost certainly absent. The user gets an opaque trait-bound error rather than the `prefer=` attribute being silently ignored on an incompatible return type. The docstring comment at `lib.rs:1550-1553` acknowledges this but describes it as intentional. Decide whether this is the desired UX and document it in a `# Notes` section in the `apply_return_pref` doc comment, or follow the plan and restrict mapping to `IntoR` only.

**2. `unwrap_in_r` + `prefer=` silently ignored — misleading comment**

`lib.rs:938-948`: when `unwrap_in_r && output_is_result(output)` is true, `apply_return_pref` is skipped entirely and `ReturnHandling::IntoR` is returned. The comment at line 944 says "Only meaningful when the auto-detected handling is IntoR" but this branch never calls `apply_return_pref`, so `prefer=` is silently a no-op on `unwrap_in_r` functions. The comment implies both that `prefer=` works on `IntoR` variants and that these are from `auto`-detection — neither is true here. Either add a note like `// Note: prefer= is silently ignored when unwrap_in_r overrides the return handling.` or handle this combination.

**3. Fixtures exercise the same type / no `Option<T>` fixture**

All three new fixtures (`attr_prefer_list`, `attr_prefer_externalptr`, `attr_prefer_native`) use `-> Hybrid` with `i32` input. Given that the `Option*` mapping is now live, there should be at least one fixture and test covering `-> Option<Hybrid>` + `prefer=` so the actual compiler behavior (trait-bound error or successful compilation) is demonstrated/documented. This also exercises the newly-live `OptionIntoR` path.

### Smaller nits

- `lib.rs:1565-1566`: `OptionIntoRUnwrap` appears in the `apply_return_pref` match arm, but `detect_return_handling_standalone_fn` converts it to `OptionIntoR` before this function is called — that arm is dead code for the standalone fn path. Harmless but worth a comment.
- The externalptr test (`test-convert-pref.R:28`) uses `typeof(result) == typeof(hybrid_as_ptr(1L))` instead of `expect_identical(result, hybrid_as_ptr(1L))`. This is correct (externalptrrs have distinct heap addresses), but the asymmetry with the list/native tests (`expect_identical(result, hybrid_as_list(1L))`, `expect_identical(result, hybrid_as_native(1L))`) looks odd to readers. A brief comment would help.

### Confirmed correct

- Three new `ReturnHandling` variants named consistently (`AsListOf`/`AsExternalPtrOf`/`AsNativeOf`).
- Both `generate_return_handling` and `generate_worker_return_handling` have exhaustive arms for all three new variants. Semantics correct: worker returns raw `T`, convert arm wraps in `AsList`/`AsExternalPtr`/`AsRNative` inside `unwind_fn`.
- `ReturnPref::Auto` is a no-op (returns `auto` unchanged) — bit-identical to pre-PR behavior.
- `wasm_registry.rs` regeneration is the same correct fix-side-effect as #540 (stub → live cdylib snapshot, content-hash was all zeros).
- R wrappers for three new fixtures are correctly shaped (same `.Call` + `error_in_r` check pattern as other wrappers).
- Cargo.lock: no churn.
- `#[allow(dead_code)]` removed from `return_pref` field in `MiniextendrFnAttrs`.

CGMossa (commented):
Review posted

CGMossa:
All three reviewer items addressed. Summary:

**Fix 1 — `OptionIntoR`/`OptionIntoRUnwrap` mapping restricted (reviewer item 1)**

`apply_return_pref` now only maps the plain `IntoR` variant to the new `As*Of` variants. `OptionIntoR` and `OptionIntoRUnwrap` pass through unchanged. The dead `OptionIntoRUnwrap` arm is deleted. Docstring updated with a `# Notes` section: "When `prefer=` is combined with `Option<T>` or `Result<T,E>` returns, only the plain-T fast path is affected — `Option`/`Result` wrappers behave as if `prefer=` were absent."

**Fix 2 — `unwrap_in_r` + `prefer=` documented as intentional no-op (reviewer item 2, decision: (b))**

Chose **(b)** — keep the no-op, add a clear comment. Rationale: in `unwrap_in_r` mode, `IntoR` operates on the whole `Result<T,E>` (the framework's `IntoR for Result` encodes it as a tagged list R decodes). Applying `prefer=` would produce `AsListOf(Result<T,E>)` requiring `Result<T,E>: IntoList` — absent. The outer `Result` wrap structurally prevents `prefer=` from reaching the inner `T`. The inline comment now says: "Note: `prefer=` is intentionally ignored when `unwrap_in_r` overrides the return handling. In `unwrap_in_r` mode the `IntoR` here operates on the whole `Result<T,E>` (the framework's `IntoR for Result` encodes it as a tagged list for R to decode), NOT on the inner `T`."

**Fix 3 — `Option<T>` no-op fixture added (reviewer item 3)**

Two new fixtures in `convert_pref_tests.rs`:
- `attr_prefer_list_option(x: Option<i32>) -> Option<i32>` with `prefer = "list"` — the `prefer=` is silently ignored
- `plain_option_i32(x: Option<i32>) -> Option<i32>` — baseline without `prefer=`

R test in `test-convert-pref.R` asserts both produce identical output, that `NULL` input → `NA_integer_` (not NULL — `Option<i32>` None maps to `NA_integer_`), and that the result is NOT a list (explicitly documenting that `prefer="list"` had no effect on the Option path).

All tests green: `[ FAIL 0 | WARN 0 | SKIP 17 | PASS 5797 ]`. Both `clippy_default` and `clippy_all` clean.

PR #542, opened 2026-05-14
feat(macros): detect inner-payload field collision in nested DataFrameRow enums (#486)

### LINKED ISSUE #486: DataFrameRow nested enum: inner-payload field named 'variant' collides with outer discriminant
## Background

#479 added compile-time collision detection for the case where a sibling field in the OUTER variant is named `<field>_variant`:

```rust
#[derive(DataFrameRow)]
enum Outer {
    // Compile error: kind_variant collides with discriminant
    Wrap { kind: Inner, kind_variant: String },
}
```

But there's a **second collision case** that is currently a documented carve-out: when the INNER enum's payload contains a field literally named `variant` (or whatever the inner's `#[dataframe(tag = "...")]` tag is, after outer prefixing).

```rust
#[derive(DataFrameRow)]
enum Inner {
    A,
    B { variant: i32 },     // <-- this field, after outer prefix, becomes `kind_variant`
}

#[derive(DataFrameRow)]
enum Outer {
    Wrap { kind: Inner },   // <-- outer discriminant ALSO becomes `kind_variant` — collision
}
```

This requires inter-macro communication (the outer macro must inspect the inner's payload field names, which it can't reach across modules at proc-macro time).

## Possible approaches

- Surface as a runtime panic when the column registry detects duplicate names (graceful but late).
- Add a separate `#[derive(DataFrameRowInfo)]` shadow trait that inner enums can opt into, exposing payload field names via `const PAYLOAD_FIELDS: &[&str]` for outer to inspect.
- Status quo: rely on documentation and the OUTER-side collision check (which catches the more common case).

No urgency.

CGMossa:
## Summary

- Introduces a `DataFramePayloadFields` trait (emitted by `#[derive(DataFrameRow)]`) exposing `FIELDS: &[&str]` (resolved column names) and `TAG: &str` (inner tag suffix).
- Outer macro emits a `const _: ()` assertion per nested-enum field that fires if any inner payload field name, when compared to the inner discriminant tag, would produce an identical column after outer prefixing.
- Adds `assert_no_payload_field_collision` const fn in `miniextendr-api::markers` — stable const eval, no nightly features.
- Replaces the documented "carve-out" for inner-payload-named-`variant` with a clean compile error pointing at the outer `#[derive(DataFrameRow)]`.
- Two new UI tests: default tag (`variant`) and custom tag (`kind`) collision cases.
- Updated two existing "no-derive" UI snapshots to include the additional `DataFramePayloadFields` bound error that now fires alongside the existing `DataFrameRow` bound error.

Note on the latent B1 fix: the plan called for using `Inner::TAG` in the B1 sibling check (line 272, currently hardcoded `"_variant"`). The B1 check already covers the sibling-field collision case; the new `assert_no_payload_field_collision` handles the inner-payload case without needing B1 to know the actual inner tag. The B1 comment was updated to clarify this split of responsibilities. Changing B1 to use `Inner::TAG` at the outer-macro parse time is not straightforward (it requires a new const assertion rather than a `syn::Error`, since B1's inner_ty is only a syn::Type). Tracked in #544.

## Closes

Closes #486

## Test plan

- [x] `cargo test -p miniextendr-macros --lib` — 307 tests pass, no regressions
- [x] `TRYBUILD=overwrite cargo test -p miniextendr-macros` — new `.stderr` snapshots reviewed, both show clear `E0080: evaluation panicked` errors at the outer `#[derive(DataFrameRow)]` site
- [x] `cargo check --workspace --all-targets` — compiles clean
- [x] `just configure && just rcmdinstall && just devtools-document` — builds clean, no NAMESPACE/wrappers changes
- [x] `just devtools-test` — 5750 pass, 0 fail
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean (default + all-features CI feature set)

## Follow-ups

- #544 — B1 sibling-collision check hardcodes `"variant"` instead of using `Inner::TAG`; needs a `const _: ()` assertion approach since `syn::Type` is not available at const-eval time.
- #545 — Multi-level nesting (outer enum → struct → inner enum) not covered by current UI tests; verify `assert_no_payload_field_collision` fires transitively and add a 3-level UI test fixture.

### DIFF
--- a/miniextendr-api/src/markers.rs
+++ b/miniextendr-api/src/markers.rs
@@ -38,6 +38,87 @@
 )]
 pub trait DataFrameRow {}
 
+/// Reflection trait for `#[derive(DataFrameRow)]` types, used for compile-time
+/// collision detection in outer `DataFrameRow` enums.
+///
+/// Automatically emitted by the `DataFrameRow` derive macro alongside `DataFrameRow`.
+/// The outer enum's codegen generates a `const _: ()` assertion that calls
+/// [`assert_no_payload_field_collision`] using these associated constants, producing
+/// a clear compile error when an inner payload field name (after outer prefix expansion)
+/// would produce the same R column as the outer discriminant column.
+///
+/// # Constants
+///
+/// - `FIELDS`: the resolved column names (after `#[dataframe(rename = "...")]`) that
+///   this type contributes directly. For structs: each column name. For enums: every
+///   payload field name across all variants (deduplicated).
+/// - `TAG`: the value of `#[dataframe(tag = "...")]` on this type, or `""` if absent.
+///   The outer macro uses this as the discriminant suffix for inner-enum nested fields.
+///
+/// You should not implement this trait manually.
+pub trait DataFramePayloadFields {
+    /// Resolved column names contributed by this type (post-rename, pre-prefix).
+    const FIELDS: &'static [&'static str];
+    /// Value of `#[dataframe(tag = "...")]` on this type, or `""` if absent.
+    const TAG: &'static str;
+}
+
+// region: const-eval collision helper
+
+/// Const-eval equality check for two `&str` values (stable since Rust 1.46).
+///
+/// Used by [`assert_no_payload_field_collision`] inside a `const {}` block.
+const fn const_str_eq(a: &str, b: &str) -> bool {
+    let a = a.as_bytes();
+    let b = b.as_bytes();
+    if a.len() != b.len() {
+        return false;
+    }
+    let mut i = 0;
+    while i < a.len() {
+        if a[i] != b[i] {
+            return false;
+        }
+        i += 1;
+    }
+    true
+}
+
+/// Compile-time assertion: no element of `fields` equals `discriminant_suffix`
+/// (the inner-enum's `#[dataframe(tag)]` value).
+///
+/// Called from `const _: ()` blocks emitted by the outer `DataFrameRow` derive for
+/// every `EnumResolvedField::Struct` nested-enum field:
+///
+/// ```rust,ignore
+/// const _: () = ::miniextendr_api::markers::assert_no_payload_field_collision(
+///     <Inner as ::miniextendr_api::markers::DataFramePayloadFields>::FIELDS,
+///     <Inner as ::miniextendr_api::markers::DataFramePayloadFields>::TAG,
+/// );
+/// ```
+///
+/// `fields` contains the resolved column names the inner type emits directly (not prefixed).
+/// `discriminant_suffix` is the inner enum's tag value (e.g. `"variant"`).
+///
+/// If any field name equals the discriminant suffix, the const evaluation panics with a
+/// message explaining the collision and suggesting a rename.
+pub const fn assert_no_payload_field_collision(fields: &[&str], discriminant_suffix: &str) {
+    let mut i = 0;
+    while i < fields.len() {
+        if const_str_eq(fields[i], discriminant_suffix) {
+            panic!(
+                "DataFrameRow inner-payload field collision: an inner enum payload field \
+                 has the same name as the inner enum's discriminant tag. After outer prefix \
+                 expansion this produces two columns with identical names, causing silent \
+                 data corruption. Rename the inner payload field or add \
+                 `#[dataframe(rename = \"...\")]` on it to use a different column name."
+            );
+        }
+        i += 1;
+    }
+}
+// endregion
+
 /// Marker trait for types that should be converted to R lists via `IntoR`.
 ///
 /// Implemented by the `PreferList` derive; you can also implement it manually.
--- a/miniextendr-api/src/markers.rs
+++ b/miniextendr-api/src/markers.rs
@@ -56,6 +56,7 @@ pub trait DataFrameRow {}
 ///   The outer macro uses this as the discriminant suffix for inner-enum nested fields.
 ///
 /// You should not implement this trait manually.
+#[doc(hidden)]
 pub trait DataFramePayloadFields {
     /// Resolved column names contributed by this type (post-rename, pre-prefix).
     const FIELDS: &'static [&'static str];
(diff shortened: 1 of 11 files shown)


CGMossa (commented):
## Review findings for #542 (issue #486 — inner-payload collision detection)

### Summary

The core mechanism is sound and well-tested. Two correctness gaps and one missing `#[doc(hidden)]` need addressing before merge.

---

### Issues

**1. B1 still hardcodes `"variant"` — plan explicitly required using `Inner::TAG` (BLOCKER)**

`enum_expansion.rs:275`:
```rust
let discriminant_col = format!("{}_variant", struct_data.base_name);
```
The plan (lines 237, 317) states this PR "also fixes the latent bug in B1 (hardcoded `_variant` suffix) by using `Inner::TAG` instead." The PR description echoes this. But the implementation leaves B1 unchanged and only updates the comment to say the hardcoding is intentional. This is a scope-cut without a follow-up issue, which violates the project rule "deferred items = GitHub issues." Either (a) fix B1 to emit a const assertion using `Inner::TAG` (similar to the payload collision assertions), or (b) open a tracked issue for it and reference it in the PR body. Note: the new `assert_no_payload_field_collision` does NOT cover the B1 case — B1 catches `Outer` having a sibling field named `kind_variant` when `kind: Inner` is flattened, which is a different collision from what the payload assertion catches.

**2. `DataFramePayloadFields` missing `#[doc(hidden)]` — leaks as user-visible trait**

`miniextendr-api/src/markers.rs:59`: `DataFramePayloadFields` has no `#[doc(hidden)]`. The trait is an internal coordination mechanism (docs say "You should not implement this trait manually") but will appear in rustdoc and user autocomplete. Mark it `#[doc(hidden)]`. Same applies to `const_str_eq` (already `const fn` not `pub`, so it's fine) and `assert_no_payload_field_collision` (this one arguably could stay visible — it documents itself well — but `DataFramePayloadFields` itself is noise at the public API surface).

**3. `FIELDS` for enums omits `as_list` and `as_factor` fields (potential false-negative)**

`enum_expansion.rs:1584`: The `payload_fields_impl` loop only collects `EnumResolvedField::Single` entries. Fields annotated `#[dataframe(as_list)]` and `#[dataframe(as_factor)]` also resolve to `Single` (confirmed at lines 82–113), so they ARE included — no bug here. However, `EnumResolvedField::ExpandedFixed`, `ExpandedVec`, `AutoExpandVec`, and `Map` fields are silently omitted. An expanded-fixed field with base name `"variant"` would not be caught. Low risk in practice (unlikely naming), but worth noting if a follow-up issue is opened.

**4. Multi-level nesting not tested (minor, follow-up issue)**

The plan notes multi-level (outer enum → struct → inner enum → payload field) as a documented limitation. No follow-up issue was opened. This is fine to defer, but per project rules it needs a tracked issue.

**5. `rpkg` fixture update missing (minor)**

The plan says: "Add a comment-only note (or a renamed fixture) that the inner-payload-field-named-`variant` pattern is now a compile error." `dataframe_enum_payload_matrix.rs` was not updated with this note. Minor, but the plan explicitly called for it.

---

### Non-issues (verified)

- **Generic propagation**: `impl_generics`/`ty_generics`/`where_clause` correctly propagated in both struct and enum `DataFramePayloadFields` impls — confirmed at `dataframe_derive.rs:1736+` and `enum_expansion.rs:1594`.
- **Post-rename column names**: `FIELDS` uses `data.col_name` (which is already the resolved/renamed name from `fa.rename.unwrap_or_else(...)`) — correct.
- **`Inner::TAG` default**: `attrs.tag.as_deref().unwrap_or("")` emits `""` when absent, and `const_str_eq(field, "")` is always false — safe.
- **`struct_cols` scope**: assertions are correctly generated only for `EnumResolvedField::Struct` fields (not `as_list`/`as_factor` Single fields) — matches the intended check.
- **UI test quality**: both new tests exercise realistic patterns (default and custom tags). The E0080 error message at the outer `#[derive(DataFrameRow)]` span is clear and actionable.
- **Existing snapshot updates**: the two updated `.stderr` files gain a second `E0277: DataFramePayloadFields not satisfied` error alongside the pre-existing `DataFrameRow` bound error — this is expected since `assert_no_payload_field_collision` requires the new bound. The error is helpful (compiler points at the missing derive site).
- **`wasm_registry.rs`**: large regenerated file, consistent with the fix-side-effect stub-repopulation pattern from #540/#541. Not a code change.

---

### Verdict: NEEDS-REWORK

Required before merge:
1. Open a GH issue for B1 tag-awareness (the `"variant"` hardcode) and reference it in the PR body.
2. Add `#[doc(hidden)]` to `DataFramePayloadFields` in `markers.rs`.

B1 scope-limit is **acceptable** provided a tracked issue exists — the new `assert_no_payload_field_collision` covers the primary issue (#486's stated scenario). The existing B1 check still works for the default-tag case. A const assertion for B1 using `Inner::TAG` would be a clean follow-up, not a blocker.

CGMossa:
## Review items addressed

**B1 follow-up issue filed (blocker resolved):** #544 — "DataFrameRow nested enum B1 sibling-collision check hardcodes `'variant'` instead of using `Inner::TAG` (#486 follow-up)". The hardcode at `enum_expansion.rs:275` is deferred per reviewer guidance; #544 documents the root cause, why a `syn::Error`-time fix isn't straightforward, and what a `const _: ()` assertion approach would look like. Referenced in PR body under "Follow-ups".

**`#[doc(hidden)]` added (blocker resolved):** Both `DataFramePayloadFields` trait and `assert_no_payload_field_collision` const fn in `miniextendr-api/src/markers.rs` now carry `#[doc(hidden)]`. Neither should surface in user-facing rustdoc or autocomplete.

**rpkg fixture comment added (minor):** `dataframe_enum_payload_matrix.rs` region 10 now carries the note: _"the inner-payload-field-named-`variant` pattern is now a compile error enforced via `assert_no_payload_field_collision` — see #486 / PR #542"_.

**Multi-level nesting follow-up issue filed (minor):** #545 — "DataFrameRow: verify collision check fires at all nesting levels (#486 follow-up)". Describes the 3-level test case and asks for a UI test fixture. Referenced in PR body under "Follow-ups".

No logic changes — only `#[doc(hidden)]` attributes and a comment.

Issue #484, opened 2026-05-10
DataFrameRow: detection limits — type aliases and Option<Map>/Option<Inner> fall through silently


CGMossa:
### Verification — issue still applicable?

Grep on `miniextendr-macros/src/dataframe_derive.rs:290-389` (`classify_field_type`) confirms the May-10 sprint shipped HashMap/BTreeMap/Struct support but no Option/alias handling. The three sub-bullets in the issue body resolve as follows:

| Sub-bullet | Status | Evidence |
|---|---|---|
| 1. Type aliases (`type Counts = HashMap<…>; field: Counts`) | **still-applies** | `dataframe_derive.rs:369-387`: the trailing path branch matches `PathArguments::None` ident not in `KNOWN_SCALARS` → returns `FieldTypeKind::Struct { inner_ty: ty }`. The compile-time `_assert_inner_is_dataframe_row::<Counts>()` then fails with a confusing "Counts does not impl DataFrameRow" error rather than something pointing at the alias. The issue's claim "falls through to Scalar" is now outdated since PR #477 — it actually falls through to `Struct`, but the symptom (silent misclassification) is the same. |
| 2. `Option<HashMap<K,V>>` / `Option<InnerEnum>` | **still-applies** | `dataframe_derive.rs:320-346`: the HashMap/BTreeMap/Vec/Box arm runs on the LAST path segment. For `Option<HashMap<…>>` the last segment is `Option` with one generic arg → matched by `seg.ident == "Vec"` etc. none of which succeed → falls through. The non-generic path branch (`dataframe_derive.rs:369-387`) requires `PathArguments::None`, so `Option<…>` (which has `AngleBracketed` args) falls all the way through to `FieldTypeKind::Scalar`. No diagnostic. |
| 3. Fully-qualified-path edge cases | **partially-fixed** | `dataframe_derive.rs:372`: `type_path.path.leading_colon.is_none()` already filters absolute paths like `::std::ffi::CString`. The doc comment at `dataframe_derive.rs:358-368` documents this as a deliberate carve-out. Remaining gap: `qself` is filtered (l. 372) but non-leading-`::` multi-segment paths with generic args on the last segment (`crate::map::HashMap<K,V>`) work today via the last-segment ident match at l. 326-345. So this bullet is mostly already handled — only nested generics outside the canonical idents miss. Not the core complaint. |

### Punchline rule (per `feedback_issue_punchline_over_options.md`)

[...]

PR #519, opened 2026-05-11
docs: drop stale --whole-archive / --freeze descriptions

CGMossa:
## Summary

- **D1** — Remove stale `(links via -Wl,-force_load on macOS / --whole-archive on Linux)` parenthetical from CLAUDE.md's "How the build works" diagram. Current Makevars.in uses `stub.c` + `codegen-units = 1` as the force-link anchor; no `-force_load` flag is emitted (per Makevars.in:24-27). Replace with a two-sentence description of the actual mechanism.
- **D1 bonus** — Fix the stale "Use `cargo revendor --freeze` at vendor time" principle in CLAUDE.md (`--freeze` was removed from `just vendor` — see `docs/CRAN_COMPATIBILITY.md:221-228`).
- **D2** — Delete the "Stale frozen-vendor recovery" subsection from CLAUDE.md, and the equivalent "Stale-freeze recovery" from `cargo-revendor/CLAUDE.md`. The recovery scenario no longer arises (`--freeze` is not invoked anywhere in current code).
- **docs/R_BUILD_SYSTEM.md** — Remove stale `-force_load librpkg.a` from the Build Flow Summary diagram.

`AGENTS.md` mirror **left for a follow-up.** That file is currently untracked in the repo; the local copies contain user-specific paths that should not be committed. Tracking it (and what content to ship) is a separate decision.

Historical references in `docs/CRAN_COMPATIBILITY.md` are intentionally preserved (authoritative removed-things log).

Investigation refs: §5.4, §7.3 in `analysis/build-system-investigation-2026-05-11.md`.

## Test plan

- [x] `grep -n 'whole-archive\|force_load' CLAUDE.md cargo-revendor/CLAUDE.md docs/R_BUILD_SYSTEM.md` — only matches in new explanatory text ("no `-force_load` / `--whole-archive` needed"), not as a claimed current flag
- [x] `grep -n 'frozen-vendor\|Stale-freeze' CLAUDE.md cargo-revendor/CLAUDE.md` — no matches
- [x] `grep -n 'Use.*--freeze\|freeze.*vendor time' CLAUDE.md cargo-revendor/CLAUDE.md` — no matches
- [x] Visual review of updated sections confirms no broken cross-references

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #543, opened 2026-05-14
fix(macros): classify_field_type accepts multi-segment paths (#514)

### LINKED ISSUE #514: DataFrameRow: classify multi-segment paths as struct fields (flatten for #477 + #485)
## Background

`miniextendr-macros/src/dataframe_derive.rs::classify_field_type` only recognises **bare-ident** non-known-scalar paths as `FieldTypeKind::Struct`:

```rust
if segs.len() == 1 && type_path.qself.is_none() && type_path.path.leading_colon.is_none() {
    let seg = segs.last().unwrap();
    if matches!(seg.arguments, syn::PathArguments::None) {
        let name = seg.ident.to_string();
        if !KNOWN_SCALARS.contains(&name.as_str()) {
            return FieldTypeKind::Struct { inner_ty: ty };
        }
    }
}
```

Multi-segment paths fall through to `Scalar`. So:

```rust
#[derive(DataFrameRow)]
struct Located {
    id: i32,
    origin: crate::geom::Point,   // NOT classified as Struct
}
```

…doesn't flatten — the user has to import `Point` into the local scope or use `#[dataframe(as_list)]`. Same limitation affects enum-variant struct-field flattening (#477) since both use the shared classifier.

## Fix sketch

After all other kinds have been ruled out, treat `Type::Path` with a bare-ident **last segment** (no generic args, ident not in `KNOWN_SCALARS`) as `FieldTypeKind::Struct` regardless of path segment count. Order matters — `HashMap` / `BTreeMap` / `Vec` / `Box` already match by last-segment ident earlier, so reaching the new branch implies no collection match.

False-positive risk to keep in mind:
- `std::ffi::CString` would become `Struct` under the naive extension. Add to `KNOWN_SCALARS` (with namespace-aware handling) or accept that users with stdlib types as fields need `#[dataframe(as_list)]`.
- Type aliases via `use foo::Bar;` — these resolve at compile time, not in the proc macro. A bare ident `Bar` is already handled.

## Acceptance

- New rpkg fixture: an outer `DataFrameRow` struct (and an enum variant) with a multi-segment-path inner type, flattening as expected.
- New UI tests in `miniextendr-macros/tests/ui/` covering the false-positive cases that should still error (e.g. `field: std::ffi::CString` without `as_list`) — or document that they're now treated as struct flatten with a clear error from the `_assert_inner_is_dataframe_row` check.
- Update `docs/CONVERSION_MATRIX.md` if it mentions the bare-ident restriction.

Priority: low — bare-ident inner types cover the typical case.

Touches the shared classifier — affects both `DataFrameRow`-struct (#485) and `DataFrameRow`-enum (#477) paths in one shot.

CGMossa:
## Summary

- `classify_field_type` previously required `segs.len() == 1` for the Struct-flatten branch, so `field: crate::mod::Point` silently degraded to a scalar column instead of flattening. Fixed by dropping the `segs.len() == 1` guard and matching on `segs.last()`, consistent with how Vec/Box/Map detection already works.

- `docs/DATAFRAME.md` "Detection caveats" updated: multi-segment path caveat removed (now works); `Option<HashMap>` and type-alias caveats retained as known gaps (tracked under #484); a new note explains that stdlib multi-segment types without `DataFrameRow` (e.g. `std::ffi::CString`) now get a clear compile-time error instead of silently becoming a scalar column.

- New rpkg fixture `QualLocated` with a `geom::QualPoint` field validates the fix end-to-end, including a new `test_that` block in `test-dataframe-struct-flatten.R`.

- New UI test `derive_dataframe_struct_field_multisegment_no_derive` captures the "better error" for stdlib multi-segment types.

## Risk: last-segment name shadowing

A user type whose last path segment happens to be named after a known-scalar (e.g. `mymod::String`) still correctly falls through because of the `KNOWN_SCALARS` check. A user type named `mymod::Option`, `mymod::Vec`, etc. would be misclassified by the existing Vec/Box/Map detection above this block — accepted per Rust naming convention. `#[dataframe(as_list)]` is the documented escape hatch for corner cases.

## New "better error" for stdlib multi-segment types

With `segs.len() == 1` removed, `std::ffi::CString` is now classified as `Struct`. The compile-time `_assert_inner_is_dataframe_row::<CString>()` assertion fires with a clear message rather than producing a silent scalar column. This is intentional and an improvement — the UI test captures the exact diagnostic.

## Deferred

Type aliases (`type Counts = HashMap<…>`) and `Option<HashMap>` / `Option<Struct>` detection remain out of scope — these require new `OptionMap` / `OptionStruct` FieldTypeKind variants and companion codegen (~250 additional lines). See #484.

Closes #514
Refs #484


CGMossa (commented):
## Review notes

### Core fix — correct and complete

`miniextendr-macros/src/dataframe_derive.rs` lines 371–372: `segs.len() == 1` guard is removed, replaced with `segs.last().unwrap()` + `PathArguments::None` on the last segment. The comment block is updated. No residual `segs.len() == 1` code guard exists in the PR branch (only the comment in line 359 references it as historical context).

The rpkg fixture (`QualLocated` / `geom::QualPoint`) correctly exercises `geom::QualPoint` — a genuinely multi-segment path. The compile-time shape assertion in the `const _: ()` block will catch regressions before any R-level test runs. R-side testthat block in `test-dataframe-struct-flatten.R` verifies colnames and values end-to-end. This matches the plan exactly.

### Issue 1 — UI test `.stderr` snapshot is fragile (medium risk)

`miniextendr-macros/tests/ui/derive_dataframe_struct_field_multisegment_no_derive.stderr` includes the `note: if you're trying to build a new CString consider using one of the following associated functions:` block with five method signatures from `$RUST/alloc/src/ffi/c_str.rs`. `trybuild` does exact matching; rustc occasionally changes which associated functions appear in this help note across minor stable releases. If a new constructor is stabilised on `CString` (e.g., `from_bytes_until_nul` was added in 1.69), CI would go red with a snapshot mismatch unrelated to the macro fix.

The second error (`E0277: DataFrameRow is not implemented for CString`) is stable and tests the real intent. The `E0599` error about `to_dataframe` not found on `CString` is also stable. The fragility is in the six-line stdlib method-suggestion block between them.

Simplest fix: change the UI test's type from `std::ffi::CString` to a local non-implementing type in the same test file (e.g., `struct NotDfr; struct Wrapper { label: inner_mod::NotDfr }`). This exercises the same macro path without relying on stdlib method suggestions. The `_assert_inner_is_dataframe_row` error (`E0277`) would still fire and the snapshot would be stable across toolchain versions.

### Issue 2 — docs caveats omit `Option<Inner>` (minor)

`docs/DATAFRAME.md` Detection caveats lists `Option<HashMap<K,V>>` as unhandled under #484, but doesn't mention `Option<Inner>` where `Inner` implements `DataFrameRow`. That case is also silently scalar (same underlying cause, same #484 fix). A user trying `pub origin: Option<geom::QualPoint>` after reading this PR's fixture would expect flattening to work, but it silently degrades. One bullet worth adding:

> - **`Option<UserStruct>`**: same issue — `Option<geom::QualPoint>` falls through to `Scalar`. Tracked under [#484](https://github.com/A2-ai/miniextendr/issues/484).

### Pre-existing site drift — acceptable

`site/content/manual/environment-variables.md` gained the `CARGO_HOME` row from PR #539 (commit `d66a6b8e` updated `docs/ENVIRONMENT_VARIABLES.md` but never ran `just site-docs`). `site/content/manual/r-build-system.md` dropped `-force_load` from PR #519 (same reason). Both are exact mirrors of source-doc changes already on main. This is legitimate stale-regen caught during `just site-docs`; acceptable per "fix warnings you see" rule.

### wasm_registry.rs — expected side-effect

The regenerated `wasm_registry.rs` (stub → live snapshot) adds the new `C_qual_located_basic` entry along with all other accumulated exports. This is the same fix-side-effect pattern as #540/#541/#542. Content-hash changes from `0000000000000000` to `abd936561f2fb3ab`. Correct.

### Scope verdict

PR stays scoped to #514. Option-wrapping (#484 deeper) is correctly deferred. Type-alias caveat retained unchanged. All plan requirements for the #514 portion are met. The two issues above are the only blockers before merge:
1. UI test snapshot fragility (medium — will cause spurious CI failures on toolchain updates)
2. Missing `Option<UserStruct>` caveat in docs (minor — documentation gap)

CGMossa:
Addressed both reviewer issues in commit fe6646f4.

**Issue 1 — UI snapshot stabilised**

Replaced `std::ffi::CString` with a local `#[derive(Debug, Clone)] pub struct NotADataFrameRow` inside a `mod local { }` block. The old snapshot was 50 lines and included a six-line stdlib block (`note: if you're trying to build a new CString consider using one of the following associated functions:` with five method signatures pointing into `$RUST/alloc/src/ffi/c_str.rs`). The new snapshot is 35 lines and contains only references to lines within the test file itself — no stdlib paths, no version-dependent content. Multi-segment path coverage is identical (`local::NotADataFrameRow`). Tests: `cargo test -p miniextendr-macros` — 308 passed.

**Issue 2 — `Option<UserStruct>` caveat added**

Added bullet under Detection caveats in `docs/DATAFRAME.md`:
> **`Option<UserStruct>`** where `UserStruct: DataFrameRow` — silently degrades to Scalar instead of recursively flattening. Tracked under [#484](https://github.com/A2-ai/miniextendr/issues/484).

Regenerated `site/content/manual/dataframe.md` via `scripts/docs-to-site.sh` — only `dataframe.md` drifted.

PR #546, opened 2026-05-14
feat(macros): DataFrameRow auto-emit IntoR for generic unit-only enums (#487)

### LINKED ISSUE #487: DataFrameRow: auto-emit IntoR for generic unit-only enums (currently carved out)
## Background

#479 auto-emits `impl IntoR for Inner` (producing a real R factor) when:

1. Inner is an enum, AND
2. ALL variants are unit (no payload), AND
3. **Inner has no generic parameters**

The third condition is a carve-out documented at `miniextendr-macros/src/dataframe_derive/enum_expansion.rs:1576-1582`:

```rust
// Skip auto-emission for generic unit enums — codegen for `impl<T> IntoR for Inner<T>`
// adds complexity for an uncommon case (generic enums are rare as DataFrameRow inner fields).
```

## What this means in practice

```rust
#[derive(DataFrameRow)]
enum Status<T> { Idle, Running, Done }     // <-- generic unit enum

// As a nested-enum field with #[dataframe(as_factor)]:
//   compile error — Inner: IntoR not satisfied.
//   User must hand-impl IntoR or remove the type parameter.
```

## Why deferred

Hand-rolling `impl<T: ...> IntoR for Inner<T>` requires correctly propagating bounds (`Inner<T>: ToString`?) and considering coherence with the existing `IntoR` blanket impls. The cost-benefit for an uncommon shape was deferred.

## Workaround

Drop the generic parameter, or hand-impl `IntoR` for the concrete instantiation.

No urgency.

CGMossa:
## Summary

- Removes the `impl_generics.to_token_stream().is_empty()` guard at
  `enum_expansion.rs` that blocked `UnitEnumFactor`/`IntoR`/`IntoList`
  auto-emission for generic unit-only enums.
- Non-generic unit enums: unchanged — still cache levels SEXP via `OnceLock`.
- Generic unit enums: emit the same three impls with `#impl_generics` / `#ty_generics`
  / `#where_clause` propagated; `IntoR` uses `build_levels_sexp` + `Rf_protect` on
  each call (generic statics not allowed in Rust — same pattern as
  `FactorOptionVec<T>: IntoR` in `miniextendr-api/src/factor.rs`).
- Also relaxes the top-level guard in `dataframe_derive.rs` to allow type/const
  params on unit-only enums, and fixes two related issues: (1) adds `PhantomData<T>`
  to the companion struct when type params are present and there are no other fields
  (E0392 fix); (2) adds explicit `Vec<(&str, SEXP)>` annotation in the `else` branch
  of `into_data_frame` for the empty-vec unit-only-no-tag case (E0282 fix).
- Adds `Signal<const VERSION: usize>` const-generic unit-only enum fixture in
  `dataframe_examples.rs` + `make_signal_factor()` exported fn + testthat coverage.
- Adds `tests/ui/pass/derive_dataframe_generic_unit_enum.rs` — trybuild compile-pass
  test for generic unit-only enum; documents why type-param enums use const-param
  proxy (Rust E0392 blocks type-param unit enums at user-definition level).

## Test plan

- [x] `cargo test -p miniextendr-macros` (307 lib + 1 integration = 308 passed)
- [x] `cargo test --test ui -p miniextendr-macros` — compile-fail + new compile-pass test pass
- [x] `cargo check --workspace --all-targets`
- [x] `cargo clippy_default` — no issues
- [x] `cargo clippy_all` — no issues
- [x] `R CMD INSTALL` succeeded
- [x] `devtools::test()` — 5753 PASS, 0 FAIL, 17 SKIP (all pre-existing skips)
- [x] Non-generic enum tests confirmed passing bit-identically (same OnceLock path)

## Notes

- Const-generic unit enums (`enum Foo<const N: usize> { A, B }`) compile in Rust and exercise the generic branch. Pure type-generic unit enums (`enum Foo<T> { A, B }`) would require an upstream proc-macro to generate them (Rust's E0392 prevents hand-writing them), but the macro now handles them correctly when they appear. The `phantom_field.is_empty()` predicate is documented inline.

## Follow-ups

- #548 — Add turbofish helper to `from_raw_pairs` to eliminate `Vec<(&str, SEXP)>` annotation drift (two sites: `vctrs_derive.rs:965` and `enum_expansion.rs`)

Closes #487


CGMossa (commented):
## Review findings

### Deviation (a) — Top-level guard in `dataframe_derive.rs` relaxed: ACCEPT

The top-level guard in `dataframe_derive.rs` is a hard early-return that fires **before** `derive_enum_dataframe` is even called (line ~697 on the PR branch). Without relaxing it, generic unit-only enums would be rejected with a compile error before `enum_expansion.rs` ever runs. The relaxation is correctly scoped: it only lets through enums where every variant is `Fields::Unit`, and still rejects structs/unions and enums with payload variants. Not scope creep.

### Deviation (b) — `PhantomData<T>` in companion struct: ACCEPT with a gap note

The `PhantomData<T>` field is correctly emitted **only** when `type_params` is non-empty. Rust allows unused const params in struct definitions without `PhantomData`, so for `Signal<const VERSION: usize>` the struct has no phantom field and no E0392 — correct. Zero-size field, no `IntoR`/`SEXP` interaction. The `from_vec_impl` init (`_phantom: ::std::marker::PhantomData`) is consistent.

**One gap:** `phantom_field.is_empty()` is used as a proxy for "no phantom needed" (`enum_expansion.rs:1249`). This is correct when the struct truly has no `PhantomData`. But the condition is evaluated at macro expansion time on a `TokenStream` value — `TokenStream::new()` is always empty, and `quote! { ... }` is never empty. Works. No bug, just fragile. Worth a comment.

### Deviation (c) — `Vec<(&str, SEXP)>` annotation for empty `from_raw_pairs`: FOLLOWUP NEEDED

`vctrs_derive.rs:965` has the same `Vec<(&str, ::miniextendr_api::ffi::SEXP)>` explicit annotation for the same reason. This is the second site. The PR comment correctly calls it a "pre-existing E0282 bug exposed by new code path." Both the vctrs and enum expansion callers hit the same root cause: Rust can't infer the element type of an empty `vec![]` passed to `from_raw_pairs`. A turbofish on `from_raw_pairs::<(&str, SEXP)>(vec![...])` or a `from_raw_pairs_empty()` helper would eliminate the annotation need site-wide. Please open a follow-up issue.

### Const-generic vs type-generic: PARTIAL — fixture is correct, but issue asked for type params

`Signal<const VERSION: usize>` exercises `impl_generics.to_token_stream().is_empty() == false`, which is the same predicate that gates the new generic branch. The codegen is identical: `#impl_generics` / `#ty_generics` / `#where_clause` are all propagated correctly regardless of whether the generic is a const or type param.

However, the original issue `#487` specifically mentioned `enum Status<T> { Idle, Running, Done }` (a phantom type parameter). The plan notes that `enum Foo<T> { A, B }` is valid Rust (compiler allows unused type params in enum definitions). A type-generic fixture would make it unambiguous that the fix covers the issue's stated case. The const-generic fixture exercises the same macro code path, but it doesn't prove that `enum Tag<T> { A, B }` compiles through the new path without E0392 in the companion struct (requires `PhantomData`). **Recommend adding a type-generic compile-test in `tests/ui/` (as the plan called for) alongside the const-generic runtime fixture.** Without it, the type-generic path is covered by code inspection only, not by a test.

### PROTECT discipline: CORRECT

The generic `IntoR` body (`enum_expansion.rs:1693–1697` on PR branch):
```
let levels = Rf_protect(build_levels_sexp(LEVELS));
let result = build_factor(&[idx], levels);
Rf_unprotect(1);
Ok(result)
```
Protect count is balanced (1 protect, 1 unprotect). `result` is the return value — it doesn't need protection because it's immediately returned. Mirrors `factor.rs:581–584` exactly.

### OnceLock non-generic path: UNTOUCHED — CORRECT

The non-generic branch is byte-for-byte the same as before. `LEVELS_CACHE: OnceLock<SEXP>` caching path is unchanged. No regression risk.

### `where_clause` propagation: CORRECT

Both `UnitEnumFactor`, `IntoR`, and `IntoList` generic impls carry `#where_clause`. Trait bound propagation is correct.

### Minor: `wasm_registry.rs` is a 11k-line side-effect

As in prior wave-2 PRs — snapshot regeneration only, unrelated to the fix. No action needed.

---

**Summary of required actions before merge:**
1. Add a UI compile-test for the type-generic case `enum Tag<T> { A, B, C }` (plan file `tests/ui/` section) to close the gap between the fixture and the issue's stated use case.
2. Open a follow-up issue for the `Vec<(&str, SEXP)>` annotation pattern (two sites now: `vctrs_derive.rs:965` and `enum_expansion.rs` new line) — turbofish or empty helper would eliminate future drift.

CGMossa:
Addressed both reviewer items in the latest commit:

**1. Type-generic UI compile-pass test (BLOCKER)**

Added `miniextendr-macros/tests/ui/pass/derive_dataframe_generic_unit_enum.rs` — a trybuild `t.pass()` test (tests/ui.rs now calls both `compile_fail` and `pass`).

**Why const-param instead of literal `enum Tag<T>`:** Rust E0392 forbids unused type parameters on enum definitions — `enum Tag<T> { A, B, C }` is rejected by the compiler *before* the proc-macro runs. Const params are exempt from E0392 and exercise the identical code path in `enum_expansion.rs` (same `impl_generics` non-empty branch, same `#impl_generics`/`#ty_generics`/`#where_clause` propagation). The test derives `Tag<const V: usize>` and exercises `TagDataFrame::<0>::from_rows(...)`, confirming the companion struct compiles.

The test includes inline documentation explaining the E0392 constraint so future readers understand why const-param is used rather than the type-param example from the issue.

**2. Follow-up issue filed: #548** — "Add turbofish helper to from_raw_pairs to eliminate Vec<( & str, SEXP)> annotation drift (#487 follow-up)". Covers both `vctrs_derive.rs:965` and `enum_expansion.rs`, proposes `from_raw_pairs_empty()` helper.

**Minor: `phantom_field.is_empty()` comment** — Added 10-line comment explaining the empty/non-empty cases and why the PhantomData-emission path is a defensive guard (Rust E0392 blocks the case it's guarding against at user-definition level).

All tests pass: `cargo test -p miniextendr-macros` (308), `clippy_default` and `clippy_all` clean, `cargo fmt --all` applied.

PR #547, opened 2026-05-14
feat(macros): from_rows_par for struct-flatten + as_list (#513)

### LINKED ISSUE #513: DataFrameRow: from_rows_par for struct-flatten + as_list-on-struct (#485 follow-up)
## Background

#485 lands struct-in-struct flattening on top-level `DataFrameRow` structs. The sequential path (`From<Vec<Row>>` / `from_rows`) works for both:

- Default struct flatten (companion holds `Vec<Inner>`)
- `#[dataframe(as_list)]`-on-struct opt-out (companion holds `Vec<List>`)

The **parallel** path (`from_rows_par`, behind `cfg(feature = "rayon")`) is gated off for both:

```rust
let from_rows_par_method = if (!flat_cols.is_empty()
    || !auto_expand_cols.is_empty()
    || has_tag)
    && !has_struct
    && !has_into_list_struct
{ ... }
```

## Why deferred

`from_rows_par` pre-allocates each column as `vec![<#ty as Default>::default(); len]` so a `ColumnWriter` can scatter rows in parallel. Neither `Inner` (an arbitrary user `DataFrameRow` struct) nor `miniextendr_api::List` implements `Default`, and forcing users to derive `Default` on every inner type just for rayon is too invasive.

## Approaches

1. **Sequential post-process for struct fields.** Scatter scalar columns in parallel as today; for each struct field, accumulate `Vec<Inner>` sequentially after the par loop joins. Minimal code change; parallel speedup degrades when struct fields dominate but stays correct.
2. **`MaybeUninit<T>`-based scatter.** `vec![MaybeUninit::<Inner>::uninit(); len]`, write each cell exactly once via `ColumnWriter`, then `assume_init` after the par loop joins. Fully parallel but adds `unsafe` and bookkeeping for partial-init safety on panic.
3. **Opt-in `Default` bound.** Generate `from_rows_par` only when the user separately opts in (e.g. `#[dataframe(par)]`) — moves the constraint into user-space.

Recommend (1) as the default landing — clean, no `unsafe`, easy to reason about. (2) is the right ceiling if someone needs full parallelism for nested DataFrame types.

## Acceptance

- `FlatLocated::to_dataframe_par(rows)` (or equivalent) returns the same companion as sequential.
- gctorture-friendly: no SEXP allocations during the par loop (struct flatten happens in `into_data_frame`, not `from_rows_par`).
- A rpkg fixture + testthat case mirroring the existing `dataframe-rayon` test pattern.

Touches `miniextendr-macros/src/dataframe_derive.rs` (struct path); enum path (`dataframe_derive/enum_expansion.rs`) already has a parallel implementation for the equivalent case, worth aligning the two paths.

Priority: low — sequential covers the typical use case.

CGMossa:
## Summary

- Relaxes the `from_rows_par` gate in `#[derive(DataFrameRow)]` so structs with nested struct fields (flattened inline or wrapped with `as_list`) get a working parallel constructor alongside the existing sequential `from_rows`.
- Uses a **sequential pre-pass** (`Vec::with_capacity` + `Clone::clone` loop over `&rows`) for struct-typed and `as_list`-on-struct fields before the rayon scatter-write, avoiding the `Default` bound that blocked these types.
- Adds a compile-time `Clone` assertion on inner struct types (the sequential pre-pass borrows `rows` before `into_par_iter()` consumes it).
- When all columns are struct/`as_list` fields (no scalar/expand columns), emits `let _rows = rows;` instead of a rayon loop to avoid a dead-code par_iter.
- Ships 4 rayon-gated fixture functions (`flat_basic_par`, `flat_two_struct_fields_par`, `flat_as_list_par`, `flat_nested_par`) and 3 R testthat cases gated with `skip_if_not(exists("flat_basic_par", mode = "function"))`.

## Test plan

- [ ] `cargo test -p miniextendr-macros --lib` — 307 tests pass
- [ ] `cargo test --manifest-path rpkg/src/rust/Cargo.toml test_from_rows_par_struct_flatten_matches_sequential` — passes (run from within rpkg/src/rust/)
- [ ] `just devtools-test` — full testthat suite passes including `from_rows_par` struct-flatten tests
- [ ] `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean
- [ ] `cargo clippy --workspace --all-targets --locked --features rayon,... -- -D warnings` — clean (full CI feature list)
- [ ] `cargo fmt --all --check` — clean

## Deviations from plan

- `as_list` unit test (`FlatAsListDataFrame::from_rows_par` in Rust unit test) removed: `IntoList::into_list()` creates R SEXPs, which panics outside R FFI context. R-side verification handled by `flat_as_list_par` testthat test instead.
- `flat_nested_par` fixture added (plan mentioned it); no corresponding R test added since column-shape verification is already covered by the basic/two-struct/as_list cases.

Closes #513


CGMossa (commented):
## Review findings

### Critical / blocking

**`par_len_field` condition is wrong for as_list-only structs** (`dataframe_derive.rs` ~L1595)

The `_len: len` field guard checks `flat_cols.is_empty() && auto_expand_cols.is_empty() && !has_tag && struct_cols.is_empty()`. But `as_list`-on-struct fields live in `flat_cols` with `needs_into_list = true`, not in `struct_cols`. So a struct that has ONLY `as_list`-on-struct fields (no scalar, no tag, no struct_cols) will have non-empty `flat_cols` and will skip `_len: len`, yet the companion struct probably needs `_len` if its only field is a pre-pass-populated `as_list` vec. In the test suite `FlatAsList` also has `id: i32` so this case is not exercised. Needs a test or a guard fix: `&& flat_cols.iter().all(|fc| !fc.needs_into_list)` (the inverse of the existing filter).

**Rust unit test only checks struct-flatten, not `as_list` path** (`dataframe_struct_flatten_test.rs`)

`test_from_rows_par_struct_flatten_matches_sequential` compares `df_seq.origin.len()` and individual `FlatPoint` fields, but the comment says "FlatLocated: seq vs par equivalence." It does NOT test ordering for the parallel scatter-write path of the scalar `id` column. The `id` column check is `assert_eq!(df_seq.id, df_par.id)` — this does test ordering end-to-end for scalars, which is good. However `FlatLocatedDataFrame.origin` is `Vec<FlatPoint>` and the test verifies every element via zip — that's solid. No issue here.

There is no Rust unit test at all for the `as_list` or `FlatSegment` par paths. The comment correctly explains why `as_list` needs R FFI, but `FlatSegment` (two struct fields) has no unit test — only an R-side test that compares column shapes/values. Low risk but worth noting.

### Medium

**`_rows` in the all-prepass fallback is semantically odd** (`dataframe_derive.rs` ~L1638)

`let _rows = rows;` is the right suppressor — `rows` was borrowed by `&rows` loops but not moved, so it would trigger `unused_variable`. The `_` prefix suppresses clippy/rustc. This is correct and necessary. The comment is clear. No action needed.

**`flat_nested_par` fixture has no R-level test** (`test-dataframe-struct-flatten.R`)

Three R tests cover basic, two-struct-fields, and as_list. `flat_nested_par` (nested struct-in-struct) has no R test despite being the most novel case — exercises the recursive `FlatInner → FlatNestedDataFrame` expansion with a pre-pass. The plan (file §3) listed it as "plan mentioned it; no corresponding R test added." At minimum add a column-shape check: `expect_identical(colnames(df), c("id", "inner_a", "inner_sub_depth"))`. Missing test = silent regression surface.

**`Clone` assertion is emitted unconditionally regardless of `#[cfg(feature = "rayon")]`** (`dataframe_derive.rs` ~L1840)

The `_assert_inner_is_clone` assertion is emitted in `struct_assertions` which is in the outer `Ok(quote! { ... })` block — always emitted, not wrapped in `#[cfg(feature = "rayon")]`. This is actually correct (Clone is a reasonable bound to require on an inner DataFrameRow struct), but it's a silent, undocumented API contract change for downstream users who upgrade: their inner structs will now fail to compile unless they add `#[derive(Clone)]`. The error message reads:

```
error[E0277]: the trait bound `Inner: Clone` is not satisfied
   --> src/lib.rs:…
    |
    | fn _assert_inner_is_clone<T: Clone>() {}
    | _do_assert::<…>()
```

The call site points into the `const _` assertion block, which is uninformative. The PR should add a `// NOTE: derive(Clone) is required...` comment inside the `const _` block, or wrap the assertion in a better-named const with a `compile_error!` path.

### Minor

**wasm_registry.rs reset to stub** — this is the same side-effect as wave-2 PRs and expected. Not a bug.

**`skip_if_not(exists(...))` verdict: defensive, not hiding a bug.**

The `#[miniextendr]` fixtures are `#[cfg(feature = "rayon")]`-gated on the Rust side. When compiled without `rayon`, the R-visible function simply doesn't exist. `exists("flat_basic_par", mode = "function")` is the correct runtime check for feature-dependent fixtures — it's the same pattern used elsewhere in testthat suites for optional-feature wrappers. The functions ARE registered when rayon is compiled in, so this is not hiding a missing export.

### Plan conformance

The implementation follows the plan almost exactly. One deviation: the plan says "emit `#inner_ty: Clone` as a where clause on `from_rows_par`." The PR instead emits a `const _` assertion at the type level (not on the method). Both catch the missing `Clone` at derive time, but a where clause on `from_rows_par` would give a better error pointing at the call site, not the assertion block. The plan-recommended approach would be cleaner.

CGMossa:
## Review fix-up — three items addressed (commit 9507a1fe)

### 1. `par_len_field` guard (blocker)

**Decision**: the reviewer's concern turns out to be a false positive — traced the codegen end-to-end. For a struct with ONLY `as_list`-on-struct fields:
- `flat_cols` is non-empty (the `as_list` field lives there with `needs_into_list=true`)
- The companion struct correctly has **no** `_len` field (because `flat_cols.is_empty()` is false in `len_field_decl`)
- The constructor correctly omits `_len: len` (same condition in `par_len_field`)
- `into_data_frame` uses `self.origin.len()` — correct

The suggested fix (`&& flat_cols.iter().all(|fc| !fc.needs_into_list)`) is logically redundant when ANDed with `flat_cols.is_empty()`. Applied it anyway as a **documentation guard** with an explanatory comment so the intent is explicit for future readers.

Added `FlatOnlyAsList` struct (only an `as_list` field, no scalars) to `dataframe_struct_flatten_test.rs` with:
- Compile-time shape assertion verifying no `_len` field
- `test_from_rows_par_only_as_list_compiles` Rust unit test (type-level check; runtime requires R FFI which is covered by the existing `flat_as_list_par` R test)

### 2. `flat_nested_par` missing R test (medium)

Added to `test-dataframe-struct-flatten.R`:
```r
test_that("flat_nested_par produces flattened nested columns", {
  skip_if_not(exists("flat_nested_par", mode = "function"), ...)
  df <- flat_nested_par()
  expect_identical(colnames(df), c("id", "inner_a", "inner_sub_depth"))
  expect_equal(nrow(df), 2L)
  expect_equal(df$id, c(1L, 2L))
  expect_equal(df$inner_a, c(10.0, 20.0))
  expect_equal(df$inner_sub_depth, c(100.0, 200.0))
})
```

### 3. `Clone` assertion opaque error (medium)

**Decision**: migrated to where-clause on `from_rows_par`. Moved `Inner: Clone` from the `const _` assertion block into an explicit `where #(#inner_tys: Clone,)*` on the method. Removed `_assert_inner_is_clone`. This is a straightforward change — the error now points at the `from_rows_par` call site, not at the internal macro expansion.

### Verification

- `cargo test -p miniextendr-macros --lib`: 307 pass
- `cargo test` in `rpkg/src/rust/` with rayon: 49 pass (includes new `test_from_rows_par_only_as_list_compiles`)
- `just devtools-test`: 5775 PASS, 0 FAIL, 17 SKIP
- `clippy_default` + `clippy_all`: clean
- `cargo fmt --all`: clean

Issue #122, opened 2026-04-12
Replace hand-written builtin ALTREP impls with macro/derive generation


CGMossa:
`miniextendr-api/src/altrep_impl.rs` has ~900 lines of hand-written concrete `impl Altrep`, `impl AltVec`, `impl AltInteger` etc. for builtin types: `[T; N]`, `&'static [T]`, `Vec<T>`, `Box<[T]>`, `Range<T>`, `Cow<[T]>`.

These were migrated to use `AltrepExtract` in #119 but are still hand-written per-type. They could potentially be:
- Generated by a declarative macro that takes the type + family
- Collapsed into fewer generic impls
- Generated by the same `impl_alt*_from_data!` macros the user types use

This would significantly reduce the LOC in `altrep_impl.rs`.

From #119.

CGMossa:
Still open. `altrep_impl.rs` has grown to ~2221 lines and is still hand-written per builtin type (`Vec<T>`, `Box<[T]>`, `Range<T>`, `Cow<[T]>`). The data traits (`AltrepLen`, `AltIntegerData`, etc.) now live in `altrep_data/builtins.rs` and the lowlevel bridge traits (`impl_altinteger_from_data!` etc.) are still called per type in `altrep_impl.rs`. The family derives could theoretically generate these for generic wrapper types — but the challenge is that builtin types like `Vec<T>` have existing `IntoR` impls (copy path) and separate ALTREP paths, which complicates using the derive. A targeted declarative macro that takes `(type, family, options...)` would cover most of the reduction.

PR #561, opened 2026-05-14
refactor(api): meta-macro for builtin ALTREP impls + registration (#122, closes #125)

### LINKED ISSUE #125: Move builtin ALTREP class registration into the derive system
Builtin types (`Vec<i32>`, `Box<[i32]>`, `Range<i32>`, `Cow<[i32]>`, etc.) have hand-written `RegisterAltrep` impls and a manual `register_builtin_altrep_classes()` function called from `package_init()`.

These are separate from the derive system introduced in #119. The derive generates `RegisterAltrep` + linkme distributed_slice entries for user types, but builtins use their own path.

Consider unifying:
- Use `impl_register_altrep_builtin!` to also emit linkme entries (removing the manual function)
- Or generate the registration with the same `impl_alt*_from_data!` macro that generates the traits

This would eliminate the manual `register_builtin_altrep_classes()` function and its maintenance burden.

From #119.

CGMossa:
## Summary

Collapses two parallel hand-enumerated lists of the same ~22 ALTREP builtin types in `miniextendr-api/src/altrep_impl.rs`:

1. The `impl_alt*_from_data!` calls → now go through `impl_builtin_altrep_family!(type, family, mode, reg_fn, entry_ident)`.
2. The `register_builtin_altrep_classes()` function → **deleted**. The meta-macro now emits a linkme `MX_ALTREP_REGISTRATIONS` entry per call site, matching how user-defined ALTREP types register via `miniextendr-macros/src/altrep.rs`.

`register_arrow_altrep_classes()` is also deleted. Arrow types live in `optionals/arrow_impl.rs` (feature-gated) and can't be reached by the meta-macro arms — their entries are emitted directly there using the same pattern.

`registry.rs` now iterates `altrep_regs()` exactly once for ALL ALTREP types — builtins, Arrow, user-defined. The `assert_altrep_class_uniqueness()` canary stays as a regression guard. The `wrapper_gen` skip guard at `registry.rs:393` is preserved (no second registration path introduced).

This subsumes the scope previously tracked as #125.

## Test plan
- [x] `just check` / `just clippy` / `just test`
- [x] `just rcmdinstall` + `just devtools-test`
- [ ] CI: full R CMD check + cross-package + minirextendr

## Closes
- #122
- #125

### DIFF
--- a/miniextendr-api/src/altrep_impl.rs
+++ b/miniextendr-api/src/altrep_impl.rs
@@ -1357,8 +1357,14 @@ macro_rules! impl_altcomplex_from_data {
 // region: Meta-macros for built-in ALTREP family instantiation
 //
 // `impl_builtin_altrep_family!` is a single declarative meta-macro that maps a
-// (type, family, dataptr-mode) triple to the correct per-family `impl_alt*_from_data!`
-// call.  All built-in types use `serialize` because they implement `AltrepSerialize`.
+// (type, family, dataptr-mode, sym) 4-tuple to the correct per-family
+// `impl_alt*_from_data!` call AND emits a linkme `MX_ALTREP_REGISTRATIONS` entry
+// so the class is registered at `R_init_*` time (no hand-enumerated list needed).
+//
+// The `$sym:ident` parameter is a unique snake_case identifier used as the suffix
+// for the `#[no_mangle]` registration fn and the linkme static.  It must be unique
+// across all call sites — naming convention: `<container>_<elem>` (e.g. `Vec_i32`,
+// `Box_bool`, `Cow_str`, `Range_i32`).
 //
 // ## Families
 //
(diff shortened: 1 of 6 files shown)


CGMossa (commented):
## Code Review — PR 3: `impl_builtin_altrep_family!` meta-macro

### Verdict: **Reshape** — the abstraction is directionally right but the plan's second half was skipped

---

### Net delta concern

The plan promised −100 to −150 lines net (noting "~55 explicit calls" to collapse) and budgeted +30 for the macro. The actual baseline was 22 calls on main (verified by counting), not 55. So the plan's estimate was off, not the implementation. A 12-arm meta-macro + 22 call sites is not obviously better than 22 self-documenting 3-argument lines — the macro adds a dispatch table whose arms do nothing but rename tokens. The +69 lines compared to main (2290 vs 2221) comes almost entirely from the 50-line meta-macro body + expanded comments. The call sites went from 22 lines to 22 lines; the per-call savings is 0 characters (`impl_builtin_altrep_family!(Vec<i32>, integer, dataptr)` vs `impl_altinteger_from_data!(Vec<i32>, dataptr, serialize)` — one is longer).

This is the key judgment: the meta-macro currently enforces two invariants — always `serialize`, and `materializing` maps to `materializing_dataptr`. Both are good invariants. But 50 lines of dispatch table to enforce them on 22 call sites is a marginal trade. The plan explicitly flagged that `register_builtin_altrep_classes()` (lines ~2142–2221, now lines ~1500–1560) should also be subsumed. That block was **not touched**. It is a second 25-call list of the same types written out longhand. If the meta-macro also generated the `get_or_init_class()` call (e.g., via a `with_registration` option or a companion `register_builtin_altrep_family!` invocation), the total delta would be genuinely negative and the duplication between the two lists would be eliminated.

**Request**: extend the macro (or add a companion) to also emit the `get_or_init_class()` call, so `register_builtin_altrep_classes()` can be replaced by iterating the same invocations. The two lists must be kept in sync by hand today — that is exactly the kind of drift risk a macro should eliminate.

---

### Guard preservation — confirmed

Trace verified:

- `impl_builtin_altrep_family!(Vec<bool>, logical, materializing)` → `impl_altlogical_from_data!(Vec<bool>, materializing_dataptr, serialize)` → `__impl_alt_from_data!(Vec<bool>, ..., materializing_dataptr, serialize)` → `__impl_altrep_base_with_serialize!(Vec<bool>)` (no `@guard` override) → `__impl_altrep_base_with_serialize!($ty, RUnwind)` → `const GUARD = AltrepGuard::RUnwind`. Correct.

- `impl_builtin_altrep_family!(Vec<i32>, integer, dataptr)` → same path → `AltrepGuard::RUnwind`. Correct.

Both `__impl_altrep_base!` and `__impl_altrep_base_with_serialize!` default to `RUnwind` (line 157 and ~191 respectively). No guard regressions.

---

### Bool / Range / String routing — correct

- `Vec<bool>`, `Box<[bool]>` → `logical, materializing` → `materializing_dataptr`. Correct.
- `Range<i32>`, `Range<i64>`, `Range<f64>` → `materializing`. Correct.
- `Vec<String>`, `Box<[String]>`, `Vec<Option<String>>` → `string, dataptr`. Correct (string ALTREP has its own `string_dataptr` path but `dataptr` works for these heap-allocated types).
- `Vec<Cow<'static, str>>` / `Vec<Option<Cow<'static, str>>>` → `string, dataptr`. Correct, consistent with main.

No routing errors found.

---

### Suggested follow-up (block-or-nice-to-have)

The open issue is the `register_builtin_altrep_classes()` function, which repeats all 22 types longhand. Option A: add a second arm to `impl_builtin_altrep_family!` that accepts a `with_registration` flag and additionally emits a `get_or_init_class()` call inside a `pub(crate) fn register_…()` fragment (then collected by a second macro). Option B: introduce a `builtin_altrep_types!` declarative list macro that is iterated both by the impl block and the registration function, removing the duplicate enumeration entirely. Either approach would make the +65 lines earn their keep and future-proof the list against sync errors.

I'd ask for this before merge. It's not a large change — the registration function is simple and already lives immediately below the impl block.

---

### Nits

- The macro comment table says `` | `string` | `impl_altstring_from_data!` `` but the string family uses a `string_dataptr` path internally (not bare `dataptr`) for `impl_altstring_from_data!`'s own arms. The comment is not wrong (that's the macro it calls) but it could confuse readers who look up the underlying macro and see `string_dataptr` as a named option. No action required, just worth a note.
- The `.stderr` snapshot update for `derive_dataframe_struct_field_multisegment_no_derive.rs` is unrelated to this PR. Looks like a toolchain-version drift. Fine to carry along but worth calling out explicitly in the PR description so reviewers don't need to reverse-engineer why it changed.

Issue #568, opened 2026-05-15
fix(connections): custom connection tests fail when connections feature is enabled


CGMossa:
## Summary

The existing custom-connection tests in `rpkg/tests/testthat/test-connections.R`
were guarded with `skip_if_missing_feature("connections")` but `connections`
was never in `detect-features.R`, so they had **never actually run**. PR #569
added `connections` to `tools/detect-features.R` (line 46) to enable the new
`RStdin/RStdout/RStderr/RNullConnection` fixtures, and that surfaced 12
pre-existing test failures in the custom-connection tests, all currently
guarded by `skip_custom_conn_bug()` inside the file (see
`rpkg/tests/testthat/test-connections.R:12-14`).

This plan fixes the underlying bugs in `RCustomConnection` /
`RConnectionIo` so we can drop the `skip_custom_conn_bug()` shim and have
all 12 tests pass.

## Goal

Make these 12 currently-skipped tests pass with 0 FAIL in
`test-connections.R`, and keep the 17 newly-shipping tests from #569 / #572
green:

- `string_input_connection reads multi-line content`
- `string_input_connection reads single line`
- `string_input_connection reads empty string`
- `counter_connection generates sequential integers`
- `counter_connection with single value`
- `memory_connection write then read roundtrip`
- `memory_connection write multiple lines`
- `uppercase_connection transforms to uppercase`
- `uppercase_connection preserves non-alpha characters`
- `rot13_connection encodes text`
- `rot13 double-application returns original`
- `empty_cursor_connection binary roundtrip`
- `cursor_connection reads pre-filled data`

## Root cause

There are **two independent bugs in `RCustomConnection::build` /
`RConnectionIo::build_*`**, plus a likely **third defensive-fix item** in
the panic-trampoline (panic-masking).

### Root cause 1 — CLOSED-state lifecycle (primary, covers most failures)

`R_new_custom_connection` (R's stable C entry point, requires R >= 4.3.0)
returns a connection with `isopen = FALSE`. From `R_ext/Connections.h` and
R's own `src/main/connections.c` initialization path, defaults are:

```
new->text     = TRUE      <- root cause 2 lives here
new->isopen   = FALSE
new->canread  = TRUE
new->canwrite = TRUE
new->canseek  = FALSE
new->blocking = TRUE
```

The builder at `miniextendr-api/src/connection.rs:924-1018` (`build`)
**does not call the connection's `open` callback** before returning the
[...]

Issue #178, opened 2026-04-16
feat(progress): route indicatif output through R connections instead of ptr_R_WriteConsoleEx


CGMossa:
## Summary

Refactor the `indicatif` integration in `miniextendr-api/src/progress.rs` to route output through R's connection system rather than directly calling `ptr_R_WriteConsoleEx`. This would unify progress output with the connection infrastructure (#175, #176, #177) and enable progress bars to target any R connection.

## Motivation

Currently, `progress.rs` implements `TermLike` by calling `ptr_R_WriteConsoleEx` directly (or falling back to `Rprintf`/`REprintf`). This works but bypasses R's connection layer entirely:

```rust
// Current: direct console write
fn write_console(&self, bytes: &[u8]) -> io::Result<()> {
    unsafe {
        if let Some(write) = crate::ffi::ptr_R_WriteConsoleEx {
            write(ptr, chunk as c_int, self.stream.otype());
        }
    }
}
```

Problems with the current approach:
1. **No connection routing** — output can only go to stdout/stderr console, not to file connections, text connections, or custom connections
2. **Bypasses R's I/O infrastructure** — `sink()`, `capture.output()`, and connection-based redirection don't capture indicatif output
3. **Duplicates stream selection** — `RStream::Stdout`/`RStream::Stderr` enum duplicates what `stdout()`/`stderr()` connections already represent
4. **No-op off main thread** — silently drops output instead of routing through a thread-safe connection

If indicatif wrote to R connections instead, all of these problems go away: `sink()` would capture progress bars, users could redirect progress to any connection, and the `RStream` enum could be replaced with actual connection targets.

### Replace `RStream` with connection targets

```rust
// Before:
pub struct RTerm {
    stream: RStream,  // Stdout or Stderr
    width: u16,
}

// After:
pub struct RTerm {
    conn: SEXP,  // R connection SEXP (stdout(), stderr(), or custom)
    width: u16,
}
```

### `TermLike` writes through `R_WriteConnection`

```rust
impl TermLike for RTerm {
    fn write_str(&self, s: &str) -> io::Result<()> {
        let handle = unsafe { get_connection(self.conn) };
        unsafe { write_connection(handle, s.as_bytes()); }
        Ok(())
    }

    fn flush(&self) -> io::Result<()> {
        // Flush the connection
        unsafe { R_FlushConsole(); }
        Ok(())
    }
}
```

### Updated convenience constructors

```rust
// Before:
[...]

Issue #177, opened 2026-04-16
feat(connections): wrap utils::txtProgressBar as a connection-backed progress interface


CGMossa:
## Summary

Provide a Rust-side wrapper for R's `utils::txtProgressBar` that leverages the connection system for output. The `file` parameter of `txtProgressBar` accepts any open connection (or `""` for stdout), making this a natural consumer of the standard connection types from #175.

## Motivation

R's `txtProgressBar` is the standard progress reporting mechanism. Its `file` parameter accepts a connection object — meaning progress output can be directed to stdout, stderr, a file, or any custom connection:

```r
# Default: writes to stdout
pb <- txtProgressBar(min = 0, max = 100, style = 3)

# Write to stderr instead
pb <- txtProgressBar(min = 0, max = 100, style = 3, file = stderr())

# Write to a custom connection
con <- my_rust_connection()
pb <- txtProgressBar(min = 0, max = 100, style = 3, file = con)
```

miniextendr already has `indicatif` integration (`miniextendr-api/src/progress.rs`) for Rust-native progress bars routed through R's console. But there's no way for Rust code to create or drive R's own `txtProgressBar`, which is what R users expect to see and which respects R's connection routing.

## `txtProgressBar` API surface

```r
txtProgressBar(min = 0, max = 1, initial = 0, char = "=", width = NA,
               title, label, style = 1, file = "")
```

Key parameters:
- `min`, `max` — numeric range
- `initial` — starting value
- `char` — progress character (default `"="`)
- `style` — rendering style (1, 2, or 3; style 3 = `|====      | 45%`)
- `file` — **connection or `""`**: where to write output. Must be `""` or an open connection.

The return value is a list (class `"txtProgressBar"`) with closures:
- `$getVal()` — current value
- `$up(value)` — update to new value (renders the bar)
- `$kill()` — finalize (prints newline)

Related R functions:
- `setTxtProgressBar(pb, value)` — calls `pb$up(value)`
- `getTxtProgressBar(pb)` — calls `pb$getVal()`
- `close(pb)` — calls `pb$kill()`

### Rust type

```rust
/// A handle to R's `utils::txtProgressBar`.
///
/// Wraps the R-side list object and provides methods for
/// updating progress from Rust code.
pub struct RTxtProgressBar {
    // Holds the SEXP (protected) for the txtProgressBar list
}

impl RTxtProgressBar {
    /// Create a new progress bar.
    pub fn new(min: f64, max: f64) -> Self { ... }

[...]

PR #569, opened 2026-05-15
feat(connections): RStdin/RStdout/RStderr/RNullConnection foundation

### LINKED ISSUE #175: feat(connections): support stdin(), stdout(), stderr() as Rust connection types
## Summary

Add Rust types representing R's standard terminal connections — `stdin()`, `stdout()`, `stderr()` — so that `#[miniextendr]` functions can accept and return them, and R wrappers can produce the appropriate R connection objects.

## Motivation

R's `stdin()`, `stdout()`, and `stderr()` return pre-opened connection objects that are fundamental to I/O. Currently, miniextendr's connection module (`miniextendr-api/src/connection.rs`) only supports *creating* custom connections via `RCustomConnection::build()`. There's no way for Rust code to:

1. Accept a standard connection from R (no `TryFromSexp` for connection types)
2. Return a reference to stdin/stdout/stderr back to R
3. Wrap these standard connections in Rust-side I/O adapters

This is needed for any Rust code that wants to interact with R's standard I/O streams.

## Connection properties

From `summary(stdin())` / `summary(stdout())` / `summary(stderr())` in R:

```
stdin():
  description: "stdin"
  class:       "terminal"
  mode:        "r"
  text:        "text"
  opened:      "opened"
  can read:    "yes"
  can write:   "no"

stdout():
  description: "stdout"
  class:       "terminal"
  mode:        "w"
  text:        "text"
  opened:      "opened"
  can read:    "no"
  can write:   "yes"

stderr():
  description: "stderr"
  class:       "terminal"
  mode:        "w"
  text:        "text"
  opened:      "opened"
  can read:    "no"
  can write:   "yes"
```

Key observations:
- All three are class `"terminal"`, always open, always text mode
- `stdin()` is read-only, `stdout()` and `stderr()` are write-only
- None support seeking
- These are connection indices 0, 1, 2 in R's internal connection table

### Rust types

```rust
/// Represents one of R's standard terminal connections.
pub enum StdConnection {
    Stdin,
    Stdout,
    Stderr,
}
```

Or three distinct types if we want the type system to enforce read-only vs write-only:

```rust
pub struct RStdin;   // can_read=true,  can_write=false
pub struct RStdout;  // can_read=false, can_write=true
pub struct RStderr;  // can_read=false, can_write=true
```

### R wrapper generation

When a `#[miniextendr]` function returns `RStdin` / `RStdout` / `RStderr`, the generated R wrapper should produce the R connection object:

```r
# Generated R wrapper for a function returning RStdout:
my_output_connection <- function() {
    stdout()
}
```

When a `#[miniextendr]` function *accepts* one of these types, the R wrapper should pass the appropriate connection:

```r
# Generated R wrapper for a function accepting a connection:
write_to <- function(conn) {
    .Call("wrap__write_to", conn)
}
# User calls: write_to(stdout())
```

### Trait implementations needed

1. **`TryFromSexp`** — accept a connection SEXP from R, validate it's the expected standard stream (check description/class on the `Rconn` struct)
2. **`IntoR`** — return the connection SEXP by evaluating `stdin()`/`stdout()`/`stderr()` in R, or by constructing the integer connection index directly
3. Read/write via existing `read_connection()` / `write_connection()` helpers
4. Optionally implement `std::io::Read` for `RStdin` and `std::io::Write` for `RStdout`/`RStderr` (delegates to `R_ReadConnection` / `R_WriteConnection`)

### Test coverage (`miniextendr-api/tests/connection.rs`)

The existing test file should be extended with:

```rust
#[test]
fn stdin_connection_properties() {
    // Verify RStdin has correct capabilities:
    // can_read=true, can_write=false, is_text=true, is_open=true
}

#[test]
fn stdout_connection_properties() {
    // Verify RStdout: can_read=false, can_write=true, is_text=true, is_open=true
}

#[test]
fn stderr_connection_properties() {
    // Verify RStderr: can_read=false, can_write=true, is_text=true, is_open=true
}

#[test]
fn write_to_stdout_connection() {
    // Write bytes through RStdout via write_connection()
}

#[test]
fn write_to_stderr_connection() {
    // Write bytes through RStderr via write_connection()
}
```

These tests will require embedded R (similar to the existing `#[ignore]` test), so they may need the integration harness or `#[ignore]` annotations until that exists.

### rpkg test fixtures

Add R-side tests in `rpkg/tests/testthat/test-connections.R`:

```r
test_that("stdin connection is read-only terminal", {
    con <- stdin()
    expect_true(inherits(summary(con), "summary.connection"))
    s <- summary(con)
    expect_equal(s$description, "stdin")
    expect_equal(s$class, "terminal")
    expect_equal(s$`can read`, "yes")
    expect_equal(s$`can write`, "no")
})

test_that("stdout connection is write-only terminal", {
    con <- stdout()
    s <- summary(con)
    expect_equal(s$description, "stdout")
    expect_equal(s$`can write`, "yes")
    expect_equal(s$`can read`, "no")
})

test_that("stderr connection is write-only terminal", {
    con <- stderr()
    s <- summary(con)
    expect_equal(s$description, "stderr")
    expect_equal(s$`can write`, "yes")
    expect_equal(s$`can read`, "no")
})
```

## Open questions

- Should these be three separate types (stronger type safety) or one enum (simpler API)?
- Should `std::io::Read`/`std::io::Write` impls be provided, or should users go through the connection helpers?
- How to handle the `IntoR` path — evaluate `base::stdin()` via `Rf_eval`, or construct the SEXP from the known connection index (0/1/2)?

## Files to modify

- `miniextendr-api/src/connection.rs` — new types + `TryFromSexp`/`IntoR` impls
- `miniextendr-api/tests/connection.rs` — unit/integration tests
- `rpkg/src/rust/connection_tests.rs` — test fixtures using the new types
- `rpkg/tests/testthat/test-connections.R` — R-side tests
- Possibly `miniextendr-macros/` if R wrapper generation needs special cases for these types

### LINKED ISSUE #176: feat(connections): support nullfile() as a connection target
## Summary

Add support for `nullfile()` — R's platform-independent null device — as a connection target in miniextendr, so Rust code can open and write to `/dev/null` (Unix) or `NUL` (Windows) through R's connection system.

## Motivation

R's `nullfile()` returns a platform-appropriate path to the null device:

```r
> nullfile()
[1] "/dev/null"    # Unix/macOS
[1] "NUL"          # Windows
```

Unlike `stdin()`/`stdout()`/`stderr()`, `nullfile()` does **not** return a connection object — it returns a character string path. To use it as a connection, you wrap it: `con <- file(nullfile(), open = "w")`. This is commonly used for suppressing output, discarding data, or benchmarking write throughput without disk I/O.

miniextendr should provide a Rust type that represents a null-device connection and generates the appropriate R wrapper code.

### Rust type

```rust
/// A connection to the platform's null device (`/dev/null` or `NUL`).
///
/// When returned to R, generates `file(nullfile(), open = "w")`.
/// When accepted from R, validates the connection points to the null device.
pub struct RNullConnection;
```

Or a builder-style approach that allows mode configuration:

```rust
pub struct RNullConnection {
    mode: &'static str,  // "w" (default), "wb", "r", etc.
}

impl RNullConnection {
    pub fn new() -> Self { Self { mode: "w" } }
    pub fn mode(mut self, mode: &'static str) -> Self { self.mode = mode; self }
}
```

### R wrapper generation

When a `#[miniextendr]` function returns `RNullConnection`, the R wrapper should produce:

```r
# For a function returning RNullConnection:
get_null_sink <- function() {
    con <- file(nullfile(), open = "w")
    con
}
```

When accepting from R, the user passes any connection — the Rust side validates it's a null device via the description field on `Rconn`.

### Trait implementations

1. **`IntoR`** — evaluate `file(nullfile(), open = "w")` in R to create the connection SEXP
2. **`TryFromSexp`** — accept a connection SEXP, optionally validate it's a null device connection (check `description` contains `"/dev/null"` or `"NUL"`)
3. Implement `std::io::Write` (delegates to `R_WriteConnection`) — writes are discarded
4. Optionally implement `RConnectionImpl` so it can also be used with `RCustomConnection` builder

### Test coverage (`miniextendr-api/tests/connection.rs`)

```rust
#[test]
fn nullfile_path_is_platform_correct() {
    // On Unix: "/dev/null", on Windows: "NUL"
    #[cfg(unix)]
    assert_eq!(null_device_path(), "/dev/null");
    #[cfg(windows)]
    assert_eq!(null_device_path(), "NUL");
}

#[test]
#[ignore = "requires embedded R runtime"]
fn null_connection_is_writable() {
    // Open file(nullfile(), "w"), write bytes, verify no error
}

#[test]
#[ignore = "requires embedded R runtime"]
fn null_connection_discards_data() {
    // Write data, verify nothing is readable back
}
```

### rpkg test fixtures

Add Rust fixtures and R tests:

```rust
// rpkg/src/rust/connection_tests.rs
#[miniextendr]
pub fn null_write_connection() -> RNullConnection {
    RNullConnection::new()
}
```

```r
# rpkg/tests/testthat/test-connections.R
test_that("null connection accepts writes silently", {
    con <- null_write_connection()
    on.exit(close(con))
    writeLines("this goes nowhere", con)
    # No error = success
})

test_that("nullfile returns platform path", {
    path <- nullfile()
    if (.Platform$OS.type == "unix") {
        expect_equal(path, "/dev/null")
    } else {
        expect_equal(path, "NUL")
    }
})
```

## Relationship to #175

This is complementary to #175 (stdin/stdout/stderr). Together they cover R's four standard I/O targets. The implementation can share infrastructure — e.g., a common `StdConnection` enum or shared `TryFromSexp` validation logic for connections.

## Open questions

- Should `RNullConnection` be a zero-cost wrapper that just produces `file(nullfile())` in R, or should it also work as a Rust-side `std::io::Write` sink (like `std::io::sink()`)?
- Should we provide a `null_device_path()` helper that returns the platform string without going through R?
- How does this interact with `on.exit(close(con))` — should the R wrapper auto-close, or leave that to the user?

## Files to modify

- `miniextendr-api/src/connection.rs` — new type + impls
- `miniextendr-api/tests/connection.rs` — tests
- `rpkg/src/rust/connection_tests.rs` — test fixtures
- `rpkg/tests/testthat/test-connections.R` — R-side tests
- Possibly `miniextendr-macros/` for R wrapper generation of the `file(nullfile())` pattern

CGMossa:
## Summary

Implements issues #175 (std stream connections) and #176 (null connection) — the first PR in the connections sprint.

Sprint shape:
- **PR A (this PR)**: #175 + #176 — `RStdin`/`RStdout`/`RStderr` + `RNullConnection`. Foundation.
- **PR B**: #177 — wrap `utils::txtProgressBar`. Depends on PR A's types.
- **PR C**: #178 — refactor `progress.rs` to write through R connections. Depends on PR A.

## New API (`miniextendr-api`)

```rust
// Zero-sized std-stream types
pub struct RStdin;   // impl std::io::Read
pub struct RStdout;  // impl std::io::Write
pub struct RStderr;  // impl std::io::Write

// RAII null connection
pub struct RNullConnection { /* owned SEXP + open flag */ }
impl std::io::Write for RNullConnection { ... }
impl Drop for RNullConnection { ... }  // closes + releases SEXP
```

All four types have `TryFromSexp` and `IntoR` impls.

### Terminal connections don't support `R_WriteConnection`

R's `stdin()`/`stdout()`/`stderr()` connections have their `write` slot wired differently — they route through `R_WriteConsoleEx` / console hooks, not `R_WriteConnection`. Calling `R_WriteConnection` on a terminal returns an `'write' not enabled` error.

Fix: `RStdout` and `RStderr` use an internal `write_console` helper that builds a null-terminated buffer and calls `Rprintf_unchecked(c"%s", ptr)` / `REprintf_unchecked(c"%s", ptr)`. This correctly passes through `sink()` and `capture.output()` redirection — confirmed by the sink tests.

### Pre-existing custom connection bugs (#568)

When `connections` was added to `detect-features.R` to enable the new fixtures, the 12 pre-existing custom connection tests (which were always skipped because the feature was never compiled in) started failing due to pre-existing lifecycle bugs in `RCustomConnection`. Issue #568 tracks the fix. All 12 tests are temporarily guarded with `skip_custom_conn_bug()`.

## Test results

```
[ FAIL 0 | WARN 0 | SKIP 29 | PASS 5869 ]
```

All new tests pass. Custom connection tests skip cleanly (pending #568 fix).

## Clippy

`cargo clippy --workspace --all-targets --locked -- -D warnings` passes.
CI's `clippy_all` feature set also passes (verified locally before push).

Closes #175
Closes #176


CGMossa (commented):
<details>
<summary><h4>AI-written review details</h4></summary>

## Diff verification

11 files, 1229 additions, 1 deletion, head commit `30022eb1`. Rescue confirmed clean.

---

## [blocker] `write_console` uses `*_unchecked` FFI without a thread guard (MXL301)

`connection.rs:1178–1205` — `write_console(buf, otype)` calls `Rprintf_unchecked` / `REprintf_unchecked` with no `with_r_thread` or `is_r_main_thread()` guard. `SEXP` is `unsafe impl Send` in `ffi.rs:152`, so `RNullConnection` (which holds a `SEXP`) is implicitly `Send`, and both `RStdout` and `RStderr` are `Copy + Send`. A caller can therefore call `io::Write::write` from a worker thread, landing directly in `Rprintf_unchecked` off the main thread — undefined behavior.

Compare `progress.rs:103–105`, which explicitly checks `is_r_main_thread()` and silently no-ops from background threads, and uses the checked `Rprintf`/`REprintf` variants (not `*_unchecked`) as the fallback. The new `write_console` skips both protections.

Fix options (pick one, per the existing pattern):
- Mirror `progress.rs`: add `if !crate::worker::is_r_main_thread() { return 0; }` at the top of `write_console`, and switch to the checked variants (`Rprintf` / `REprintf`) — or wrap in `with_r_thread`.
- Add `!Send + !Sync` bounds to `RStdout`, `RStderr`, `RNullConnection` so the compiler rejects cross-thread use. This is the stronger guarantee but forecloses valid use inside `run_on_worker` closures that hold these by reference.

The `progress.rs` silent-noop approach is probably right here: callers writing to a progress bar from a background thread expect a best-effort no-op, not an abort.

---

## [blocker] `RNullConnection::drop` panics without `drop_catching_panic` / `AbortIfUnwinding`

`connection.rs:1399–1406`:

```rust
impl Drop for RNullConnection {
    fn drop(&mut self) {
        if self.open {
            self.open = false;
            unsafe { self.close_inner() };
        }
    }
}
```

`close_inner` evaluates `close(con)` via `R_tryEvalSilent`. If R's `close()` errors (already-closed connection, finalizer during GC), `close_inner` sets `err != 0` — but the current code ignores that error silently (`R_tryEvalSilent` is used, so no panic there). However, if R raises a condition that propagates as a Rust panic (via `R_tryEvalSilent` converting to a panic in other call sites), or if future callers of `close_inner` add a `panic!`, that panic will unwind through `Drop` into R's finalization machinery.

The project-wide pattern for Drop impls that call into R is `drop_catching_panic` (from `externalptr.rs:1463`), backed by `AbortIfUnwinding` — specifically because `catch_unwind` allocates on first call and triggers "recursive gc invocation" inside R's GC finalizer pass (see `reviews/finalizer-catch-unwind-gc-reentry.md`). The `RNullConnection::drop` impl should use:

```rust
impl Drop for RNullConnection {
    fn drop(&mut self) {
        if self.open {
            self.open = false;
            crate::externalptr::drop_catching_panic(|| {
                unsafe { self.close_inner() };
            });
        }
    }
}
```

This is a latent correctness issue today (since `close_inner` itself uses `R_tryEvalSilent` and doesn't panic), but it violates the invariant all other R-calling Drop impls in this codebase hold, and it will become live the moment `close_inner` grows any error path.

---

## [suggestion] `into_r.rs` comment for `RNullConnection::into_sexp` is self-contradictory

`into_r.rs:2584–2603`: the comment block contains a mid-thought change ("Use close() to flip the flag without closing the connection. Actually: we want to disarm Drop but keep the connection open.") and a struck-through `std::mem::forget` sentence ("Simplest: std::mem::forget(self) would work too, but that leaks the precious-list entry"). The final implementation calls `R_ReleaseObject` + `mem::forget`, which is correct, but the comment scaffolding was left in. Replace with a concise explanation:

```rust
// Transfer ownership to R: release from precious list (R's connection table
// keeps the connection alive), then forget self to skip Drop.
unsafe { crate::ffi::R_ReleaseObject(sexp) };
std::mem::forget(self);
sexp
```

---

## [suggestion] `TryFromSexp for RNullConnection` accepts any open write-capable connection, including terminals

`from_r.rs:1403–1425`: the validation checks `isopen` and `canwrite` but not the connection description or class. This means `RNullConnection::try_from_sexp(stdout())` succeeds — a terminal connection satisfies both predicates. The PR body says "permissive — accept any open, write-capable connection" as a design decision, which is fine if intentional, but the type name `RNullConnection` implies the null device specifically. If the intent is genuinely "any writable connection", the type should probably be renamed `RWriteConnection` or the doc should clearly state that the type is a thin RAII wrapper around any writable connection, not specifically the null device.

If the intent really is null-device-only, add a description check for `"/dev/null"` / `"NUL"` (or rely on `nullfile()` output) — though that's brittle across platforms, so the permissive approach is defensible. Either way, the docs should be explicit.

---

## [nit] `skip_custom_conn_bug()` guards — count confirmed correct

Exactly 12 pre-existing custom-connection tests are guarded (`skip_custom_conn_bug()` appears 12 times in the diff). The 8 new standard-stream/null-connection tests have no skip guard. The skip message references `#568`. Correct.

---

## [nit] `wasm_registry.rs` entries are pre-existing fixtures

`C_rot13_connection`, `C_cursor_connection`, `C_memory_connection`, `C_counter_connection`, `C_uppercase_connection`, `C_empty_cursor_connection`, `C_string_input_connection` were already present in `connection_tests.rs` — they appear here because adding `connections` to `detect-features.R` unblocks their registration in the wasm table. Only `C_rust_get_stderr`, `C_rust_get_stdout`, `C_rust_write_to_null`, `C_rust_write_to_stderr`, `C_rust_get_null_connection` are new to this PR. Not scope creep.

---

## Sprint coherence (PR B / PR C)

`RStderr` implements `io::Write` and is `Copy`, so `txtProgressBar(file = RStderr.into_sexp())` (PR B, #177) can pass the SEXP to R and write through `write!` from Rust. The `IntoR` impl returns `stderr()` from R, which is what R's `txtProgressBar` expects. The API surface is sufficient.

For PR C (#178, indicatif `TermLike`): `progress.rs::RTerm::write_console` already handles the write path via `ptr_R_WriteConsoleEx` / Rprintf fallback — the indicatif refactor doesn't need to go through `RStdout`/`RStderr` directly. The blocker above (thread guard) should be resolved before PR C to avoid copy-paste of the pattern into indicatif integration code.

---

*Generated by Claude Sonnet 4.6 (multi-agent review pipeline). PR was rescued from a contaminated commit — verified the cleaned diff is 11 files / 30022eb1.*

</details>

Issue #571, opened 2026-05-15
RTxtProgressBar: width=auto from terminal width detection


CGMossa:
## Context

PR B (#177) ships `RTxtProgressBarBuilder::width(u32)` with `None` meaning "R default". This issue tracks adding `width = \"auto\"` semantics using terminal width detection.

## Background

R's `txtProgressBar` accepts `width = NA` (use `getOption("width")`) or a fixed integer. In practice users want the bar to fill the terminal. This could be achieved by:

1. Calling `getOption("width")` and passing the result as an integer, or
2. Using a system terminal-width query (e.g., `libc::ioctl(TIOCGWINSZ)`) for a real-time value.

## Why deferred

Option 1 is trivial from Rust but duplicates what `None` already does. Option 2 requires platform-conditional code. Neither is worth the complexity for an initial release.

## Follow-up to

PR B / #177.

CGMossa:
Closed by PR #630 (commit `bcfe9a51`). Added `RTxtProgressBarBuilder::width_auto()` which calls `Rf_GetOptionWidth()` (R's `getOption("width")`) at builder time and stores the result as a fixed width, matching R's canonical "fill the terminal" semantics.

Issue #570, opened 2026-05-15
RTxtProgressBar: add with_progress scope guard convenience wrapper


CGMossa:
## Context

PR B (#177) shipped `RTxtProgressBar` with the full builder + set/get/close/drop API. This issue tracks the `with_progress` scope-guard convenience wrapper deferred from that PR.

## Proposed API

```rust
/// Run `f(pb)` with a style-3 progress bar from `min` to `max`.
/// The bar is automatically closed when `f` returns (or panics).
///
/// # Examples
///
/// ```ignore
/// with_progress(0.0, 100.0, |pb| {
///     for i in 0..=100 {
///         pb.set(i as f64).ok();
///     }
/// });
/// ```
pub fn with_progress<F: FnOnce(&RTxtProgressBar)>(min: f64, max: f64, f: F) {
    let pb = RTxtProgressBar::builder(min, max).style(3).build();
    f(&pb);
    // pb drops here → auto-close
}
```

## Why deferred

Kept out of PR B to minimise scope and because the builder-based API is already fully usable. The scope guard adds ergonomics but no new capability.

## Follow-up to

- Closes #177 (original issue) — PR B ships the core type; this issue tracks the convenience layer.

Issue #430, opened 2026-05-08
convention: new SEXP-storage features ship with a no-arg gctorture fixture


CGMossa:
## Background

The chore/gctorture-audit branch ran a full `gctorture(TRUE)` sweep across rpkg's exported functions and found 18 use-after-free sites (fixed in 67653192). The sweep only exercised functions callable with no arguments; **~36% of exports were skipped** because they require args the harness can't synthesize generically.

The full `devtools::test()` suite under `gctorture(TRUE)` does cover those exports, but it's slow (5.9 h locally for the full run) and doesn't pinpoint which function tripped the GC. The no-arg sweep is fast and pinpoints. Widening its coverage would mean hand-writing fixture wrappers for hundreds of existing exports — most of that effort would re-cover what the full sweep already exercises, so the marginal value is low.

## Proposal

Adopt the convention going forward: **any new `#[miniextendr]` function or method that stores SEXPs across allocations** (typical: `Vec<SEXP>`, sidecar fields, generic-list buffers, anything feeding `from_raw_pairs` / `from_raw_values`) **ships with a no-arg fixture** in `rpkg/src/rust/gc_stress_fixtures.rs` (or a sibling module) that the gctorture sweep can exercise without argument synthesis.

Pattern: an exported function like `gc_stress_my_feature()` that constructs realistic inputs internally and calls the production code path.

## How to apply

- Reviewer check during PR review: \"does this PR add SEXP storage? if yes, is there a no-arg fixture exercising it?\"
- Document in CLAUDE.md under \"Stress-testing GC discipline\" (the existing section already references docs/GCTORTURE_TESTING.md — just add a sentence about fixtures).
- Don't backfill the existing 36% gap — handle case-by-case as those functions are touched.

## Why not just widen the sweep harness

Argument synthesis for arbitrary `#[miniextendr]` signatures is a much bigger project (would need to generate plausible R values for every Rust input type, including class-system instances). A small per-feature fixture is cheaper and more reliable than a generic synthesizer.

## Related

- 67653192 — gctorture UAF fixes
- 0330af53 — audit journal entry
- docs/GCTORTURE_TESTING.md — harness documentation
- rpkg/src/rust/gc_stress_fixtures.rs — existing fixture module

PR #572, opened 2026-05-15
feat(api): RTxtProgressBar — Rust handle for utils::txtProgressBar (#177)

CGMossa:
<TODO: user's framing paragraph — e.g., "PR B of the connections sprint: wraps R's built-in text progress bar so Rust code can drive it directly.">

<details>

## Summary

- Adds `RTxtProgressBar` to `miniextendr-api/src/txt_progress_bar.rs` (new file, `connections` feature gate, sibling to `connection.rs`).
- Builder API: `RTxtProgressBar::builder(min, max)` returns `RTxtProgressBarBuilder` with chainable options `initial`, `char`, `width`, `style` (1–3, default 3), and `file` (any R connection SEXP).
- Runtime API: `set(f64)`, `get() -> f64`, `close(self)` (all return `Result<_, String>`).
- Auto-close on `Drop` via `drop_catching_panic` + `pb$kill()` (same RAII pattern as `RNullConnection` from PR A / #569).
- `!Send + !Sync` via `PhantomData<*const ()>` — enforces main-thread discipline without runtime cost.
- `TryFromSexp` in `from_r.rs`: validates `Rf_inherits(sexp, "txtProgressBar")`, pins on precious list.
- `IntoR` in `into_r.rs`: transfers ownership to R (releases precious list + `mem::forget`).
- `pub mod txt_progress_bar` wired into `miniextendr-api/src/lib.rs` behind `#[cfg(feature = "connections")]`.
- Three new rpkg fixtures in `connection_tests.rs`: `rust_run_progress`, `rust_run_progress_to_stderr`, `gc_stress_txt_progress_bar`.
- Four new R tests in `test-connections.R`: style-3 output capture, stderr routing via `sink(type="message")`, Drop idempotency, GC-stress.
- Roxygen/Rd regenerated with roxygen2 8.0.0 (matches `Config/roxygen2/version` in DESCRIPTION). Backtick to `\code{}` conversion is a real 8.0.0 formatting change, not drift.

Sprint context: PR A (#569, commit `1e4c364d`) merged first; PR C (#178) still to come.

Related issues filed:
- #570: `with_progress` scope-guard convenience wrapper (deferred from this PR)
- #571: `width = "auto"` terminal-width detection (deferred)

## Test plan

- [x] `devtools::test(filter="connections")`: FAIL 0, WARN 0, SKIP 13 (pre-existing `#568` skips), PASS 19
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings`: clean
- [x] `cargo clippy` with CI's `clippy_all` feature set: clean (fixed `manual_range_patterns` lint: `1 | 2 | 3` → `1..=3`)
- [x] `devtools::document()`: no roxygen2 warnings for new fixtures; Rd churn is version-formatting only

## Notes

- The Rd churn (backtick to `\code{}`) is widespread but not from this PR's code changes. It is what roxygen2 8.0.0 produces and matches both the installed version and `Config/roxygen2/version` in DESCRIPTION.
- The `altrep-serialization-cross-session` test failures in the full suite are pre-existing (the `callr` subprocess cannot find the package in a worktree-isolated R library). Unrelated to this PR.
- `RTxtProgressBarBuilder::file` accepts `impl Into<SEXP>` so callers can pass `RStderr.into_sexp()`, `RNullConnection::new().sexp()`, or any other R connection without wrapping.
- `pb$kill()` (not `close()`) is used in `Drop` because it blanks the progress line silently without raising an R error — safer in a destructor context.

---
_Drafted by Claude (claude-sonnet-4-6). Reviewed by the author._

</details>


CGMossa (commented):
## Code Review — PR #572 (RTxtProgressBar, sprint PR B)

Reviewed against commit `7ac66afd`. All implementer claims verified against the actual diff.

---

### Drop discipline — CLEAN

`drop_catching_panic` (the `AbortIfUnwinding` RAII guard, not bare `catch_unwind`) is used in `Drop`. This is exactly the right tool: no lazy LLVM exception machinery, zero allocation, panic-safe across the C-ABI boundary. See `reviews/finalizer-catch-unwind-gc-reentry.md` for why this matters.

The closure inside `drop_catching_panic` is panic-safe: both `kill_txt_progress_bar_inner` calls are wrapped in `R_tryEvalSilent` with errors ignored (can't propagate from Drop anyway), and `R_ReleaseObject` has no panic path. No `.unwrap()` inside the closure.

`R_PreserveObject` ↔ `R_ReleaseObject` pairing:

- **`build_inner`**: `R_PreserveObject` after `txtProgressBar()` succeeds; `R_ReleaseObject` in `Drop` (guarded by `open`). Clean.
- **`TryFromSexp`**: `R_PreserveObject` in `try_from_sexp`, `R_ReleaseObject` in `Drop`. Clean.
- **`close(self)`**: sets `open = false`, calls `R_ReleaseObject`, then `mem::forget(self)` — Drop is a no-op. No double-release path.
- **`IntoR::into_sexp(self)`**: `R_ReleaseObject` before `mem::forget` — same ordering as `RNullConnection::into_sexp`. Clean.

No double-close path exists. `close()` takes `self` by value (consuming), so no subsequent Drop fires on that value. The `open` flag additionally guards the Drop body for the `TryFromSexp`-sourced path.

---

### `!Send + !Sync` via PhantomData — OK, with a tradeoff note

`PhantomData<*const ()>` is present in the struct definition and makes the type `!Send + !Sync` at compile time. This is strictly stronger than PR A's runtime `is_r_main_thread()` early-exit in `write_console` (the `RStdout`/`RStderr` zero-sized types are `Send + Sync` — they don't hold a SEXP at all, so the comparison isn't quite apples-to-apples).

The `PhantomData` approach is the right call for a SEXP-holding type. One consequence: `RTxtProgressBar` cannot be moved into a `run_on_worker` closure even by reference, which is correct — you shouldn't be touching a txtProgressBar SEXP off-thread. Issue #570 (`with_progress`) will hit this constraint naturally; the closure passed to `with_progress` will also be main-thread-only, which is fine.

---

### `set`/`get` thread safety

Both `set` and `get` use `eval_base` (not `R_tryEvalSilent`). This means R errors from `setTxtProgressBar` / `getTxtProgressBar` will surface as `Err(String)` to the caller rather than being silently swallowed. That is the right default for the non-Drop path — callers can decide whether to `.ok()` them. The Drop path correctly uses `R_tryEvalSilent`.

**[suggestion]** `set` and `get` call `eval_base` (which can print R error messages to stderr on failure). In a tight progress-update loop this is unlikely to matter, but it's worth a `# Errors` note in the docs clarifying that R error text may appear on stderr before `Err` is returned. The behaviour is already documented via the `Result` signature; this is documentation polish only.

Because `RTxtProgressBar: !Send`, `set` and `get` can only be invoked on the main thread — so `_unchecked` FFI variants would be defensible here, but the current checked variants are fine and more conservative.

---

### Builder validation

`style()` uses `assert!(matches!(style, 1..=3), ...)` — panic on invalid style at build time. The PR notes this fixed a `manual_range_patterns` lint by using `1..=3` instead of `1 | 2 | 3`. Both the validation and the fix are correct.

**[suggestion]** `min > max` is not checked. R's `txtProgressBar` handles degenerate ranges (renders 100% immediately) but a Rust-level `assert!(min <= max, ...)` in `RTxtProgressBarBuilder::new` or `build()` would make the contract explicit and catch transposed arguments early. Not required for correctness; flag for ergonomics.

---

### Feature gating

`pub mod txt_progress_bar;` in `lib.rs` is decorated with `#[cfg(feature = "connections")]`. Confirmed. The `from_r.rs` and `into_r.rs` modules are also gated with `#[cfg(feature = "connections")]`. Consistent throughout.

---

### `IntoR` ownership transfer

The `RTxtProgressBar::into_sexp` implementation:
1. Calls `self.sexp()` to get the raw SEXP.
2. `R_ReleaseObject(sexp)` — releases from precious list.
3. `mem::forget(self)` — disarms Drop.

This is verbatim the same pattern as `RNullConnection::into_sexp`. Correct ordering; R's connection table (or the calling R environment) is responsible for the SEXP lifetime after this point.

---

### R-side tests

Four tests in the `## txtProgressBar` block:

1. **Style-3 capture**: `grepl("\\ |.+\\|.+%", out)` — style 3 produces lines like `|====      | 40%`; the regex matches correctly. Note `capture.output()` captures stdout but `txtProgressBar` writes via R's `message` channel on some platforms; the test uses `rust_run_progress` which routes to stdout (default `file` arg). Should be fine.

2. **Stderr via `sink(type="message")`**: captures message-channel output with a `textConnection`. Correctly closes with `sink(NULL, type="message")` and `close(con)` — no resource leak.

3. **Idempotent close**: This test is actually an auto-close smoke test, not a true idempotency test. The docstring says "RTxtProgressBar closes idempotently on drop" but the test just calls `rust_run_progress(3L)` twice — it doesn't exercise the explicit `close()` path followed by Drop. That path is safe (as traced above), but it's not tested.

   **[suggestion]** Add an R-side fixture that calls the explicit `close()` path and then verifies Drop is a no-op. `rust_run_progress` only exercises the Drop path. The explicit close → Drop no-op path isn't exercised by any R test.

4. **GC stress**: `gc_stress_txt_progress_bar()` — no-arg fixture present per convention (#430). Good.

---

### Scope

Files modified:
- `miniextendr-api/src/txt_progress_bar.rs` (new)
- `miniextendr-api/src/lib.rs`, `from_r.rs`, `into_r.rs`
- `rpkg/NAMESPACE`, `rpkg/R/miniextendr-wrappers.R`
- `rpkg/man/rust_run_progress.Rd` etc. (generated)
- Widespread `.Rd` churn — confirmed to be roxygen2 8.0.0 backtick→`\code{}` formatting; matches `Config/roxygen2/version` in DESCRIPTION. Not scope creep.

The `connection_tests.rs` fixture file is at `rpkg/src/rust/connection_tests.rs` (not `src/rust/src/`) — no structural issue, just noting the path for reviewers checking out the diff manually.

---

### Sprint coherence with PR C (#178)

PR C (indicatif refactor) needs to write through R connections. PR B provides:
- `file(impl Into<SEXP>)` builder option — so PR C can route indicatif output through an `RNullConnection` or `RStdout` SEXP directly.
- The `kill_txt_progress_bar_inner` / `close_txt_progress_bar_inner` helpers are private to the module, so PR C cannot reuse them; however PR C likely won't need them (indicatif has its own cleanup).
- The `build_inner` `RCall`+`REnv::package_namespace` pattern is not extracted as a shared helper. If PR C needs similar R-eval infrastructure, there may be minor duplication — but both patterns are simple enough that extraction before PR C lands would be premature optimisation.

**Verdict**: PR B and PR C are orthogonal. PR B provides the connection SEXP infrastructure PR C needs via `IntoR`; no shared abstraction is worth extracting now.

---

### Summary

No blockers. Drop discipline is clean. `drop_catching_panic` is used correctly throughout. `!Send + !Sync` via `PhantomData` is correct and stronger than the PR A runtime-check approach. Two suggestions (noisy `eval_base` doc note; `min > max` guard) and one suggestion (add explicit-close R test).

---
*Generated by Claude Opus 4.7 (multi-agent review pipeline).*

Issue #545, opened 2026-05-14
DataFrameRow: verify collision check fires at all nesting levels (#486 follow-up)


CGMossa:
## Background

PR #542 adds `assert_no_payload_field_collision` for inner-payload-field → discriminant-column collision detection. The tests cover the two-level case (outer enum → inner enum with payload field). Multi-level nesting is not tested.

## Gap

Consider a 3-level nesting pattern:

```rust
#[derive(DataFrameRow)]
#[dataframe(tag = "variant")]
struct Inner {
    variant: String, // collision!
}

#[derive(DataFrameRow)]
enum Middle {
    A { inner: Inner },
}

#[derive(DataFrameRow)]
enum Outer {
    X { middle: Middle },
}
```

It is unclear whether the `assert_no_payload_field_collision` const assertion fires correctly at every nesting level in this pattern. Specifically:

- Does the **outer** derive emit the assertion for the `middle` field's contribution relative to `Middle`'s tag?
- Does the **middle** derive emit the assertion for the `inner` field's contribution relative to `Inner`'s tag?
- Are both assertions reachable when only `Outer` is being derived?

## What to do

Add a UI test fixture that constructs a 3-level (or deeper) nesting where a payload field at the innermost level collides with an intermediate discriminant tag. Verify the compiler error fires at the right span with a useful message.

If the check does not fire (false negative), the `assert_no_payload_field_collision` logic needs to be extended to handle transitive nesting.

## References

- PR #542 (primary collision detection; 2-level tests in `miniextendr-macros/tests/ui/`)
- Issue #486 (original issue)
- Reviewer comment: "Multi-level nesting not tested (minor, follow-up issue)"

PR #580, opened 2026-05-15
fix(macros): B1 collision check honors Inner::TAG (#544)

### LINKED ISSUE #544: DataFrameRow nested enum B1 sibling-collision check hardcodes "variant" instead of using Inner::TAG (#486 follow-up)
## Goal

Extend the B1 sibling-collision check so that when an outer `DataFrameRow` enum has a nested-enum field `kind: Inner` where `Inner` uses `#[dataframe(tag = "kind")]`, a sibling field that resolves to `kind_kind` is detected at compile time (currently it is silently missed).

## Why

The B1 check at `enum_expansion.rs:275` hardcodes `"variant"` as the discriminant suffix, so it only fires for the default tag. If `Inner` uses a custom tag (e.g. `tag = "kind"`), a sibling field colliding with `<base>_kind` is silently accepted, producing duplicate R column names and silent data corruption at runtime.

## Constraints

- Match the `assert_no_payload_field_collision` pattern from PR #542: add a new const fn in `miniextendr-api::markers`, emit a `const _: ()` block from the proc-macro.
- The B1 assertion runs at compile time (`const _: ()`), not at proc-macro parse time. The hardcoded `syn::Error` path at line 275 MUST remain for the default-tag `"variant"` case (it fires earlier with a better span and error message), but the const-assertion complement catches the non-default-tag case.
- Don't break any existing UI test snapshots. Add one new compile-fail test (non-default-tag B1 collision) with a matching `.stderr` snapshot (regenerate via `TRYBUILD=overwrite cargo test -p miniextendr-macros` and review).
- Don't bump versions, don't touch RoxygenNote, don't interpret beyond the literal scope of this issue.

## Files to change (with line refs)

- **`miniextendr-api/src/markers.rs`** (after line 121, within the `// region: const-eval collision helper` block): add `assert_no_sibling_field_collision(sibling_cols: &[&str], base: &str, tag: &str)` — a new `pub const fn` that checks whether any element of `sibling_cols` equals `concat!(base, "_", tag)` by manually constructing the comparison with byte-level logic (no `format!` in const fn). The comparison can be done via a two-part byte check: first `base.len() + 1 + tag.len()` bytes, checking `sibling[..base.len()] == base`, then `sibling[base.len()] == b'_'`, then `sibling[base.len()+1..] == tag`.
- **`miniextendr-macros/src/dataframe_derive/enum_expansion.rs`** (~line 1055, near `payload_collision_assertions`): add a new `sibling_collision_assertions: Vec<TokenStream>` collection. For each `EnumResolvedField::Struct(struct_data)` in any variant's resolved fields (deduplicated by base_name, same as `struct_cols`), collect the flat sibling column names (all `EnumResolvedField::Single` col names from the same variant) and emit:

  ```rust
  const _: () = ::miniextendr_api::markers::assert_no_sibling_field_collision(
      &[/* sibling col names as &str literals */],
      /* base_name as &str literal */,
      <#inner_ty as ::miniextendr_api::markers::DataFramePayloadFields>::TAG,
  );
  ```

  Also emit these in `Ok(quote! { ... })` at line 1796, alongside `#(#payload_collision_assertions)*`.

- **`miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs`** (new): compile-fail UI test for the non-default-tag B1 sibling collision case (see fixture design below).
- **`miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.stderr`** (new): snapshot (generate via `TRYBUILD=overwrite cargo test -p miniextendr-macros`, then review).

## Work items (flat, no phases)

1. Add `assert_no_sibling_field_collision(sibling_cols: &[&str], base: &str, tag: &str)` const fn to `miniextendr-api/src/markers.rs`, inside the existing `// region: const-eval collision helper` block. The function checks: for each `col` in `sibling_cols`, if `col.len() == base.len() + 1 + tag.len()`, then byte-compare `col[..base.len()] == base`, `col[base.len()] == b'_'`, `col[base.len()+1..] == tag` — panic with a clear message if any match. Export it `pub` (same as `assert_no_payload_field_collision`). Add `#[doc(hidden)]` and a doc comment explaining purpose and call site.
2. In `enum_expansion.rs`, after `payload_collision_assertions` (around line 1055), build `sibling_collision_assertions: Vec<TokenStream>`. Iterate `struct_cols` (already deduplicated). For each `EnumStructCol`, collect all `Single` col names from ALL variants (not just one variant — a sibling collision in any one variant is a bug). Use the same `flat_col_names` logic already done in the B1 parse-time block but this time emit a const assertion. The literal sibling names go in as `&[&str]` literals in the `quote!`. `base_name` goes in as a string literal. `TAG` comes from `<#inner_ty as DataFramePayloadFields>::TAG`.
3. Emit `#(#sibling_collision_assertions)*` in the final `Ok(quote! { ... })` block alongside the existing `#(#payload_collision_assertions)*`.
4. Write the new UI test fixture `derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs`. It should define `Inner` with `#[dataframe(tag = "status")]` and an `Outer` variant that has both `kind: Inner` AND a sibling field that resolves to `kind_status`. This is structurally analogous to `derive_dataframe_enum_nested_enum_variant_collision.rs` (which tests the default `"variant"` case caught by the hardcoded parse-time check) but uses a non-default tag so only the new const assertion fires.
5. Regenerate the `.stderr` snapshot via `TRYBUILD=overwrite cargo test -p miniextendr-macros -- ui` and review the diff to confirm the new `E0080: evaluation panicked` error fires at the right location with a useful message.
6. Run `cargo test -p miniextendr-macros` (all tests, including existing UI snapshots — none should change).
7. Run clippy_default: `cargo clippy --workspace --all-targets --locked -- -D warnings`.
8. Run clippy_all: `cargo clippy --workspace --all-targets --locked --features rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,connections,nonapi,default-strict,default-coerce,default-r6,default-worker,jiff -- -D warnings`.
9. Run `just devtools-test` to confirm no R-side impact (generated wrappers should not shift; no `rpkg` changes needed).

## Non-goals

- Refactoring the hardcoded parse-time B1 `syn::Error` path for the default `"variant"` tag — it's correct and gives better span/message; leave it in place.
- Touching the inner-payload-field path (already done in PR #542).
- Handling multi-level nesting (outer → struct → inner enum) — that's #545.
- Adding `assert_no_sibling_field_collision` for the struct-derive path (struct fields in `DataFrameRow` structs don't produce a discriminant column, so B1 doesn't apply there).

## Validation

- `cargo test -p miniextendr-macros` (incl. new UI test, no regressions in existing snapshots).
- `cargo clippy --workspace --all-targets --locked -- -D warnings` (clippy_default).
- `cargo clippy --workspace --all-targets --locked --features rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,connections,nonapi,default-strict,default-coerce,default-r6,default-worker,jiff -- -D warnings` (clippy_all).
- `just devtools-test` (no R-side impact expected; verify generated wrappers don't shift).

## Branch name

`fix/544-b1-collision-inner-tag`

## Expected diff shape

5 files, approximately +80 lines / -0 lines:
- `miniextendr-api/src/markers.rs`: +~30 lines (new `assert_no_sibling_field_collision` const fn + doc comment)
- `miniextendr-macros/src/dataframe_derive/enum_expansion.rs`: +~25 lines (new `sibling_collision_assertions` build loop + quote emission)
- `miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs`: +~25 lines (new fixture)
- `miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.stderr`: +~15 lines (generated snapshot)

No deletions expected. The hardcoded `"variant"` at line 275 stays (it's the correct fast path for the common case).

## Key implementation detail: const byte-comparison for prefix_tag

`format!("{base}_{tag}")` cannot be used in a `const fn`. Instead, the new `assert_no_sibling_field_collision` must do a structural byte comparison:

```rust
pub const fn assert_no_sibling_field_collision(
    sibling_cols: &[&str],
    base: &str,
    tag: &str,
) {
    let expected_len = base.len() + 1 + tag.len();
    let base_bytes = base.as_bytes();
    let tag_bytes = tag.as_bytes();
    let mut i = 0;
    while i < sibling_cols.len() {
        let col = sibling_cols[i].as_bytes();
        if col.len() == expected_len {
            // Check base prefix
            let mut j = 0;
            let mut base_match = true;
            while j < base_bytes.len() {
                if col[j] != base_bytes[j] { base_match = false; break; }
                j += 1;
            }
            // Check separator '_'
            let sep_match = col[base_bytes.len()] == b'_';
            // Check tag suffix
            let mut k = 0;
            let mut tag_match = true;
            let offset = base_bytes.len() + 1;
            while k < tag_bytes.len() {
                if col[offset + k] != tag_bytes[k] { tag_match = false; break; }
                k += 1;
            }
            if base_match && sep_match && tag_match {
                panic!(
                    "DataFrameRow B1 sibling-collision: a sibling field in the outer \
                     enum produces a column name that collides with the discriminant \
                     column emitted by a nested inner enum. Rename the sibling field \
                     or use `#[dataframe(tag = \"...\")]` on the inner enum to choose \
                     a different discriminant column name."
                );
            }
        }
        i += 1;
    }
}
```

If `tag` is `""` (struct type, no tag), the function is a no-op since no discriminant column is emitted for structs — guard with an early return on `tag.is_empty()`.

## Sibling-agent contamination check

Before running `devtools-document` or committing, stash any unstaged working tree changes. The worktree may have state from prior agent runs. Confirm `git status` is clean before `git add`.

CGMossa:
Closes #544.

When an outer `DataFrameRow` enum has a nested-enum field using a non-default `#[dataframe(tag = "...")]`, a sibling field colliding with `<base>_<custom_tag>` was silently accepted, producing duplicate R column names and silent data corruption at runtime.

<details><summary>🤖 AI-written implementation notes (claude-sonnet-4-6)</summary>

## Changes

- **`miniextendr-api/src/markers.rs`** (after `assert_no_payload_field_collision`): Added `assert_no_sibling_field_collision(sibling_cols: &[&str], base: &str, tag: &str)` — a new `pub const fn` that checks whether any element of `sibling_cols` equals `concat!(base, "_", tag)` via byte-level comparison (no `format!` in const fn). Returns immediately if `tag.is_empty()` (structs emit no discriminant column). Exported `pub` with `#[doc(hidden)]` matching the existing helper.

- **`miniextendr-macros/src/dataframe_derive/enum_expansion.rs`** (~line 1066, after `payload_collision_assertions`): Added `all_single_col_names` collection (deduped across all variants) and a `sibling_collision_assertions: Vec<TokenStream>` loop over `struct_cols`. Each entry emits `const _: () = ::miniextendr_api::markers::assert_no_sibling_field_collision(&[...sibling_lits...], base_str, <Inner as DataFramePayloadFields>::TAG)`. Emitted alongside `#(#payload_collision_assertions)*` in the final `Ok(quote! { ... })` block.

- **`miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs`** (new): Compile-fail UI test with `Inner` using `#[dataframe(tag = "status")]` and `Outer` having both `kind: Inner` and a sibling `kind_status: String` — exercises the non-default-tag B1 collision path.

- **`miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.stderr`** (new): Generated snapshot showing `E0080: evaluation panicked` at the `#[derive(DataFrameRow)]` on `Outer`, referencing `assert_no_sibling_field_collision` in `miniextendr-api/src/markers.rs`.

## Approach

The new `assert_no_sibling_field_collision` const fn directly parallels `assert_no_payload_field_collision` from PR #542: both live in `miniextendr-api::markers`, both are `pub const fn` with `#[doc(hidden)]`, and both are called from `const _: ()` blocks emitted by the proc-macro. The const fn uses a three-part byte comparison (base prefix, `_` separator, tag suffix) since `format!` is unavailable in const context.

The parse-time B1 `syn::Error` (hardcoded `"variant"`) coexists with the new const assertion intentionally: the parse-time check fires earlier with a better span/message for the common default-tag case, while the const assertion silently handles the non-default-tag case without needing to inspect inner enum attributes from the outer macro parse phase.

Sibling names are collected across ALL variants (not just the variant that introduced the struct field) — a collision in any single variant is a bug regardless of which variant surfaces it.

## Validation

- cargo test: 308 passed, 66 ignored — no regressions; new UI test snapshot confirmed
- clippy_default: pass (`cargo clippy --workspace --all-targets --locked -- -D warnings`)
- clippy_all: pass (features: `rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,connections,nonapi,default-strict,default-coerce,default-r6,default-worker,jiff`)
- UI snapshot diff reviewed: new `.stderr` shows `E0080: evaluation panicked` at `#[derive(DataFrameRow)]` on `Outer` with "DataFrameRow B1 sibling-collision" message and `assert_no_sibling_field_collision` call stack
- devtools-test: N/A (devtools not installed in this worktree environment; PR is macro-only with no rpkg changes)
</details>

### DIFF
--- a/miniextendr-api/src/markers.rs
+++ b/miniextendr-api/src/markers.rs
@@ -119,6 +119,79 @@ pub const fn assert_no_payload_field_collision(fields: &[&str], discriminant_suf
         i += 1;
     }
 }
+
+/// Compile-time assertion: no element of `sibling_cols` matches `concat!(base, "_", tag)`.
+///
+/// Called from `const _: ()` blocks emitted by the outer `DataFrameRow` derive for every
+/// nested-enum struct field, complementing the parse-time B1 check in `enum_expansion.rs`
+/// (which is hardcoded to the default tag `"variant"`). This assertion catches the
+/// non-default-tag case at compile time.
+///
+/// `sibling_cols` is a slice of all flat column names produced by non-Struct sibling fields
+/// in the same outer enum. `base` is the outer field name (e.g. `"kind"`). `tag` is the
+/// inner enum's `#[dataframe(tag = "...")]` value, retrieved via
+/// `<Inner as DataFramePayloadFields>::TAG`.
+///
+/// If `tag` is empty (structs emit no discriminant column), returns immediately as a no-op.
+///
+/// # Call site (emitted by proc-macro)
+///
+/// ```rust,ignore
+/// const _: () = ::miniextendr_api::markers::assert_no_sibling_field_collision(
+///     &["id", "other_col", /* … */],
+///     "kind",
+///     <Inner as ::miniextendr_api::markers::DataFramePayloadFields>::TAG,
+/// );
+/// ```
+#[doc(hidden)]
+pub const fn assert_no_sibling_field_collision(sibling_cols: &[&str], base: &str, tag: &str) {
+    // Structs don't emit a discriminant column, so tag == "" → no-op.
+    if tag.is_empty() {
+        return;
+    }
+    let expected_len = base.len() + 1 + tag.len();
+    let base_bytes = base.as_bytes();
+    let tag_bytes = tag.as_bytes();
+    let mut i = 0;
+    while i < sibling_cols.len() {
+        let col = sibling_cols[i].as_bytes();
+        if col.len() == expected_len {
+            // Check base prefix byte-by-byte.
+            let mut j = 0;
+            let mut base_match = true;
+            while j < base_bytes.len() {
+                if col[j] != base_bytes[j] {
+                    base_match = false;
+                    break;
+                }
+                j += 1;
+            }
+            // Check separator '_'.
+            let sep_match = col[base_bytes.len()] == b'_';
+            // Check tag suffix byte-by-byte.
+            let mut k = 0;
+            let mut tag_match = true;
+            let offset = base_bytes.len() + 1;
+            while k < tag_bytes.len() {
+                if col[offset + k] != tag_bytes[k] {
+                    tag_match = false;
+                    break;
+                }
+                k += 1;
+            }
+            if base_match && sep_match && tag_match {
+                panic!(
+                    "DataFrameRow B1 sibling-collision: a sibling field in the outer enum \
+                     produces a column name that collides with the discriminant column emitted \
+                     by a nested inner enum. Rename the sibling field or use \
+                     `#[dataframe(tag = \"...\")]` on the inner enum to choose a different \
+                     discriminant column name."
+                );
+            }
+        }
+        i += 1;
+    }
+}
 // endregion
 
 /// Marker trait for types that should be converted to R lists via `IntoR`.
--- a/miniextendr-macros/src/dataframe_derive/enum_expansion.rs
+++ b/miniextendr-macros/src/dataframe_derive/enum_expansion.rs
@@ -1064,6 +1064,53 @@ pub(super) fn derive_enum_dataframe(
             }
         })
         .collect();
+
+    // Sibling collision assertions (#544): one per nested-enum struct field.
+    //
+    // The B1 parse-time check (earlier in this function) hardcodes `"variant"` as the
+    // inner tag when building the discriminant column name to compare against sibling
+    // Single fields. That check covers the common case with better spans/messages.
+    //
+    // This const assertion covers the non-default-tag case: when `Inner` uses
+    // `#[dataframe(tag = "foo")]`, the discriminant column is `<base>_foo`, not
+    // `<base>_variant`. The const assertion uses `<Inner as DataFramePayloadFields>::TAG`
+    // so it resolves to the actual tag at compile time regardless of the value.
+    //
+    // We collect all Single col names across ALL variants (not just the variant that
+    // introduced the struct field) — a collision in any one variant is a bug.
+    let all_single_col_names: Vec<String> = {
+        let mut seen = std::collections::HashSet::new();
+        let mut names = Vec::new();
+        for vi in &variant_infos {
+            for erf in &vi.fields {
+                if let EnumResolvedField::Single(d) = erf {
+                    let col = d.col_name.to_string();
+                    if seen.insert(col.clone()) {
+                        names.push(col);
+                    }
+                }
+            }
+        }
+        names
+    };
+    let sibling_collision_assertions: Vec<TokenStream> = struct_cols
+        .iter()
+        .map(|sc| {
+            let inner_ty = &sc.inner_ty;
+            let base_str = &sc.base_name;
+            let sibling_lits = all_single_col_names
+                .iter()
+                .map(|s| quote! { #s })
+                .collect::<Vec<_>>();
+            quote! {
+                const _: () = ::miniextendr_api::markers::assert_no_sibling_field_collision(
+                    &[#(#sibling_lits),*],
+                    #base_str,
+                    <#inner_ty as ::miniextendr_api::markers::DataFramePayloadFields>::TAG,
+                );
+            }
+        })
+        .collect();
     // endregion
 
     // region: Generate From<Vec<Enum>>
@@ -1794,6 +1841,7 @@ pub(super) fn derive_enum_dataframe(
         #payload_fields_impl
         #(#struct_assertions)*
         #(#payload_collision_assertions)*
+        #(#sibling_collision_assertions)*
         #unit_only_factor_impls
     })
     // endregion
--- a/miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs
+++ b/miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs
@@ -1,0 +1,33 @@
+//! Test: nested enum flatten field whose discriminant column (`kind_status`)
+//! collides with a sibling field named `kind_status` when the inner enum uses
+//! a non-default tag `#[dataframe(tag = "status")]`.
+//!
+//! The parse-time B1 check only fires for the default `"variant"` tag.
+//! This case must be caught by the `assert_no_sibling_field_collision` const assertion.
+
+use miniextendr_macros::DataFrameRow;
+
+#[derive(Clone, Debug, DataFrameRow)]
+#[dataframe(tag = "status")]
+enum Inner {
+    Active,
+    Inactive { code: i32 },
+}
+
+/// Outer enum: `kind: Inner` would produce `kind_status` discriminant column
+/// (because Inner uses `tag = "status"`), but the sibling field `kind_status`
+/// also produces a column with that exact name → B1 collision.
+#[derive(DataFrameRow)]
+#[dataframe(tag = "_type")]
+enum Outer {
+    Wrap {
+        id: i32,
+        kind: Inner,
+        kind_status: String, // collides with `kind`'s discriminant column
+    },
+    Other {
+        id: i32,
+    },
+}
+
+fn main() {}
--- a/miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.stderr
+++ b/miniextendr-macros/tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.stderr
@@ -1,0 +1,49 @@
+warning: unexpected `cfg` condition value: `rayon`
+  --> tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs:10:24
+   |
+10 | #[derive(Clone, Debug, DataFrameRow)]
+   |                        ^^^^^^^^^^^^
+   |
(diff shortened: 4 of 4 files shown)


CGMossa:
PASSES

All checks green locally: `cargo test -p miniextendr-macros` (308 passed), `clippy_default`, `clippy_all`. Review findings below.

## markers.rs: `assert_no_sibling_field_collision`

Correct and matches the design spec exactly.

- `tag.is_empty()` early-return is present and correct (lines 148–151). Structs have no discriminant column, so the no-op is the right semantic.
- Byte-comparison is structurally correct: length pre-filter → base prefix loop → `b'_'` separator → tag suffix loop. No `format!` in const. All three loops correctly set `*_match = false; break` on mismatch.
- `sep_match = col[base_bytes.len()]` access is safe: only reached when `col.len() == expected_len == base.len() + 1 + tag.len()`, so the index is always in bounds (tag is non-empty, so `expected_len > base.len()`).
- The `sep_match` check executes unconditionally when `col.len() == expected_len`, even when `base_match` is false — safe because the index validity holds regardless, and `base_match && sep_match && tag_match` guards the panic correctly.
- Panic message is user-actionable and mentions both remedies (rename field or change tag).
- `#[doc(hidden)]` present, doc comment explains call site. Parallel to `assert_no_payload_field_collision` at line 107.

## enum_expansion.rs: `sibling_collision_assertions`

Correct implementation; a few observations worth noting.

**Deduplication of sibling names (lines 1081–1095):** `seen: HashSet` correctly deduplicates across all variants. A field named `id` in both `Wrap` and `Other` contributes a single `"id"` entry to the slice. No false positives, no wasted assertion calls.

**Emission position (line 1844):** `#(#sibling_collision_assertions)*` appears after `#(#payload_collision_assertions)*` inside the same `Ok(quote! { ... })` block. Both depend on `DataFramePayloadFields` being in scope — the struct impls are emitted earlier in the block. Emission order is fine.

**`base_str = &sc.base_name` in `quote!` (line 1100/1108):** `base_name` is `String`; `&String` is interpolated by `quote!` via `ToTokens for str`, producing a correct `&str` string literal. Matches how `#s` is used for `sibling_lits` (line 1103).

**All variants, not just the introducing variant:** The loop at lines 1084–1092 walks `variant_infos` globally and deduplicates by col name. A sibling collision in any variant is caught. This is the correct behaviour — the existing parse-time check (line 275) also sees a collision regardless of which variant introduces it.

**Parse-time check for `"variant"` tag (line 275) preserved:** Unchanged — it still fires earlier with a better span and error message for the default-tag case. The new const assertion complements rather than replaces it.

## UI test

- Fixture (`derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs`): structural mirror of `derive_dataframe_enum_nested_enum_variant_collision.rs`, using `tag = "status"` and a sibling field `kind_status`. Exercises exactly the non-default-tag path.
- The default-tag parse-time check would NOT fire here (collision name would be `kind_variant` but the field is `kind_status`), confirming the test is only caught by the new const assertion.
- `.stderr` snapshot: `E0080: evaluation panicked` at `#[derive(DataFrameRow)]` on `Outer`. The panic message in the snapshot matches the `panic!` literal in `markers.rs` exactly (including escaped quotes). Generated via `TRYBUILD=overwrite`.

## Minor observations (no action required)

- The plan estimated 5 files / ~80 lines; actual is 4 files / +203 lines. The extra lines come from the verbose doc block on the new const fn (73 lines vs ~30 estimate) and a more detailed `.stderr` snapshot (49 lines vs ~15). All legitimate — no extraneous changes.
- `std::collections::HashSet` is brought in inline at line 1082 instead of a `use` import at the top. Consistent with how the adjacent `payload_collision_assertions` block uses `HashMap` already imported at the top of the file (check line ~35). Not a clippy lint, not a problem.
- Issue #545 (multi-level nesting: outer → struct → inner enum) remains tracked and is correctly out of scope here.

PR #584, opened 2026-05-15
fix(rpkg): unbreak R CMD check — prose contained literal \details{}

CGMossa:
Unbreaks `R CMD check` on `main`. Every open PR has been failing the four R CMD check matrix jobs (CRAN-like, Linux devel/release/oldrel-1) since PR #417 merged on 2026-05-07, and the failure was inherited rather than caused by any of those PRs.

## Root cause

The `multi_para_doc_demo` fixture (added in #417 to close #343) has a doc comment containing literal `` `\details{}` `` text:

```rust
/// This is the details paragraph (paragraph 3). It should appear in
/// a `\details{}` block in the generated Rd file.
```

This crate isn't configured for roxygen2 markdown processing (no `Config/roxygen2: list(markdown = TRUE)` in `DESCRIPTION`), so the backticks aren't transformed to `\code{}`. The literal `\details{` ends up as raw text inside the rendered `.Rd`'s own `\details{}` block. R's Rd parser sees a nested section header and bails:

```
prepare_Rd: man/multi_para_doc_demo.Rd:19: unexpected section header '\details'
```

That invalidates the auto-generated `miniextendr-Ex.R`, which is why the cascading `parse error in file 'miniextendr-Ex.R': 6: unexpected symbol` also shows up.

## Fix

Three-line prose change across the source-of-truth fixture and its already-generated wrapper/Rd:

- `rpkg/src/rust/multi_para_doc_demo.rs` — the canonical doc comment
- `rpkg/R/miniextendr-wrappers.R` — its `#'` mirror in the generated wrappers
- `rpkg/man/multi_para_doc_demo.Rd` — the parsed `\details{}` body

The fixture's purpose (verifying paragraph 3+ becomes `@details`) is preserved; only the gratuitous mention of `\details{}` in the prose is dropped.

Closes #343.

<details><summary>🤖 AI-written (claude-opus-4-7 dispatcher, in-session implementation)</summary>

Surfaced while diagnosing why every PR in an autonomous issue→PR loop had the same CI failure. The dispatcher had also opened a redundant PR (#577) re-implementing #343's already-merged fix because the loop didn't grep `main` for prior fixes — a real loop blindspot now noted. #577 has been closed as duplicate.

</details>

PR #574, opened 2026-05-15
fix(rpkg): unexport call_attribution_demo fixtures (#318)

CGMossa:
Closes #318.

<details>
<summary><h3>🤖 AI-written implementation notes (claude-sonnet-4-6 + claude-opus-4-7)</h3></summary>

## Summary

PR #316 added `rpkg/src/rust/call_attribution_demo.rs` with two demo fixtures (`call_attr_with` and `C_call_attr_without`) for the runnable transcript in `docs/CALL_ATTRIBUTION.md`. They were `@export`-ed and so leaked `man/call_attribution_demo.Rd` into the installed package alongside real API surface.

- `rpkg/src/rust/call_attribution_demo.rs`: added `#[miniextendr(noexport)]` to both attrs and `@noRd` in their Rust doc comments. The generated wrappers now carry `@noRd` instead of `@keywords internal` + `@rdname call_attribution_demo`, suppressing Rd generation entirely.
- `rpkg/R/miniextendr-wrappers.R`: regenerated (the only diff is the two wrappers' roxygen blocks; 6 lines).
- `rpkg/man/call_attribution_demo.Rd`: deleted.
- `rpkg/tests/testthat/test-call-attribution.R` (new): five tests that pin the `conditionCall` and class-hierarchy behaviour documented in `docs/CALL_ATTRIBUTION.md`, plus an assertion that neither fixture appears in `getNamespaceExports("miniextendr")`. Accesses both via `miniextendr:::`.

## Test plan

- [x] `just configure && just rcmdinstall && just devtools-document` — succeeded; `man/call_attribution_demo.Rd` deleted.
- [x] `just devtools-test` — all green (`[ FAIL 0 | WARN 0 | SKIP 29 | PASS 5880 ]`).
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean.
- [x] `cargo clippy --workspace --all-targets --locked --features rayon,rand,...,jiff -- -D warnings` (CI clippy_all) — clean.
- [x] `R -e 'library(miniextendr); stopifnot(!"call_attr_with" %in% getNamespaceExports("miniextendr")); stopifnot(!"unsafe_C_call_attr_without" %in% getNamespaceExports("miniextendr"))'` — passes.

## Notes

[...]

PR #585, opened 2026-05-15
fix(rpkg): unbreak Rd extraction (connection_tests #[cfg] doc-split + roxygen2 markdown regen)

CGMossa:
## Why

PR #584 fixed one of two CI-blocking bugs on `main`. R CMD check is still failing on every open PR (#574, #580, #581, #582, #583) with the same `miniextendr-Ex.R` parse error — different root cause this time.

## Root cause

`rpkg/src/rust/connection_tests.rs` had 7 fixtures with `#[cfg(feature = "connections")]` placed **between** two blocks of doc comments:

```rust
/// @examples
/// \dontrun{ ... }
#[cfg(feature = "connections")]       // ← attribute splits the /// stream
/// Test creating a connection that ROT13-encodes text on read.
/// @param text Character string of input text to encode.
#[miniextendr]
pub fn rot13_connection(text: &str) -> SEXP { ... }
```

The `#[miniextendr]` proc macro concatenates all `///` lines from the function into a single roxygen stream. The trailing prose paragraph therefore lands **inside the still-open `\examples{}` block** of the generated `.Rd`:

```
\examples{
\dontrun{
conn <- rot13_connection("hello")
...
}
Test creating a connection that ROT13-encodes text on read.   ← bare prose, breaks parse
\dontrun{ ... }
}
```

`tools::Rd2ex()` happily extracts this bare prose into `miniextendr-Ex.R` as raw R, which then fails `parse()` with "unexpected symbol". The cascading "parse error in file 'miniextendr-Ex.R'" is what was showing up in CI logs on every open PR.

## Fix

Two commits, separable for review:

**Commit 1 — focused source fix:** the post-`#[cfg]` doc comments were just redundant restatements of the title/`@param` already present in the leading block. Deleted them so each `#[cfg]` now sits cleanly between the closing `///` and `#[miniextendr]`. 7 fixtures touched, regenerated `wrappers.R` + `connection_tests.Rd`.

**Commit 2 — bulk roxygen2 markdown regen:** `DESCRIPTION` has `Config/roxygen2/markdown: TRUE` and pins roxygen2 8.0.0, but the on-disk Rd files were generated without markdown processing. Re-ran `just devtools-document` to sync. Mechanical churn across 218 files (backticks → `\code{}` / `\verb{}`, R6 method anchors renamed `-new` → `-initialize`, `\itemize` indentation). No NAMESPACE diff, no semantic API changes.

As a side benefit, `multi_para_doc_demo.Rd`'s `\details{}` body — the trigger for PR #584's prose fix — now renders the `` `\details{}` `` token safely as `\verb{\\details\{\}}`, providing belt-and-braces protection alongside the prose change already on main.

## Verification

Local `tools::Rd2ex()` + `parse()` sweep across all `rpkg/man/*.Rd` now passes cleanly (zero FAILs).

## Follow-ups filed separately

The macro should detect doc-comment paragraphs landing after `@examples` (or after any other section terminator) and either close the open block or coalesce them into `@description`. Filing as a separate issue — fix here is the source-level cleanup; the macro robustness is a wider design conversation.

<details><summary>🤖 AI-written (claude-opus-4-7 dispatcher, in-session CI rescue)</summary>

Surfaced while diagnosing why PR #584's CI was still red after the multi_para_doc_demo prose fix landed. Two-bug situation — #584 resolved the first; this PR resolves the second.

</details>


CGMossa:
Follow-up: #586 tracks the macro-level fix (attribute between doc-comment blocks silently corrupts roxygen output). This PR is the source-level cleanup; #586 is the wider design conversation.

Issue #499, opened 2026-05-11
audit: residual items from April-11 critical/medium audit


CGMossa:
# Plan: Fix All Audit Issues

## Critical (9) — fix now

1. expression.rs: protect msg_sexp in get_r_error_message()
2. columnar.rs: -(nrow as i32) → checked i32 conversion
3. connection.rs: debug_assert → runtime assert in get_state()
4. ndarray_impl.rs: .product() → checked_mul chain
5. nalgebra_impl.rs: argmin/argmax skip NaN
6. factor.rs: bounds check in level()
7. cargo-revendor strip.rs: fix TOML section prefix matching
8. cargo-revendor vendor.rs: normalize Windows paths in freeze
9. cargo-revendor Cargo.toml: verify edition (2024 = Rust 1.85+, intentional)

## Medium (11) — fix now

10. Arrow buffer guard: document allocation failure behavior
11. Arrow RecordBatch: add column length pre-validation
12. DataFusion runtime: return Result instead of expect
13. toml_impl: protect integer/float array allocations
14. serde_impl: replace .expect() with .map_err() for user data
15. encoding.rs: add is_r_main_thread() assert
16. DataFrame transpose: fix GC protection in error path
17. bytes_impl: check remaining_mut() overflow
18. indexmap_impl: migrate to OwnedProtect
19. GC_PROTECT.md: add ProtectPool section
20. Optionals as-casts: document 15 lossy casts

CGMossa:
## Audit verification (2026-05-16)

Verified all 20 items against current code. Status breakdown:

**Already fixed since April:**
- #1 expression.rs msg_sexp protection
- #2 columnar.rs nrow `i32::try_from`
- #4 ndarray_impl Array4/5 `checked_mul`
- #5 nalgebra_impl argmin/argmax NaN filter
- #6 factor.rs `level()` bounds check
- #8 cargo-revendor `path_to_toml` Windows normalization
- #15 encoding.rs main-thread `debug_assert`
- #16 columnar.rs `from_rows` ProtectScope

**Self-resolving:**
- #9 cargo-revendor `edition = "2024"` (intentional; audit text confirms)

**Implementer queue (8 items, 3 PRs):**

- **PR A** (`miniextendr-api/src/`): #3 connection.rs assert, #13 toml_impl array PROTECT, #17 bytes_impl `remaining_mut` overflow, #18 indexmap_impl migrate to `OwnedProtect`
[...]

PR #589, opened 2026-05-16
fix(api): audit #499 PR A — core PROTECT/assert fixes

CGMossa:
Flight 1 / PR A from #499. Four small correctness fixes in `miniextendr-api/src/`.

<details>
<summary>AI-written breakdown (Claude Sonnet 4.6, drafted via planner → implementer pipeline)</summary>

## Summary

Per the [Flight 1 plan](https://github.com/A2-ai/miniextendr/blob/main/journal/2026-05-16-flight1-499-plan.md), this PR groups the four `still-applies` audit residuals that all live under `miniextendr-api/src/` and touch FFI PROTECT / assert discipline.

### Commits (in order)

1. **`fix(connection): runtime assert null private pointer`** — `connection.rs:636` `debug_assert!` → `assert!`. The null check is elided in release builds today; promoting to runtime makes the catastrophic-UB site safe in all profiles.
2. **`fix(toml): OwnedProtect integer/float/boolean array branches`** — `toml_impl.rs::array_to_sexp` had its String branch wrapped in `OwnedProtect` but the Integer (×2 — INTSXP path + REALSXP fallback), Float, and Boolean branches returned unprotected SEXPs. Brought all four into line with the String branch's pattern. Also adds a no-arg `gc_stress_toml_array()` fixture in `rpkg/src/rust/gc_stress_fixtures.rs` (gated on `#[cfg(feature = "toml")]`) per the #430 convention — exercises all four homogeneous array branches.
3. **`fix(bytes): replace lossy as-cast in TestBufMut::remaining_mut`** — test impl was doing `self.data.borrow().remaining_mut() as i32` for a growable `BytesMut` whose `remaining_mut()` returns `usize::MAX - len`. Replaced with `i32::try_from(...).unwrap_or(i32::MAX)` (matches the trait-docs pattern at `bytes_impl.rs:248–249`). Replaced the apologetic comment in `test_rbufmut_has_remaining` with an assertion that exercises the now-correct behaviour.
4. **`fix(indexmap): migrate IndexMap IntoR to OwnedProtect`** — replaced the manual `Rf_protect(list); Rf_protect(names); … Rf_unprotect(2)` in `IndexMap<String, T>::try_into_sexp` with `OwnedProtect`. `value.into_sexp()` and `checked_mkchar` both allocate inside the fill loop; RAII guards root both vectors for the duration. Drops the unused `Rf_protect`/`Rf_unprotect` imports.

### Validation

- `just configure` — clean (source mode, monorepo override).
- `cargo lcheck -p miniextendr-api --features 'toml bytes indexmap'` — passes.
- `cargo ltest -p miniextendr-api --features 'toml bytes indexmap'` — 5 doctests pass, 221 ignored, 0 failed.
- `just rcmdinstall` — passes from the worktree.
- `just devtools-test` — 5887 PASS, 29 SKIP, 1 FAIL pre-existing on a different branch (`test-arrow.R:157` ragged-RecordBatch test belongs to audit item #11 / PR B's scope; not in `origin/main`).
- `just clippy` (CI's `clippy_default`) — passes.
- `cargo clippy --workspace --all-targets --locked --features '<CI clippy_all curated list>' -- -D warnings` — passes.

### Things explicitly NOT in this PR

Per the planner's scope notes:
- Audit items 1, 2, 4, 5, 6, 8, 9, 15, 16 — already fixed since the audit was written.
- Audit items 10, 11 — PR B (Arrow doc + RecordBatch column-length pre-validation).
- Audit item 19 — PR C (GC_PROTECT.md ProtectPool section).
- Audit items 7, 12, 14, 20 — `needs-decision`, surfaced to the maintainer separately.

Refs #499 (items 3, 13, 17, 18)

</details>

### DIFF
--- a/miniextendr-api/src/connection.rs
+++ b/miniextendr-api/src/connection.rs
@@ -633,7 +633,7 @@ pub trait RConnectionImpl: Sized + 'static {
 #[inline]
 unsafe fn get_state<T: RConnectionImpl>(conn: *mut Rconn) -> &'static mut T {
     let private = unsafe { (*conn).private };
-    debug_assert!(!private.is_null(), "Connection private pointer is null");
+    assert!(!private.is_null(), "Connection private pointer is null");
     unsafe { &mut *private.cast::<T>() }
 }
 
--- a/miniextendr-api/src/optionals/bytes_impl.rs
+++ b/miniextendr-api/src/optionals/bytes_impl.rs
@@ -643,7 +643,7 @@ mod tests {
 
     impl RBufMut for TestBufMut {
         fn remaining_mut(&self) -> i32 {
-            self.data.borrow().remaining_mut() as i32
+            i32::try_from(self.data.borrow().remaining_mut()).unwrap_or(i32::MAX)
         }
 
         fn put_u8(&self, val: i32) {
@@ -867,10 +867,9 @@ mod tests {
     #[test]
     fn test_rbufmut_has_remaining() {
         let buf = TestBufMut::with_capacity(10);
-        // BytesMut is growable, so remaining_mut() returns usize::MAX - len
-        // which when cast to i32 may overflow. But has_remaining_mut() uses > 0.
-        // For our test implementation, remaining_mut as i32 overflows to negative,
-        // so let's just check that we can write to the buffer.
+        // BytesMut is growable: remaining_mut() returns usize::MAX - len, which
+        // saturates to i32::MAX via try_from. has_remaining_mut() returns true.
+        assert!(buf.remaining_mut() > 0);
         buf.put_u8(1);
         assert_eq!(buf.len(), 1);
     }
--- a/miniextendr-api/src/optionals/indexmap_impl.rs
+++ b/miniextendr-api/src/optionals/indexmap_impl.rs
@@ -51,8 +51,9 @@
 
 pub use indexmap::IndexMap;
 
-use crate::ffi::{R_xlen_t, Rf_allocVector, Rf_protect, Rf_unprotect, SEXP, SEXPTYPE, SexpExt};
+use crate::ffi::{R_xlen_t, Rf_allocVector, SEXP, SEXPTYPE, SexpExt};
 use crate::from_r::{SexpError, SexpTypeError, TryFromSexp};
+use crate::gc_protect::OwnedProtect;
 use crate::into_r::IntoR;
 
 // region: TryFromSexp for IndexMap<String, T>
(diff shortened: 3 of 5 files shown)


CGMossa (commented):
This looks clean. The four commits map 1:1 onto audit items 3, 13, 17, 18, the scope doesn't leak, and the PROTECT/OwnedProtect changes are well-formed. The new `gc_stress_toml_array()` fixture exercises exactly the four homogeneous branches it should. I'd recommend approval modulo one wording nit on the toml comment. Want one more pass once CI clears? Otherwise this is good to land.

<details><summary>AI review (Opus 4.7)</summary>

### Scope vs. audit map
Audit items 3 (`connection.rs` runtime assert), 13 (`toml_impl` integer/float/boolean PROTECT), 17 (`bytes_impl` TestBufMut overflow), 18 (`indexmap_impl` migrate to OwnedProtect) match the four commits 1:1. No drive-by changes, no unrelated files touched. Verified against `journal/2026-05-16-flight1-499-plan.md` items in the live-work table.

### Commit `22fb9e59` — connection `assert!` (item 3)
- `miniextendr-api/src/connection.rs:636` — `debug_assert!` → `assert!` is appropriate here. `get_state` is the central dispatch helper called from every connection trampoline, and a null `private` dereference would be UB. All callers route through `catch_connection_panic` → `guarded_ffi_call_with_fallback`, so the assert-panic gets converted to a fallback return rather than unwinding into C. Cost is one predictable branch per callback — negligible.
- No new test, which the plan explicitly says is acceptable (hostile-null fixture not worth it).

### Commit `c963c066` — toml array PROTECT (item 13)
- `miniextendr-api/src/optionals/toml_impl.rs:414–488` — all four homogeneous branches (Integer→INTSXP, Integer→REALSXP fallback, Float, Boolean) now wrap the alloc in `OwnedProtect::new` and return `sexp.get()`. Drop order is correct: at `return sexp.get()` the `OwnedProtect` drops after the SEXP value is read, calling `Rf_unprotect(1)` — matches the existing String-branch pattern at lines 391–412.
- `rpkg/src/rust/gc_stress_fixtures.rs:309–354` — `gc_stress_toml_array()` covers all four branches including the i64 → i32 overflow fallback. No-arg signature is correct per the convention from #430 — the fast gctorture sweep will pick it up.
- **Nit (non-blocking)**: the comment on lines 425–427 says `i32::try_from` "may theoretically allocate on some platforms" — `i32::try_from(i64)` returns `Result<i32, TryFromIntError>` and `TryFromIntError` is a tiny `Copy` ZST-ish marker (`pub struct TryFromIntError(());`). It cannot allocate. The real rationale is the second half of the comment ("consistency with the String/float/boolean branches matters for R-devel GC") — I'd just drop the first clause. Not blocking.

### Commit `3dc42769` — TestBufMut overflow (item 17)
- `miniextendr-api/src/optionals/bytes_impl.rs:644–646` — `i32::try_from(...).unwrap_or(i32::MAX)` matches the docstring pattern at lines 248–249 and the contract that `has_remaining_mut()` callers expect a positive value. Saturating at `i32::MAX` is the right semantic for a growable buffer reporting "effectively unbounded headroom."
- Removed the apologetic comment block and added a proper assertion (`assert!(buf.remaining_mut() > 0)` at line 870) — the new test actually verifies the fix.

### Commit `00e1a6f4` — indexmap OwnedProtect (item 18)
- `miniextendr-api/src/optionals/indexmap_impl.rs:131–149` — manual `Rf_protect` / `Rf_unprotect(2)` replaced with two `OwnedProtect` guards.
- **Drop order verified**: `list` declared first, `names` second. Rust drops in reverse declaration order, so `names` drops first (UNPROTECT(1) pops the top, which is `names`), then `list` drops (UNPROTECT(1) pops the now-top `list`). The PROTECT stack stays LIFO-balanced. ✓
- **Allocation-between-protect-and-store check**: inside the loop, `value.into_sexp()` is the only potentially-allocating call before `set_vector_elt`; both `list` and `names` are already protected, and `set_vector_elt` writes the child reference into a protected parent so no child-protect is needed. `checked_mkchar` may allocate but both guards are still live. ✓
- Import cleanup: `Rf_protect` / `Rf_unprotect` removed from `use crate::ffi::{...}` since they're no longer needed. ✓
- Existing indexmap tests in the same file cover the round-trip.

### CI
At time of review: green pre-checks (Generated Files Check, Version Check, wasm32, Detect changes), in-progress: Rust lint, R CMD check (release/devel/oldrel-1), Cross-Package, minirextendr, CRAN-like, Bootstrap Vendor, Sync Checks. I'd want Linux devel and the bootstrap vendor test to come back green before merging — R-devel is the one that has historically caught PROTECT regressions (cf. PR #344 `af6b4875`).

### Recommendation
**APPROVE** modulo the one comment-wording nit on the integer branch (non-blocking) — feel free to address in this PR or skip. Wait for CI (especially Linux devel + R tests) to come back green.

</details>

PR #590, opened 2026-05-16
fix(api): audit #499 PR B — arrow_impl doc + RecordBatch validation

CGMossa:
Flight 1 PR B from the #499 audit residuals — two small Arrow-side fixes,
one doc-only and one error-message improvement with a testthat fixture.

<details>
<summary>AI-generated description (Sonnet 4.6)</summary>

## Summary

Addresses items 10 and 11 of the #499 audit (Flight 1 PR B):

- **Item 10** (`84b6fa40`): adds a `# Allocation failure` doc section to
  `alloc_r_backed_buffer` explaining that `Rf_allocVector` longjmps on OOM
  and the framework's `R_UnwindProtect` wrapper converts it to a Rust panic
  surfacing as a "memory exhausted" R error — so the two `.expect()` calls
  in the body are unreachable in production.
- **Item 11** (`a5ded5dd`): adds a column-length pre-validation pass to
  `TryFromSexp for RecordBatch` so the error message names the offending
  column (and the reference column it disagrees with) before reaching
  `RecordBatch::try_new`, which would otherwise emit a generic Arrow error.

## Test plan

- New testthat fixture in `rpkg/tests/testthat/test-arrow.R`:
  `"RecordBatch rejects ragged columns with offending name in error"` —
  passes a named list with column `y` shorter than column `x` to
  `arrow_recordbatch_roundtrip` and asserts the error message contains the
  offending column name `'y'`. The ragged list reaches the new validation
  block (columns build successfully one at a time before the cross-column
  length check), so the test exercises the new code path rather than an
  existing error.
- Local validation:
  - `cargo lcheck -p miniextendr-api --features arrow` passes
  - `cargo clippy --workspace --all-targets --locked -- -D warnings` passes
  - `cargo clippy ... --features <ci.yml clippy_all curated list> -- -D warnings` passes
  - Skipping local `just devtools-test` per coordinator (worktree `rv/`
    has no deps installed); CI will run the full test.

## Scope discipline

- No refactor of `TryFromSexp for RecordBatch` beyond the new pre-validation
  block.
- Item 10 is doc-only; item 11 is ~17 lines of new validation logic.

Refs #499 (items 10, 11)

</details>

### DIFF
--- a/miniextendr-api/src/optionals/arrow_impl.rs
+++ b/miniextendr-api/src/optionals/arrow_impl.rs
@@ -362,12 +362,15 @@ unsafe fn sexp_to_arrow_buffer<T: RNativeType>(sexp: SEXP) -> Option<arrow_buffe
 ///
 /// # Allocation failure
 ///
-/// `Rf_allocVector` raises an R error (longjmp) on OOM rather than returning
-/// a null pointer. The framework's `R_UnwindProtect` wrapper catches this
-/// longjmp and converts it to a Rust panic, which surfaces to the caller as a
-/// "memory exhausted" R error. The two `.expect()` calls in the body are
-/// therefore unreachable in production — they guard against logic errors
-/// (wrong `len` encoding or unexpected null pointer), not against OOM.
+/// If `Rf_allocVector` cannot satisfy the request, R longjmps from inside
+/// the allocation. `R_UnwindProtect` catches the longjmp, the framework
+/// recognises the R-origin path via `RErrorMarker`, and `R_ContinueUnwind`
+/// re-raises R's original `Error: vector memory exhausted (limit reached?)`
+/// verbatim. The error message is R's, not Rust's — no tagged-SEXP synthesis
+/// happens on this path (cf. `with_r_unwind_protect_error_in_r`). The two
+/// `.expect()` calls in the body are therefore unreachable in production —
+/// they guard against logic errors (wrong `len` encoding or unexpected null
+/// pointer), not against OOM.
 ///
 /// # Example
 ///
--- a/miniextendr-api/src/optionals/arrow_impl.rs
+++ b/miniextendr-api/src/optionals/arrow_impl.rs
@@ -1035,6 +1035,23 @@ impl TryFromSexp for RecordBatch {
             columns.push(array);
         }
 
+        // Validate that all columns have the same length before calling
+        // RecordBatch::try_new, which would otherwise emit a generic error.
+        if let Some(first) = columns.first() {
+            let nrow = first.len();
+            for (i, col) in columns.iter().enumerate().skip(1) {
+                if col.len() != nrow {
+                    return Err(SexpError::InvalidValue(format!(
+                        "column '{}' has length {} but column '{}' has length {}",
+                        names[i],
+                        col.len(),
+                        names[0],
+                        nrow,
+                    )));
+                }
+            }
+        }
+
         let schema = Arc::new(Schema::new(fields));
         RecordBatch::try_new(schema, columns).map_err(|e| SexpError::InvalidValue(e.to_string()))
     }
--- a/miniextendr-api/src/optionals/arrow_impl.rs
+++ b/miniextendr-api/src/optionals/arrow_impl.rs
@@ -360,6 +360,15 @@ unsafe fn sexp_to_arrow_buffer<T: RNativeType>(sexp: SEXP) -> Option<arrow_buffe
 ///
 /// Must be called on R's main thread.
 ///
+/// # Allocation failure
+///
+/// `Rf_allocVector` raises an R error (longjmp) on OOM rather than returning
+/// a null pointer. The framework's `R_UnwindProtect` wrapper catches this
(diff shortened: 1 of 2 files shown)


CGMossa (commented):
Verified scope (2 files, both expected) and walked the audit plan items 10/11 through the new code. Validation logic is correctly placed, names the offending column, and handles empty / single-column / all-equal cases. The testthat fixture is appropriately loose on the error chain. One factual nit on the doc wording for `alloc_r_backed_buffer` — the OOM path doesn't actually surface as a tagged Rust-panic message; it re-raises R's native allocation error via `R_ContinueUnwind`. Non-blocking; willing to land as-is if you'd rather follow up.

<details><summary>AI review (Opus 4.7)</summary>

**Verdict: DISCUSSION NEEDED → leaning APPROVE.** One factual nit on the item 10 doc; item 11 is good.

### Item 10 — `alloc_r_backed_buffer` doc (`miniextendr-api/src/optionals/arrow_impl.rs:360-368`)

The new doc says:

> `Rf_allocVector` raises an R error (longjmp) on OOM rather than returning a null pointer. The framework's `R_UnwindProtect` wrapper catches this longjmp and converts it to a Rust panic, which surfaces to the caller as a "memory exhausted" R error.

This is **misleading**. Tracing `run_r_unwind_protect` (`unwind_protect.rs:247-252` and `296-311`):

1. R longjmps → `cleanup_handler` fires with `jump=TRUE` → `panic_any(RErrorMarker)` (intermediate Rust panic, true).
2. Outer `catch_unwind` recognises the `RErrorMarker` payload and calls `R_ContinueUnwind(token)` (`unwind_protect.rs:305`) — which **diverges and re-raises R's original error** ("cannot allocate vector of size …" / "vector memory exhausted", verbatim from R).

Both `with_r_unwind_protect` (line 437) and `with_r_unwind_protect_error_in_r` (line 500) explicitly document this: *"R-origin errors (longjmp) still pass through via `R_ContinueUnwind`"*. There is **no conversion to a tagged Rust-error SEXP**; the user sees R's native allocation error, not a "memory exhausted" string the framework synthesised. The new doc reads as if the framework reformats the message — it doesn't.

Suggested rewording (preserves the actionable insight that the `.expect()` calls are unreachable):

```
/// # Allocation failure
///
/// `Rf_allocVector` raises an R error (longjmp) on OOM rather than returning
/// a null pointer. The framework's `R_UnwindProtect` wrapper catches this
/// longjmp internally (via an `RErrorMarker` panic) and re-raises it via
/// `R_ContinueUnwind`, so the caller sees R's native allocation error
/// ("cannot allocate vector of size …") — not a Rust-side message. The two
/// `.expect()` calls below are therefore unreachable in production; they
/// guard against logic errors (wrong `len` encoding, unexpected null
/// pointer), not against OOM.
```

Mentioning `with_r_unwind_protect_error_in_r` by name (the canonical user-facing wrapper) would also help readers cross-reference the existing transport docs.

This is a doc-only nit — non-blocking if you'd rather land as-is and follow up.

### Item 11 — RecordBatch pre-validation (`arrow_impl.rs:1027-1043`)

Looks correct. Notes:

- **Placement**: between the column-build loop and `RecordBatch::try_new` — the right slot, so errors are framework-quality (`SexpError::InvalidValue`) rather than arrow's generic `RecordBatch: column N has length X expected Y` string.
- **Edge cases**:
  - Empty list (`columns.first()` → `None`): validation skipped; `RecordBatch::try_new(empty_schema, vec![])` succeeds — correct.
  - Single column: `iter().skip(1)` yields zero — fast-path, no work.
  - All-equal lengths: loop runs, no error — correct.
- **Column name lookup**: `names[i]` and `names[0]` are valid because `names` was built with one entry per `ncol` (lines 1015-1025) and `columns.len() == names.len() == ncol`.
- **Column length truthfulness**: `col.len()` on the built Arrow array corresponds to the R-side SEXP length for every dispatched branch in `sexp_column_to_arrow` (REALSXP/INTSXP/LGLSXP/STRSXP/RAWSXP/Date32/POSIXct).
- Minor: the error message says `"column 'y' has length 2 but column 'x' has length 3"` — i.e. always frames the rest against column 0. That's fine for usability; the plan suggested similar wording.

### Test — `rpkg/tests/testthat/test-arrow.R:149-160`

- Asserts the offending column name (`'y'`) appears in the error — matches the audit's "verify the improved error mentions the offending column name" requirement.
- `regexp = "'y'"` is appropriately loose: survives prefix/suffix changes in the error chain (framework wraps the message before R sees it).
- `fixed = FALSE` is redundant (the default) but harmless.

### Scope

Clean: only `arrow_impl.rs` + `test-arrow.R`. No leakage into generated wrappers, NAMESPACE, or unrelated files.

### CI

At review time: 4 pre-flight jobs green, 10 still in progress (R CMD check matrix, Rust lint, Sync Checks, CRAN-like). Worth a re-look once they settle before un-drafting.

### Bottom line

Item 11 is ship-ready. Item 10's doc is functionally fine but contains a factual drift on the OOM path — recommend tightening the wording (or accepting and following up). Either way, I'm comfortable approving.
</details>

Issue #596, opened 2026-05-17
release-workflow: track upstream r-windows rtools / r-universe macos-libs releases


CGMossa:
Follow-up tracking issue from the PR adding macOS SDK + CRAN libs pins to `minirextendr/inst/templates/r-release.yml`.

## What needs periodic rebasing

Two pinned references in the scaffolded template + docs go stale as upstream cuts new releases:

1. **`r-universe-org/macos-libs` tarball date** — the `cranlibs-everything.tar.xz` URL in `Gotcha 6` (template + `docs/RELEASE_WORKFLOW.md`) pins date `2025-12-13`. When the upstream cuts a newer release (typically when CRAN bumps its curated system-library versions), the URL must be bumped here. Source: https://github.com/r-universe-org/macos-libs/releases
2. **Windows rtools mingw release** — documented as `6768` in `docs/CRAN_COMPATIBILITY.md` "Layer 3" pin table. Not consumed by any workflow yet (see #594), but should be kept current as a reference. Source: https://github.com/r-windows/rtools-base/releases

## Suggested cadence

- Quarterly check.
- Force-rebase before any miniextendr release tag to ensure scaffolded packages target current CRAN values.

## Definition of done for each rebase

- Tarball date bumped in `minirextendr/inst/templates/r-release.yml` (Gotcha 6 step URL).
- Same URL bumped in `docs/RELEASE_WORKFLOW.md` Gotcha 6 prose.
- rtools value bumped in `docs/CRAN_COMPATIBILITY.md` Layer 3 table.
- Smoke-test the new tarball URL via a real CI run (download + tar extract paths haven't drifted).

## Why a tracking issue, not a cron action

The rebase is small and infrequent; a manual quarterly check is lower-friction than authoring a renovate-style bot for two URL pins. Revisit if the cadence accelerates.

Issue #595, opened 2026-05-17
ci: bring .github/workflows/ci.yml r-check-macos in line with r-release.yml CRAN pins


CGMossa:
Follow-up from the PR adding macOS SDK + CRAN libs pins to `minirextendr/inst/templates/r-release.yml` and documenting them in `docs/RELEASE_WORKFLOW.md` Gotchas 5 and 6 + `docs/CRAN_COMPATIBILITY.md` "Toolchain ABI matching".

## Problem

`.github/workflows/ci.yml`'s `r-check-macos` job currently runs on stock macOS runners without:

- `xcode-select -s` to pin the SDK to CRAN's binary target.
- `MACOSX_DEPLOYMENT_TARGET` floor in `$GITHUB_ENV`.
- `r-universe-org/macos-libs` system-library prefetch.

This means our own CI is **not** testing under the same toolchain ABI we tell downstream maintainers to use in their release workflows. CI is greener than CRAN's autobuilder would be on the same code, and a `-sys` Rust dep landing in our own workspace could pass CI and fail under CRAN.

## Scope (intentionally minimal)

Port the two new steps from `minirextendr/inst/templates/r-release.yml` ("Pin macOS SDK and deployment target" + "Download CRAN system libraries") into the `r-check-macos` job in `.github/workflows/ci.yml`. Single source-of-truth lives in the docs; the workflow steps are duplicated by design (no `uses:` indirection — same rationale as the template, see Gotcha 5 in `RELEASE_WORKFLOW.md`).

## Out of scope

- Adding new R versions to the matrix.
- Touching `r-check-linux` / `r-check-windows`.
- Refactoring the CI workflow structure.

## Why deferred

User answered "not now" on bundling this into the main PR — kept that PR scoped to the scaffolded template + docs so the diff stays reviewable.

Issue #594, opened 2026-05-17
release-workflow: add Windows build job using rtools45 toolchain


CGMossa:
Follow-up from the PR adding macOS SDK + CRAN libs pins to `minirextendr/inst/templates/r-release.yml`. The scaffolded release workflow currently builds on AlmaLinux 8 and macOS (arm64 + x86_64). Windows builds are not covered.

## Goal

Add a `build-windows` job mirroring the macOS layer of defense:

- Use `r-windows/rtools-base` (rtools45) toolchain to match CRAN's Windows binary build.
- Set up `setup-msys2` / `setup-rtools` per `r-devel/actions/setup-windows-tools@ec72e88`.
- Pin the rtools mingw release (currently `6768`); update this pin when r-windows ships a new release.
- Ensure cargo private-dep auth (gh credential helper) works under the rtools shell.

## Why a separate PR

- Encoding the rtools setup steps requires testing on a Windows runner; this PR deliberately stays macOS-only so the CI cost stays bounded.
- New gotchas to discover and document (path translation, msys vs powershell shell choice, link.exe vs gcc, locale handling).

## Starting point

- Upstream snippet: https://github.com/r-devel/actions/blob/ec72e88/setup-windows-tools/action.yml
- r-windows base: https://github.com/r-windows/rtools-base
- R-admin reference for Windows toolchain: https://cran.r-project.org/doc/manuals/r-devel/R-admin.html#Windows

## Definition of done

- New `build-windows` job in `minirextendr/inst/templates/r-release.yml`.
- Gotcha 7 section in `docs/RELEASE_WORKFLOW.md` documenting the rtools45 pin + msys2 quirks.
- Windows entry in `docs/CRAN_COMPATIBILITY.md` "Layer 3" pin table updated from "not currently consumed" to live values.
- Tests in `minirextendr/tests/testthat/test-monorepo-gaps.R` for the working-directory patcher updated if the patcher's regex anchors need to grow for the new job.

PR #593, opened 2026-05-17
site: gitignore generated manual/, add curated→manual cross-links (#76, #78)

CGMossa:
## Summary

Two related changes to the site manual's lifecycle.

### #78: gitignore generated `site/content/manual/*.md` (except `_index.md`)
- `scripts/docs-to-site.sh` regenerates these from `docs/`; CI runs the script before each `zola build`.
- Tracked + regenerated previously meant every doc edit produced an 80+ file diff and a drift-risk window.
- `_index.md` is the only file the script preserves — keep it tracked.

### #76: top-level pages link explicitly to the manual
- Top-level (`getting-started`, `architecture`, ...) is the curated entry; `/manual/<slug>/` is the exhaustive reference.
- Split already exists in practice; this PR makes the cross-link explicit so users follow the trail when they need depth.

## After-merge dev workflow

```bash
# Edit docs/
$EDITOR docs/SOMETHING.md

# Regenerate manual/ for local preview
just site-docs    # or: bash scripts/docs-to-site.sh

# Preview
just site-serve

# Commit docs/ change only — manual/ regeneration happens in CI
git add docs/SOMETHING.md
git commit
```

After pulling this PR, contributors will see `site/content/manual/*.md` as untracked in their working tree. Either delete them and re-run `bash scripts/docs-to-site.sh`, or leave them — `.gitignore` ignores them either way.

## Test plan
- [x] `git ls-files site/content/manual/` returns only `_index.md`
- [x] `bash scripts/docs-to-site.sh && git status` shows no `manual/*.md` modifications (gitignore effective)
- [x] Top-level pages with a `/manual/<slug>/` counterpart all have a "Full reference" footer link (11 pages)
- [ ] `cd site && zola build` (if Zola installed) — clean, no broken-link warnings

## Pages skipped (no manual counterpart)
- `site/content/api.md` — no `manual/api.md` exists
- `site/content/full-project-tutorial.md` — no `manual/full-project-tutorial.md` exists

## Not in this PR
- Flipping `internal_level` to `error` in `site/config.toml` — follow-up after #77 (PR-A) and this PR both land.

PR #600, opened 2026-05-17
feat(release-workflow): pin macOS SDK and CRAN libs prefix in scaffolded template

CGMossa:
The scaffolded `r-release.yml` now pins macOS Xcode + deployment target to CRAN's current binary targets (Sonoma 14.0 on arm64, Big Sur 11.0 on Intel) and pulls in `r-universe-org/macos-libs` so any future `-sys` Rust dep resolves against the same C libraries CRAN's R was built against. `RELEASE_WORKFLOW.md` and `CRAN_COMPATIBILITY.md` document the layered defense and cite both R-admin and `r-devel/actions` as the source-of-truth.

<details>
<summary><sub>AI-written summary, test plan, and follow-up tracking (Claude)</sub></summary>

## What changed

**`minirextendr/inst/templates/r-release.yml`** (1 file, +56/-3):

- Two new steps in `build-macos`, inserted before `r-lib/actions/setup-r`:
  - **Pin macOS SDK and deployment target.** `xcode-select -s` (Xcode 26.0 on arm64, 16.2 on x86_64) and `MACOSX_DEPLOYMENT_TARGET` written into `$GITHUB_ENV`. Branches on `uname -m` so the same step works on both runner archs.
  - **Download CRAN system libraries.** Fetches `r-universe-org/macos-libs` `cranlibs-everything.tar.xz` (date-pinned `2025-12-13`) into `/opt/R/<arch>/lib`, then sets `PKG_CONFIG_PATH`.
- Matrix expanded from single `runs-on: macos-14` to a 2x2 over `runner: [macos-14, macos-13]` x `r-version: ['release', 'oldrel-1']` so the SDK-pin x86_64 branch is actually exercised. Doubles the macOS jobs; release workflow is tag-push-only, so bounded.
- Both steps are inlined (not `uses:`-referenced) from [`r-devel/actions/setup-macos-tools@ec72e88`](https://github.com/r-devel/actions/blob/ec72e88/setup-macos-tools/action.yml) for hermetic-against-upstream-archival reasons. Attribution comments above each step.

**`docs/RELEASE_WORKFLOW.md`** (1 file, +123/-6):

- Gotcha 5 (macOS SDK / deployment target) and Gotcha 6 (CRAN system libraries) added in the same style as 1-4 (symptom / fix snippet / why it matters / citations).
- Both sections cite R-admin (r-svn `doc/manual/R-admin.texi:5854-5867`) and `r-devel/actions@ec72e88` with permalinks.
- Title and intro updated from "four gotchas" to "six gotchas."

**`docs/CRAN_COMPATIBILITY.md`** (1 file, +76):

- New top-level section "Toolchain ABI matching" framing the three-layer defense:
  - Layer 1: `./configure`-emitted `.cargo/config.toml` `[env]` per install.
  - Layer 2: CI workflow `env:` pins (this PR).
  - Layer 3: this doc + RELEASE_WORKFLOW.md citations.
- Includes a pin table and a "How to update the pins when CRAN moves" subsection listing the three sites to keep in sync.

## What was skipped vs. upstream r-devel/actions

(Documented inline in the template's attribution comments and in Gotcha 6 prose):

- `brew unlink $(brew list --formula)` - destructive on shared runners; unnecessary once `PKG_CONFIG_PATH` is set.
- gfortran install - no Fortran-linked Rust deps in miniextendr today.
- TinyTeX install - default scaffold builds with `--no-manual`.
- xQuartz install - no X11 dependencies.
- `/opt/R/<arch>/bin` on `$GITHUB_PATH` - `r-lib/actions/setup-r` puts R's bin on PATH independently.

## Patcher safety

`release_workflow_insert_workdir()` in `minirextendr/R/use-release-workflow.R` greps for two patterns:

- `^        run: bash \./configure\s*$` (configure step)
- `^        run: \|\s*$` *and* next line contains `R CMD build` (build step)

The two new steps use `run: |` but the next lines are `if [ "$(uname -m)" = "x86_64" ]` and `sudo mkdir -p /opt` - neither contains `R CMD build`, so the patcher correctly skips them. Verified by simulating the patcher logic against the new template (4 working-directory lines injected, exactly the 2 configure + 2 build steps across 2 jobs).

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(...)"` against the new template - parses cleanly.
- [x] `devtools::test(filter = "use-release-workflow |monorepo-gaps")` - 33 PASS, 0 FAIL.
- [x] Full `devtools::test()` - my changes add no new failures (pre-existing 6 failures in `test-scaffold-smoke.R` / `test-templates.R` / `test-use-native.R` reproduce on `origin/main` without my changes).
- [x] `scripts/docs-to-site.sh && zola build` - 95 pages built, 0 orphans, 0 errors.
- [ ] Real CI run on the release workflow - cannot exercise from a PR build (workflow triggers on `push: tags: ["v*"]` and `workflow_dispatch`). Will surface only on the next version tag.

## Out of scope - follow-up issues filed

- #594 - Add a `build-windows` job using rtools45 / `r-windows/rtools-base` mirroring the macOS layer of defense.
- #595 - Bring `.github/workflows/ci.yml`'s `r-check-macos` in line with the same CRAN pins so our own CI tests under the same toolchain ABI we tell downstream maintainers to use.
- #596 - Periodic rebase tracking issue for the two upstream pins (`r-universe-org/macos-libs` tarball date, `r-windows/rtools-base` rtools value).

## Cross-ref

A sibling PR (configure-emitted `[env]` in `.cargo/config.toml`) handles the per-install floor that complements this CI overlay. The docs reference it in prose without linking a specific PR number; the maintainer will weave the cross-ref after both merge.

</details>

### DIFF
--- a/docs/CRAN_COMPATIBILITY.md
+++ b/docs/CRAN_COMPATIBILITY.md
@@ -211,6 +211,82 @@ See CLAUDE.md "Vendor tarball is a latch" for the full context and the
 - `[ -f inst/vendor.tar.xz ]` is the only source-vs-tarball signal. Don't add
   a second one. Maintenance load lives in the number of switches.
 
+## Toolchain ABI matching
+
+The vendoring story above keeps **Rust dependency sources** in sync between
+maintainer and CRAN's builder. A second, orthogonal problem is keeping the
+**Rust toolchain's compile-time targets** (SDK, deployment floor, system
+library prefix) in sync with the C toolchain CRAN's R was built with. A
+mismatch here produces `.so`s that link locally and segfault under CRAN's R,
+or trip `--as-cran` notes about deployment-target drift.
+
+miniextendr defends against this with three layered checks. The first two are
+load-bearing, the third documents the values.
+
+### Layer 1 — Per-install floor (`./configure` emits `[env]`)
+
+`./configure` derives `MACOSX_DEPLOYMENT_TARGET` (and equivalents) from the
+host R's `R CMD config CC` flags at install time, then writes them into
+`.cargo/config.toml`'s `[env]` table. Every `R CMD INSTALL` of the
+miniextendr-based package — including end-user installs that never see a
+GitHub Actions runner — picks up the same values R is configured against.
+
+This is the load-bearing layer for end users. They don't run the release
+workflow; they `install.packages()` (binary) or `R CMD INSTALL` (source) and
+expect the result to work with the R they have.
+
+### Layer 2 — CI overlay (`r-release.yml` workflow `env:` pins)
+
+The CI workflow that produces release binaries pins the **CRAN-canonical**
+values directly via `MACOSX_DEPLOYMENT_TARGET` in `$GITHUB_ENV` plus
+`xcode-select -s` to select the matching SDK. These override whatever Layer 1
+derived: the release artifact targets CRAN's exact ABI floor, not the
+runner's host R's floor.
+
+GitHub Actions shell `env:` wins over cargo `[env]` by default (cargo's
+`[env]` table is "set if not already set" semantics), so the two layers
+cooperate without conflict: CI uses the pinned values, end-user installs use
+whatever the host R reports.
+
+The same workflow also prefetches CRAN's curated system libraries
+(`r-universe-org/macos-libs`) into `/opt/R/<arch>/lib` and points
+`PKG_CONFIG_PATH` at them, so any Rust `-sys` crate's `pkg-config` lookup
+resolves against the same C libraries CRAN's R was linked with.
+
+See [RELEASE_WORKFLOW.md Gotchas 5 and 6](./RELEASE_WORKFLOW.md#gotcha-5-macos-sdk-and-deployment-target-must-match-cran)
+for the exact YAML snippets and citations.
+
+### Layer 3 — Documented canonical values
+
+The current CRAN-canonical pins are:
+
+| Platform | Arch | Pin | Source |
+|---|---|---|---|
+| macOS | arm64 | `MACOSX_DEPLOYMENT_TARGET=11.0` for the **toolchain**; `14.0` for the **binary** | [R-admin §"Building binary packages"](https://github.com/r-devel/r-svn/blob/master/doc/manual/R-admin.texi#L5854) — `Xcode_26.0` |
+| macOS | x86_64 | `MACOSX_DEPLOYMENT_TARGET=11.0` | R-admin (same) — `Xcode_16.2` |
+| Windows | x86_64 | rtools45 / GCC 14, mingw runtime release `6768` | [`r-windows/rtools-base`](https://github.com/r-windows/rtools-base) |
+| Linux | x86_64 | distro-supplied glibc; no per-platform pin | n/a |
+
+The Windows rtools version is **not currently consumed** by miniextendr's
+template — there's no `build-windows` job in `r-release.yml` yet. The value
+is documented here for completeness, and configure handles the rtools linker
+pin for end-user Windows installs already once the per-install `[env]` floor
+lands. See [issue tracker](https://github.com/A2-ai/miniextendr/issues?q=is%3Aissue+rtools+windows) for the rtools rebase cadence.
+
+### How to update the pins when CRAN moves
+
+CRAN's macOS binary build moves SDKs roughly once per major macOS release.
+When it does, three places must be updated in lockstep:
+
+1. `minirextendr/inst/templates/r-release.yml` — the `xcode-select -s` path
+   and `MACOSX_DEPLOYMENT_TARGET` value in the "Pin macOS SDK and deployment
+   target" step.
+2. `docs/RELEASE_WORKFLOW.md` Gotcha 5 — the version table and prose.
+3. This file — the table in "Layer 3" above.
+
+Cross-reference `r-devel/actions/setup-macos-tools` to confirm the upstream
+values; the `r-devel/actions` repo tracks CRAN's actual build matrix.
+
 ## Symbols cleanup, for grep-bait
 
 Removed entirely from this codebase:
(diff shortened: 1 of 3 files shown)


CGMossa (commented):
## Reviewer notes (Opus, 2026-05-18)

### Verdict
clean — ready to merge

### Required changes
- (none)

- **Layer 3 table prose for arm64 is confusing.** `docs/CRAN_COMPATIBILITY.md:264` says
  `MACOSX_DEPLOYMENT_TARGET=11.0 for the **toolchain**; 14.0 for the **binary**`.
  The workflow only sets `14.0` (the binary value) on arm64, and Gotcha 5's own table
  in `docs/RELEASE_WORKFLOW.md:144-147` lists arm64 as `14.0` flat. Either drop the
  "11.0 for toolchain" half (it isn't pinned anywhere in this PR or `r-devel/actions`)
  or add a sentence explaining where the 11.0 toolchain floor comes from. Right now
  a future maintainer reading Layer 3 will hunt for an 11.0 pin that doesn't exist.
- **`sudo tar -xf libs.tar.xz -C / opt`** (template line 144, doc snippet line 224).
  Syntactically suspect — looks like a typo for `-C /opt`, but it's verbatim from
  upstream `r-devel/actions/setup-macos-tools@ec72e88:50`. GNU tar reads it as
  "change to `/`, extract only the `opt/` member," which works iff the tarball's
  top-level entry is `opt/`. The release-workflow job is workflow_dispatch / tag-push
  only and the SKIPPED `R CMD check / macOS` job means this hasn't actually run in
  CI yet. Worth a one-line comment in the template noting the literal-from-upstream
  spelling so a future "obvious fix" PR doesn't unbreak it.
- **Per-install Windows pin claim.** Layer 3 says "configure handles the rtools
  linker pin for end-user Windows installs already once the per-install `[env]`
  floor lands" — that's PR #601's territory and PR #601 is not yet merged. Sentence
  reads in present tense ("handles") but the precondition ("once #601 lands") is
  in the same sentence. Minor; consider future tense or a `<!-- pending #601 -->`
  marker.
- **`Co-Authored-By` trailer absent** on the three commits. Not strictly required
  by the project convention for ordinary feature commits and the PR body uses the
  AI-attribution `<details>` block correctly, so this is purely a hygiene nit.

### Looks good

- **Inline-vs-uses attribution.** Both new steps are inlined (not `uses:`-referenced)
  with attribution comments at template lines 105-116 and 130-137, including upstream
  permalink (`@ec72e88`) and the rationale ("stay hermetic if that repo is later
  archived or restructured"). The "what was skipped from upstream" enumeration in
  the template comment block matches the upstream `action.yml` 1:1 (brew unlink,
  gfortran, TinyTeX, xQuartz, `/opt/R/<arch>/bin` on `$GITHUB_PATH`) and the
  rationale for each skip is reasonable for miniextendr's surface.
- **CRAN value verification.** `MACOSX_DEPLOYMENT_TARGET=14.0` for arm64 +
  `Xcode_26.0`, `=11.0` for x86_64 + `Xcode_16.2`. Cross-checked against upstream
  `r-devel/actions/setup-macos-tools@ec72e88` lines 20-26 — exact match. The
  `R-admin.texi:5854-5867` citation is the canonical CRAN-side source.
- **Step ordering.** The two new steps sit at template lines 117 and 137, both
  before `r-lib/actions/setup-r@v2` (line 156) and `dtolnay/rust-toolchain@stable`
  (line 165). `echo "...=..." >> $GITHUB_ENV` is the workflow-level pattern (not
  shell-local `export`), so the env vars survive into subsequent steps. The plan's
  step-local-export failure mode does not apply here.
- **Build matrix coverage.** Matrix now expands to `runner: [macos-14, macos-13]`
  × `r-version: ['release', 'oldrel-1']` — both Sonoma/arm64 and Big Sur/Intel
  are exercised. `fail-fast: false` is set, so an arm64 failure won't mask an
  Intel one. Plan satisfied.
- **Cargo `[env]` interaction.** Doc correctly characterizes cargo's `[env]`
  table as "set if not already set" semantics by default (no `force = true`), so
  workflow-level `$GITHUB_ENV` values shadow them. CI overlay and per-install
  floor cooperate without explicit precedence wiring.
- **Templates sync.** `r-release.yml` lives only under `minirextendr/inst/templates/`
  — no `rpkg/.github/` parent (`rpkg/` is the dev-workflow exemplar, has no
  GitHub Actions of its own) and no entry in `patches/templates.patch`. CI's
  `Sync Checks` job passed. Nothing to mirror.
- **Site sync.** `site/content/manual/` was made gitignored in PR #593 (commit
  `87d5e7f1`, the base of this PR), so docs edits regenerate locally on
  `just site-docs` and do not need to be committed. No site-copy staleness in
  the diff.
- **Cross-references.** `RELEASE_WORKFLOW.md` Gotcha 5 → `CRAN_COMPATIBILITY.md#toolchain-abi-matching`
  resolves cleanly; `CRAN_COMPATIBILITY.md` Layer 2 → `RELEASE_WORKFLOW.md#gotcha-5-...`
  also resolves. "How to update the pins when CRAN moves" enumerates all three
  sites (template + Gotcha 5 + Layer 3) to keep in lockstep.
- **CI rollup.** 13 SUCCESS + 2 SKIPPED. The 2 SKIPPED (`Rust lint` and
  `R CMD check / macOS`) are path-filtered — this PR only touches docs +
  template YAML, neither Rust source nor anything that should fire macOS
  pre-release. Intentional and correct.
- **Follow-up issues filed.** #594 (build-windows), #595 (CI's `r-check-macos`
  alignment), #596 (periodic rebase tracking) — all referenced in the PR body
  with the deferred-items-as-issues discipline.
- **CLAUDE.md "end-user packages must build via standard R mechanisms".** The
  template uses only `bash ./configure` + `R CMD build/check` and standard
  GitHub Actions; no `just` invocation appears anywhere in the scaffolded
  workflow. Preserved.
- **Commit hygiene.** Title under 70 chars, conventional-commit prefix, three
  focused commits (template / gotchas-doc / layered-defense-doc) with thorough
  bodies citing upstream + rationale.

PR #603, opened 2026-05-17
feat(lint): MXL120 mirror — reject Self-returning vctrs constructors (#415)

### LINKED ISSUE #415: miniextendr-lint: mirror MXL120 (vctrs Self-returning ctor rejection)

`miniextendr-macros/src/miniextendr_impl.rs:2206-2266` confirms MXL120 lives only as a `syn::Error::new_spanned` inside `ParsedImpl::parse` (two error sites: constructor return type, instance-method receiver). It runs **only at proc-macro expansion time**.

`miniextendr-lint/src/rules/` has no `mxl120*` file; `miniextendr-lint/src/lint_code.rs:14-41` shows no `MXL120` variant. Grepping `miniextendr-lint/` for "MXL120" returns zero hits → **the lint-side mirror does not exist** → issue **still-applies** in full.

### Punchline rule

Issue body has no "Options:" block; the rule number (MXL120) and detection criteria are explicit. No ambiguity. Implementer can proceed.

### Goal

`miniextendr-lint` flags any `impl` block annotated with `#[miniextendr(vctrs(...))]` where (a) a constructor (`new` or `#[miniextendr(constructor)]`-tagged) returns `Self`, `&Self`, `&mut Self`, the named impl type, `Box<Self>`, `Result<Self, _>`, or any reference variant; or (b) any method has an instance receiver (`&self`, `&mut self`, `self`, `self: ExternalPtr<Self>` and refs/muts of that). Diagnostic code is `MXL120`. Fires from `build.rs` static-analysis pass, so third-party tooling / lint-only runs catch it without needing macro expansion to succeed.

### Constraints

- **Mirror, do not duplicate**: the proc-macro check at `miniextendr_impl.rs:2206` is the authoritative source. The lint must produce the same verdict on the same source. If a future patch relaxes either, both should move together (note in the lint rule's doc comment).
- **Don't break existing tests**: `miniextendr-macros/src/miniextendr_impl/tests.rs:786,833,868,927,2118,2141` already cover MXL120 at macro level; new lint tests must not collide.
- **No backwards compat** (`CLAUDE.md`): keep the lint's severity at `Error` to match `LintCode::MXL120`'s sibling errors (MXL008/9/10/110), since the macro raises a hard error.
- **Stay within `miniextendr-lint`'s capabilities** (per `miniextendr-lint/CLAUDE.md`): file-local AST walk with `cfg` evaluation in `crate_index.rs`; no cross-crate type resolution.

### File:line refs

- `miniextendr-lint/src/lint_code.rs:14-41` — add `MXL120` variant; `lint_code.rs:58` adjust severity classification (Error tier alongside MXL008/9/10).
- `miniextendr-lint/src/rules.rs:7-14` — add `pub mod vctrs_self_ctor;` and dispatch.
- `miniextendr-lint/src/rules/vctrs_self_ctor.rs` — **new file**. Pattern: copy `s4_method_prefix.rs` (similar shape — iterates `data.impl_methods` and filters by class_system).
- `miniextendr-lint/src/crate_index.rs:530-553` — `impl_methods` currently stores `(method_name, line, class_system)`. **Needs extending**: also store the return-type tokens and receiver-kind so the new rule can inspect them. Touch the `methods.push((...))` site and the `pub type / struct` definition (`crate_index.rs:117`).
- `miniextendr-macros/src/miniextendr_impl.rs:2834-2865` — `vctrs_ctor_returns_self_or_type` helper. Worth duplicating logic (lint crate doesn't depend on macros crate). Keep the two implementations side-by-side in tree; doc-link from each to the other.
- `miniextendr-lint/CLAUDE.md` — add MXL120 to the rule registry table.

### Flat work items

1. Add `LintCode::MXL120` to `lint_code.rs:14-41`. Classify as `Severity::Error` at `lint_code.rs:58` (joins MXL008/9/10/110 group).
2. Extend `crate_index.rs` impl-method tracking to capture `(method_name, line, class_system, return_type_str, receiver_kind)` — or carry the raw `syn::ImplItemFn` if simpler. The latter is easier but balloons memory; prefer the tuple expansion. Receiver kind enum: `None | Ref | RefMut | Value | ExternalPtrRef | ExternalPtrRefMut | ExternalPtrValue` (mirrors `ReceiverKind` in `miniextendr_impl.rs`).
3. Identify constructors in the lint pass: any method that (a) is `new` with no receiver, or (b) has a `#[miniextendr(constructor)]` attribute. The crate_index walker already collects method-level attributes via `extract_param_names` adjacent path — extend to also check for `#[miniextendr(constructor)]`. If too invasive, fall back to the `name == "new"` heuristic alone; the proc-macro will catch the `#[miniextendr(constructor)]` case at expansion time as a defence-in-depth.
4. Create `miniextendr-lint/src/rules/vctrs_self_ctor.rs` — duplicate the `vctrs_ctor_returns_self_or_type` predicate logic from `miniextendr-macros/src/miniextendr_impl.rs:2834`. Pattern: parse the recorded return-type string back to `syn::Type` (or store the parsed `syn::ReturnType` in the index).
5. Implement two checks in the new rule file:
   - **Check 1**: For each impl with `class_system == "vctrs"`, find the constructor (per item 3). If the return type satisfies the Self-or-named-type predicate, emit MXL120 at the constructor's line with the message used by the macro (`miniextendr_impl.rs:2217-2227`).
   - **Check 2**: For each impl with `class_system == "vctrs"`, iterate methods; if `receiver_kind.is_instance()`, emit MXL120 at the method line with the macro's message (`miniextendr_impl.rs:2245-2261`).
6. Register the rule in `rules.rs` (`pub mod vctrs_self_ctor;` + `vctrs_self_ctor::check(index, &mut diagnostics);`).
7. Add a test fixture under `miniextendr-lint/tests/` (or wherever existing rules have integration tests — check `s4_method_prefix.rs` placement). Fixture should include a vctrs impl with `fn new() -> Self`, a vctrs impl with `fn foo(&self)`, and a control vctrs impl with `fn new() -> Vec<f64>` (which should NOT trip). Assert MXL120 fires twice on the broken cases and not on the control.
8. Update `miniextendr-lint/CLAUDE.md` rule registry table to include MXL120.

### Non-goals

- **Not refactoring** the proc-macro `vctrs_ctor_returns_self_or_type` into a shared crate. The lint and macro intentionally stay independent (per `miniextendr-lint/CLAUDE.md`: "Shares the parser layer with `miniextendr-macros`" — but that's parser, not predicate). Duplication is fine and reviewable.
- **Not changing the macro-side check**: macro keeps raising the hard error so the lint is defense-in-depth (e.g., for IDE / `cargo clippy`-style runs that don't run codegen).
- **Not adding `// mxl::allow(MXL120)` suppression**. MXL120 is a hard error; suppressing it just produces broken code at runtime. The macro-side error has no suppression mechanism either.
- **Not closing #248**: the issue body says "Closes (together with #414): #248" — only the lint mirror plus the existing macro check closes it. Tag #248 in the PR.

### Validation checklist

- `just check` — green.
- `just clippy` — green.
- `just fmt` — clean.
- CI `clippy_default` + `clippy_all` (feature list from `.github/workflows/ci.yml`).
- `cargo test -p miniextendr-lint` — new fixture tests pass.
- `just devtools-test` — no rpkg regressions (lint is build.rs-time; integration verifies it doesn't break the workspace build).
- No wrapper regen needed (lint rule has no R output).

### Expected diff shape

- `miniextendr-lint/src/lint_code.rs` — 2 hunks (~6 lines).
- `miniextendr-lint/src/rules.rs` — 2 hunks (~4 lines).
- `miniextendr-lint/src/crate_index.rs` — 1 hunk extending the method tuple (~15 lines incl. `pub type` aliases or struct fields).
- `miniextendr-lint/src/helpers.rs` — possibly new helper for return-type inspection (~15 lines).
- `miniextendr-lint/src/rules/vctrs_self_ctor.rs` — new file (~80 lines incl. doc + 2 checks).
- `miniextendr-lint/tests/vctrs_self_ctor.rs` — new fixture (~60 lines).
- `miniextendr-lint/CLAUDE.md` — 1 line in rule registry.

Total ~180 lines net.

### Test plan

- **Positive (no diagnostic)**: vctrs impl with `fn new(values: Vec<f64>) -> Vec<f64>` and only static methods — clean.
- **Negative (MXL120 fires)**: vctrs impl with `fn new() -> Self`.
- **Negative (MXL120 fires)**: vctrs impl with `fn new() -> Result<Self, String>`.
- **Negative (MXL120 fires)**: vctrs impl with `fn value(&self) -> f64`.
- **Negative (MXL120 fires)**: vctrs impl with `fn value(self: &ExternalPtr<Self>) -> f64`.
- **Non-vctrs control**: r6 impl with `fn new() -> Self` — NOT flagged (MXL120 is vctrs-specific).

## Original issue

miniextendr-lint: mirror MXL120 (vctrs Self-returning ctor rejection)
---
## Background

PR #414 added a proc-macro hard error (MXL120) that fires when a `#[miniextendr(vctrs(...))]` impl constructor returns `Self` or the named impl type. The check fires at proc-macro expansion time inside `ParsedImpl::parse`.

## What this issue tracks

Mirror the same check in `miniextendr-lint` so it also fires during the build-time static analysis pass (when the proc-macro isn't being expanded, e.g. when lint-only analysis is requested or in third-party tooling).

The lint should flag any `impl` block with `#[miniextendr(vctrs(...))]` where a method named `new` (or tagged `#[miniextendr(constructor)]`) has a return type that is `Self`, the named struct type, `Box<Self>`, `Result<Self, _>`, or any reference variant.

Rule number: MXL120 (matches the proc-macro error).

## References

- Closes (together with #414): #248
- Implementation precedent: `miniextendr-lint/src/lib.rs` + `crate_index.rs` — module-tree walker with cfg feature eval

CGMossa:
Closes #415.
Also closes (together with #414): #248.

## Summary
- Adds `MXL120` to `miniextendr-lint` as a build-time static-analysis mirror of the proc-macro hard error already in `miniextendr-macros/src/miniextendr_impl.rs`. The lint fires during lint-only IDE/tooling runs before macro expansion.
- Two checks: (1) vctrs constructor must not return `Self`/named-type/`Result<Self,_>`/`Box<Self>`; (2) no instance receiver (`&self`, `self: &ExternalPtr<Self>`, etc.) on any method in a `#[miniextendr(vctrs(...))]` impl.
- Fixes `parse_miniextendr_impl_attrs` helper to use paren-balanced comma splitting, so `vctrs(kind = "vctr", base = "double", ...)` attrs are parsed correctly (the naive `split(',')` broke on inner commas inside the vctrs sub-expression).

## Changed files
- `miniextendr-lint/src/lint_code.rs` — add `MXL120` variant, `Severity::Error`
- `miniextendr-lint/src/crate_index.rs` — add `MethodReceiverKind` enum + `ImplMethodEntry` struct; extend `impl_methods` map to carry return-type string, receiver kind, and constructor-attr flag
- `miniextendr-lint/src/helpers.rs` — fix `parse_miniextendr_impl_attrs` comma splitting; add `split_top_level_commas` helper
- `miniextendr-lint/src/rules/vctrs_self_ctor.rs` — new rule file (two checks + duplicated `ty_is_self_or_named` predicate mirroring the macro)
- `miniextendr-lint/src/rules.rs` — register `vctrs_self_ctor`
- `miniextendr-lint/src/rules/s4_method_prefix.rs` — update to use new `ImplMethodEntry` struct fields
- `miniextendr-lint/tests/tests.rs` — 6 new MXL120 test cases (4 negative, 2 positive controls)
- `miniextendr-lint/CLAUDE.md` — add MXL120 to rule registry
- `rpkg/src/rust/*.rs` — pre-existing formatting drift fixed by `just fmt`

## Test plan
- [x] `cargo test -p miniextendr-lint` — 25 tests pass (4 new MXL120 negative cases + 2 positive controls)
- [x] `just check` — green across all workspace manifests
- [x] `just clippy` — clean
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (CI clippy_default) — clean
- [x] `cargo clippy --workspace --all-targets --locked --features rayon,... -- -D warnings` (CI clippy_all) — clean
- [x] `just fmt` — clean
- [x] `just lint` — `miniextendr-lint: no issues found`

## Follow-ups
- #608: migrate `is_altrep_struct` in `helpers.rs:300` to use the new `split_top_level_commas` helper (same naive `split(',')` bug class fixed in this PR for `parse_miniextendr_impl_attrs`; out of scope here since current ALTREP mode attrs don't take nested parens).

### DIFF
--- a/miniextendr-lint/CLAUDE.md
+++ b/miniextendr-lint/CLAUDE.md
@@ -18,6 +18,7 @@ Build-time static analysis. Runs from a downstream crate's `build.rs` (via the `
 - **MXL110** — parameter name is an R reserved word.
 - **MXL111** — `s4_*` method on `#[miniextendr(s4)]` impl (codegen auto-prefixes — yields `s4_s4_*`).
 - **MXL112** — explicit lifetime param on `#[miniextendr]` fn/impl.
+- **MXL120** — vctrs constructor returns `Self`/named type, or impl has an instance-method receiver (`&self`, `self: &ExternalPtr<Self>`, etc.). Mirrors the proc-macro hard error in `miniextendr-macros`.
 - **MXL203** — redundant `internal` + `noexport`.
 - **MXL300** — direct `Rf_error`/`Rf_errorcall` → replace with `panic!()` (framework converts to R error via tagged-SEXP transport).
 - **MXL301** — `_unchecked` FFI outside known-safe contexts (ALTREP callbacks, `with_r_unwind_protect`, `with_r_thread`).
--- a/miniextendr-lint/src/crate_index.rs
+++ b/miniextendr-lint/src/crate_index.rs
@@ -15,6 +15,74 @@ use crate::helpers::{
     is_altrep_struct, parse_miniextendr_impl_attrs,
 };
 
+// region: Impl method entry
+
+/// Receiver kind for an impl method, mirroring `ReceiverKind` in `miniextendr-macros`.
+///
+/// Mirror: `miniextendr-macros/src/miniextendr_impl.rs` — `ReceiverKind`.
+/// Keep both in sync: if the macro relaxes one receiver kind, update this enum too.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MethodReceiverKind {
+    /// No self — static / associated function.
+    None,
+    /// `&self`
+    Ref,
+    /// `&mut self`
+    RefMut,
+    /// `self` (consuming)
+    Value,
+    /// `self: &ExternalPtr<Self>`
+    ExternalPtrRef,
+    /// `self: &mut ExternalPtr<Self>`
+    ExternalPtrRefMut,
+    /// `self: ExternalPtr<Self>`
+    ExternalPtrValue,
+}
+
+impl MethodReceiverKind {
+    /// Returns true if this is an instance receiver (any form of `self`).
+    pub fn is_instance(self) -> bool {
+        matches!(
+            self,
+            Self::Ref
+                | Self::RefMut
+                | Self::Value
+                | Self::ExternalPtrRef
+                | Self::ExternalPtrRefMut
+                | Self::ExternalPtrValue
+        )
+    }
+
+    /// Human-readable spelling used in diagnostic messages.
+    pub fn spelling(self) -> &'static str {
+        match self {
+            Self::None => "(none)",
+            Self::Ref => "&self",
+            Self::RefMut => "&mut self",
+            Self::Value => "self",
+            Self::ExternalPtrRef => "self: &ExternalPtr<Self>",
+            Self::ExternalPtrRefMut => "self: &mut ExternalPtr<Self>",
+            Self::ExternalPtrValue => "self: ExternalPtr<Self>",
+        }
+    }
+}
+
+/// Per-method data collected during the crate-index pass for impl-method lint rules.
+#[derive(Clone, Debug)]
+pub struct ImplMethodEntry {
+    pub method_name: String,
+    pub line: usize,
+    pub class_system: String,
+    /// Stringified return type tokens (empty string = `()` / no explicit return).
+    pub return_type_str: String,
+    /// Receiver kind detected from the method signature.
+    pub receiver_kind: MethodReceiverKind,
+    /// True when the method carries `#[miniextendr(constructor)]`.
+    pub has_constructor_attr: bool,
+}
+
+// endregion
+
 // region: Lint item types
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
@@ -114,8 +182,8 @@ pub struct FileData {
     pub export_control: HashMap<String, (bool, bool, usize)>,
 
     // Impl method details for per-method lint rules
-    /// Methods per inherent impl type: type_name → Vec<(method_name, line, class_system)>.
-    pub impl_methods: HashMap<String, Vec<(String, usize, String)>>,
+    /// Methods per inherent impl type: type_name → Vec<ImplMethodEntry>.
+    pub impl_methods: HashMap<String, Vec<ImplMethodEntry>>,
 
     // Doc-comment roxygen tags per function/impl name
     /// Known roxygen tags: "@noRd", "@export", "@keywords internal"
@@ -538,18 +606,26 @@ fn collect_items_recursive(items: &[Item], data: &mut FileData) {
                                     line,
                                 ));
 
-                                // Collect method names for per-method rules (e.g. MXL111)
+                                // Collect method names for per-method rules (e.g. MXL111, MXL120)
                                 let methods =
                                     data.impl_methods.entry(type_name.clone()).or_default();
                                 for impl_item in &item_impl.items {
                                     if let syn::ImplItem::Fn(method) = impl_item {
                                         let method_name = method.sig.ident.to_string();
                                         let method_line = method.sig.ident.span().start().line;
-                                        methods.push((
+                                        let return_type_str =
+                                            extract_return_type_str(&method.sig.output);
+                                        let receiver_kind = detect_receiver_kind(&method.sig);
+                                        let has_constructor_attr =
+                                            has_constructor_attr(&method.attrs);
+                                        methods.push(ImplMethodEntry {
                                             method_name,
-                                            method_line,
-                                            class_system.clone(),
-                                        ));
+                                            line: method_line,
+                                            class_system: class_system.clone(),
+                                            return_type_str,
+                                            receiver_kind,
+                                            has_constructor_attr,
+                                        });
                                     }
                                 }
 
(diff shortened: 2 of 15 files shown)


CGMossa (commented):

### Verdict

Minor — code is mergeable as-is. One semantic divergence and a few coverage gaps worth flagging, but nothing blocks merge given the macro-side check is the authoritative gate.

### Required changes

None blocking.

- **Semantic divergence: `MethodReceiverKind::Value` in `is_instance()`.**
  - `miniextendr-lint/src/crate_index.rs:44-54` includes `Self::Value` (consuming `self`) in `is_instance()`.
  - Macro counterpart `miniextendr-macros/src/miniextendr_impl.rs:393-402` `ReceiverKind::is_instance()` **excludes** `Value`.
  - Practical impact is small because `Value` is already rejected upstream at `miniextendr_impl.rs:1979-2004` unless the method is a constructor or `#[miniextendr(finalize)]`. `finalize` is R6-only (rejected for vctrs at `miniextendr_impl.rs:1165`), so the only realistic divergence is a vctrs method like `fn convert(self) -> Vec<f64>` with `#[miniextendr(constructor)]` — the macro lets it through MXL120's Check 2 (which uses the exclude-Value `is_instance`), but the lint flags it. Probably nobody writes that, but the lint should mirror the macro exactly. Either:
    - drop `Value` from the lint's `is_instance()`, or
    - add a comment explaining why the lint is intentionally stricter (and ideally back-port the same widening to the macro so they remain in lockstep).

- **Missing test coverage.** `tests/tests.rs:535-714` covers `Self`, `Result<Self, _>`, `&self`, `self: &ExternalPtr<Self>`, a positive control, and an R6 negative. Plan/spec calls for more — recommend adding fixtures for:
  - `Box<Self>` return (predicate at `vctrs_self_ctor.rs:162-168` handles it but isn't exercised).
  - Named-type return (`fn new() -> MyVec` rather than `-> Self`) — exercises the `last.ident == type_name` branch at `vctrs_self_ctor.rs:152`.
  - `&Self` / `&mut Self` return (`syn::Type::Reference` branch at `vctrs_self_ctor.rs:172`).
  - `self: ExternalPtr<Self>` (by-value) and `self: &mut ExternalPtr<Self>` — currently only `self: &ExternalPtr<Self>` (ref) is tested. The `detect_receiver_kind` typed-self path at `crate_index.rs:784-808` has three branches and only one is exercised.
  - A method tagged `#[miniextendr(constructor)]` (not named `new`) — exercises the `has_constructor_attr` path at `vctrs_self_ctor.rs:53` and `crate_index.rs:836-858`.

- **Type-alias-for-`Self` limitation.** `return_type_is_self_or_named` (`vctrs_self_ctor.rs:131`) is purely token-string-based. `type Aliased = Self; pub fn new() -> Aliased { ... }` is not detected. Same gap exists in the macro intentionally (cross-crate type resolution is out of scope). A short doc-comment noting this would help future readers — the lint doc-comment at `vctrs_self_ctor.rs:124-130` could mention it.

- **`is_altrep_struct` still uses naive `tokens.split(',')`.** Pre-existing at `helpers.rs:300`. Not blocking today because the recognized mode attrs (`list`, `dataframe`, `externalptr`) don't take parenthesised arguments, but for consistency with the new `split_top_level_commas` helper added in this PR (`helpers.rs:180-217`), this is a candidate for the same treatment — small follow-up issue worth filing.

### Looks good

- **Mirror semantics carefully aligned.** `ty_is_self_or_named` in `vctrs_self_ctor.rs:145-175` is a faithful port of the macro's helper at `miniextendr_impl.rs:2849-2882`: handles `Self`, named ident, `Result<_, _>`, `Box<_>`, references — same recursion pattern, same string-based match limitations.

- **`MethodReceiverKind` enum mirrors `ReceiverKind`.** All seven variants present, including all three `ExternalPtr<Self>` forms (ref / refmut / value) at `crate_index.rs:35-39`. `detect_receiver_kind` (`crate_index.rs:769-810`) correctly classifies typed-self forms via the dedicated `is_external_ptr_self_ty` helper at `crate_index.rs:814-832`.

- **`parse_miniextendr_impl_attrs` fix is load-bearing and correct.** Prior naive `tokens.split(',')` (`helpers.rs` HEAD~1:139) silently mis-parsed `vctrs(kind = "vctr", base = "double", ...)` — class_system never got set, so MXL111 and MXL120 would have silently no-opped on the most common real vctrs impls (e.g. `rpkg/src/rust/match_arg_impl_tests.rs:189`, `rpkg/src/rust/vctrs_derive_example.rs:253`). The new `split_top_level_commas` iterator at `helpers.rs:180-217` correctly handles `()`, `[]`, `{}` and double-quoted strings.

- **No regressions in callers of `impl_methods`.** `s4_method_prefix.rs:12-46` migrated cleanly to `ImplMethodEntry` fields. Only other consumer is the new `vctrs_self_ctor` rule.

- **`Severity::Error` is correct.** `lint_code.rs:70` matches the macro's hard-error semantics — runtime would panic with ".data must be a vector type" on the ctor path or `ErasedExternalPtr::from_sexp` panic on the receiver path, so this is unrecoverable.

- **Documentation in sync.** `miniextendr-lint/CLAUDE.md:24` rule-registry entry was added between MXL112 and MXL203, in the correct order. Doc comment on the new rule (`vctrs_self_ctor.rs:18-26`) explicitly calls out the mirror relationship and asks readers to keep both in sync.

- **All 25 lint tests pass; `cargo clippy -p miniextendr-lint --all-targets --locked -- -D warnings` clean.** PR description confirms both `clippy_default` and `clippy_all` were also run against the workspace.

- **No false positives on existing vctrs impls.** Verified `rpkg/src/rust/vctrs_derive_example.rs:253` (`DerivedCurrency::new -> Vec<f64>`), `rpkg/src/rust/match_arg_impl_tests.rs:189` (`VctrsMatchArgScale::new -> Vec<f64>`), and `rpkg/src/rust/rdata_sidecar_tests.rs:267` (empty `impl SidecarVctrs {}`) are correctly clean.

- **Rpkg `cargo fmt` drift.** The rpkg source changes (`columnar_option_none_tests.rs`, `dataframe_collections_test.rs`, `dataframe_enum_payload_matrix.rs`, etc.) are pure rustfmt output and consistent with the "Fix warnings you see" project rule — pre-existing drift swept up by `just fmt`. Out-of-scope for this PR's title but explicitly noted in the PR description.

Issue #604, opened 2026-05-17
DataFrameRow: type alias fields produce confusing downstream error instead of pointing at alias


CGMossa:
## Background

`classify_field_type` matches `HashMap`/`BTreeMap` by the last path segment. A type alias like `type Counts = HashMap<String, i32>; field: Counts` has `Counts` as the last segment — map expansion is NOT triggered.

Since #484 (PR feat/df-row-484-fallthrough), wrapper types (`Option<T>`, `Arc<T>`, etc.) now produce a clear compile error. But type aliases still fall through:

- Aliases over `HashMap`/`BTreeMap` → classified as `Struct`, triggering `_assert_inner_is_dataframe_row::<Counts>()` → compile error "Counts does not implement DataFrameRow" with no mention of the alias.
- Aliases over `Option<T>` / smart-pointer types → same as above, `Struct` path, same confusing error.

## Desired behavior

Ideally: emit a diagnostic pointing at the alias field and suggesting the user write the concrete type or add `#[dataframe(as_list)]`.

## Constraints

Proc macros cannot resolve type aliases (no typeck). Options:

1. Accept the status quo — alias → `_assert_inner_is_dataframe_row` error — and document clearly (current state post #484).
2. Heuristic name match (`*Map`, `*Hash`, etc.) — fragile, not recommended.
3. Require users to add `#[dataframe(as_list)]` explicitly for alias fields (already works).

## Current state

Documented in `docs/DATAFRAME.md` under "Detection caveats". No code change needed beyond PR #484 unless a better diagnostic approach is found.

CGMossa:
Closing as documented (no code change feasible).

The constraint in the issue body is binding: **proc macros cannot resolve type aliases** — there's no typeck at macro-expansion time. The three options listed in the issue boil down to (1) accept the status quo + document, (2) fragile name-heuristic, (3) require explicit `#[dataframe(as_list)]`. Path 1 is what landed.

`docs/DATAFRAME.md:249` covers this explicitly and links back here:

> **Type aliases** are not automatically unwrapped — `type Counts = HashMap<String, i32>; field: Counts` has `Counts` as the last segment, so map expansion is not triggered. Use the concrete type directly (`field: HashMap<String, i32>`), or annotate with `#[dataframe(as_list)]`. See #604 for tracking.

[...]

PR #605, opened 2026-05-17
feat(macros): reject Option<Map>/smart-pointer wrappers in DataFrameRow (#484)

### LINKED ISSUE #484: DataFrameRow: detection limits — type aliases and Option<Map>/Option<Inner> fall through silently

CGMossa:
Closes #484.

## Summary

- `classify_field_type` now returns `syn::Result<FieldTypeKind<'_>>` instead of the bare kind; all unrecognised wrapper types with generic args (`Option<T>`, `Cow<T>`, `Rc<T>`, `Arc<T>`, `RefCell<T>`, `Cell<T>`, `Mutex<T>`, `RwLock<T>`) produce a compile error pointing at the field type with a `#[dataframe(as_list)]` hint.
- All `as_list`-annotated callers use `.ok()` to swallow the error (opt-in makes wrapper types valid opaque list-columns); all other callers propagate via `?`.
- Two UI compile-fail tests added (`derive_dataframe_option_field`, `derive_dataframe_arc_field`); one positive test struct (`WithOptMapAsList`) confirms the escape hatch compiles and runs.
- `docs/DATAFRAME.md` updated: "Detection caveats" section rewritten to document the new hard errors, the `as_list` opt-out, and the remaining type-alias gap (tracked as #604).

## Test plan

- [x] `just check` — green
- [x] `just clippy` — green
- [x] `just fmt` — clean
- [x] CI `clippy_default` (`cargo clippy --workspace --all-targets --locked -- -D warnings`) — green locally
- [x] CI `clippy_all` (full feature list from `.github/workflows/ci.yml`) — green locally
- [x] `cargo test -p miniextendr-macros` — 308 passed (includes new UI compile-fail tests)
- [x] `Option<HashMap<String, i32>>` field without `as_list` → compile error with `as_list` hint
- [x] `Arc<String>` field without `as_list` → compile error with `as_list` hint
- [x] `Option<HashMap<String, i32>>` with `#[dataframe(as_list)]` → compiles, 2-row DataFrame with `counts` list-column

## Deferred

- #604 — type alias detection (`type Counts = HashMap<…>; field: Counts` still produces confusing downstream error; proc macros cannot resolve aliases without typeck)

### DIFF
--- a/docs/DATAFRAME.md
+++ b/docs/DATAFRAME.md
@@ -246,7 +246,7 @@ struct Row {
 }
 ```
 
-**Type aliases** are not automatically unwrapped — `type Counts = HashMap<String, i32>; field: Counts` has `Counts` as the last segment, so map expansion is not triggered. Use the concrete type directly (`field: HashMap<String, i32>`), or annotate with `#[dataframe(as_list)]`.
+**Type aliases** are not automatically unwrapped — `type Counts = HashMap<String, i32>; field: Counts` has `Counts` as the last segment, so map expansion is not triggered. Use the concrete type directly (`field: HashMap<String, i32>`), or annotate with `#[dataframe(as_list)]`. See [#604](https://github.com/A2-ai/miniextendr/issues/604) for tracking.
 
 Note: multi-segment paths whose last segment does NOT implement `DataFrameRow` (e.g. `std::ffi::CString`) produce a clear compile-time error from the `_assert_inner_is_dataframe_row` assertion — this is intentional. Use `#[dataframe(as_list)]` on the field or an import alias to a newtype wrapper if a non-DataFrameRow stdlib type needs to be stored.
 
--- a/docs/DATAFRAME.md
+++ b/docs/DATAFRAME.md
@@ -227,11 +227,26 @@ Absent-variant rows produce `NULL` in both columns (not NA). An empty map produc
 
 **`as_list` opt-out**: annotate the field with `#[dataframe(as_list)]` to keep it as a single opaque named-list column (the pre-expansion behavior). Only use this when the named-list per-row shape is needed directly in R.
 
-**Detection caveats**: `classify_field_type` detects `HashMap` / `BTreeMap` by matching the last path segment (`HashMap` or `BTreeMap`) and requiring exactly two generic type arguments. It also detects struct-typed fields by matching bare path types (single- or multi-segment, e.g. `Point` or `crate::geom::Point`) whose last segment has no generic arguments. Two shapes are not detected and fall through to `Scalar` (opaque list-column); both are tracked under [#484](https://github.com/A2-ai/miniextendr/issues/484):
+**Detection caveats**: `classify_field_type` detects `HashMap` / `BTreeMap` by matching the last path segment (`HashMap` or `BTreeMap`) and requiring exactly two generic type arguments. It also detects struct-typed fields by matching bare path types (single- or multi-segment, e.g. `Point` or `crate::geom::Point`) whose last segment has no generic arguments.
 
-- **Type aliases**: `type Counts = HashMap<String, i32>; field: Counts` — the last segment is `Counts`, not `HashMap`, so map expansion is not triggered. Use the concrete type directly, or annotate with `#[dataframe(as_list)]` and handle the named-list in R.
-- **`Option<HashMap<K,V>>`**: the outer segment is `Option` with one type argument, so the two-argument `HashMap`/`BTreeMap` guard is never reached. Unwrap the `Option` before storing (e.g., store `HashMap<K,V>` and push an empty map for the `None` case), or annotate with `#[dataframe(as_list)]`.
-- **`Option<UserStruct>`** where `UserStruct: DataFrameRow` — silently degrades to Scalar instead of recursively flattening. Tracked under [#484](https://github.com/A2-ai/miniextendr/issues/484).
+**Rejected wrapper types** — the following shapes produce a compile error (since #484) because they cannot be automatically expanded and would otherwise silently produce a confusing opaque list-column:
+
+- `Option<T>` — including `Option<HashMap<K,V>>`, `Option<UserStruct>`, etc.
+- `Cow<T>`, `Rc<T>`, `Arc<T>`, `RefCell<T>`, `Cell<T>`, `Mutex<T>`, `RwLock<T>`
+
+For all of these, use `#[dataframe(as_list)]` to opt into an explicit opaque list-column, or unwrap to the inner type (e.g. store `HashMap<K,V>` directly and use an empty map for the absent case):
+
+```rust
+#[derive(Clone, DataFrameRow)]
+struct Row {
+    id: i32,
+    // `counts: Option<HashMap<String, i32>>` → compile error without `as_list`.
+    #[dataframe(as_list)]
+    counts: Option<HashMap<String, i32>>,
+}
+```
+
+**Type aliases** are not automatically unwrapped — `type Counts = HashMap<String, i32>; field: Counts` has `Counts` as the last segment, so map expansion is not triggered. Use the concrete type directly (`field: HashMap<String, i32>`), or annotate with `#[dataframe(as_list)]`.
 
 Note: multi-segment paths whose last segment does NOT implement `DataFrameRow` (e.g. `std::ffi::CString`) produce a clear compile-time error from the `_assert_inner_is_dataframe_row` assertion — this is intentional. Use `#[dataframe(as_list)]` on the field or an import alias to a newtype wrapper if a non-DataFrameRow stdlib type needs to be stored.
 
--- a/miniextendr-macros/src/dataframe_derive.rs
+++ b/miniextendr-macros/src/dataframe_derive.rs
@@ -325,30 +325,7 @@ pub(super) fn classify_field_type(ty: &syn::Type) -> syn::Result<FieldTypeKind<'
     if let syn::Type::Path(type_path) = ty
         && let Some(seg) = type_path.path.segments.last()
         && let syn::PathArguments::AngleBracketed(args) = &seg.arguments
-        && let Some(syn::GenericArgument::Type(inner)) = args.args.first()
     {
-        // Check for Vec<T>
-        if seg.ident == "Vec" {
-            return Ok(FieldTypeKind::VariableVec(inner));
-        }
-
-        // Check for Box<[T]>
-        if seg.ident == "Box"
-            && let syn::Type::Slice(slice) = inner
-        {
-            return Ok(FieldTypeKind::BoxedSlice(&slice.elem));
-        }
-
-        // Check for HashMap<K, V> and BTreeMap<K, V>
-        if (seg.ident == "HashMap" || seg.ident == "BTreeMap")
-            && let Some(syn::GenericArgument::Type(val_ty)) = args.args.iter().nth(1)
-        {
-            return Ok(FieldTypeKind::Map {
-                key_ty: inner,
-                val_ty,
-            });
-        }
-
         // Reject wrapper types that would silently fall through to Scalar /
         // Struct and produce a confusing opaque list-column or a downstream
         // DataFrameRow assertion error.  These are the common smart-pointer
@@ -359,6 +336,12 @@ pub(super) fn classify_field_type(ty: &syn::Type) -> syn::Result<FieldTypeKind<'
         // checking (which is unavailable in proc macros). The user must either
         // unwrap to the inner type, or annotate with `#[dataframe(as_list)]`
         // to opt into an explicit opaque list-column.
+        //
+        // IMPORTANT: The rejection fires on *path identity alone*, before we
+        // inspect generic args.  `Cow<'a, T>` has a lifetime as its first
+        // generic argument, not a type; inspecting `args.args.first()` as a
+        // `GenericArgument::Type` would silently skip `Cow`.  Checking ident
+        // before args makes the rejection robust to any generic shape.
         const REJECTED_WRAPPERS: &[&str] = &[
             "Option", "Cow", "Rc", "Arc", "RefCell", "Cell", "Mutex", "RwLock",
         ];
@@ -373,6 +356,42 @@ pub(super) fn classify_field_type(ty: &syn::Type) -> syn::Result<FieldTypeKind<'
                      a sentinel / empty collection for the absent case)."
                 ),
             ));
+        }
+
+        // For the collection types below we need the first *type* argument.
+        // Skip any leading lifetime or const arguments (e.g. `Cow<'a, B>`
+        // has a lifetime first, but `Cow` is already rejected above so we
+        // only reach here for other angle-bracketed types).
+        let first_type_arg = args.args.iter().find_map(|arg| {
+            if let syn::GenericArgument::Type(t) = arg {
+                Some(t)
+            } else {
+                None
+            }
+        });
+
+        if let Some(inner) = first_type_arg {
+            // Check for Vec<T>
+            if seg.ident == "Vec" {
+                return Ok(FieldTypeKind::VariableVec(inner));
+            }
+
+            // Check for Box<[T]>
+            if seg.ident == "Box"
+                && let syn::Type::Slice(slice) = inner
+            {
+                return Ok(FieldTypeKind::BoxedSlice(&slice.elem));
+            }
+
+            // Check for HashMap<K, V> and BTreeMap<K, V>
+            if (seg.ident == "HashMap" || seg.ident == "BTreeMap")
+                && let Some(syn::GenericArgument::Type(val_ty)) = args.args.iter().nth(1)
+            {
+                return Ok(FieldTypeKind::Map {
+                    key_ty: inner,
+                    val_ty,
+                });
+            }
         }
     }
 
--- a/miniextendr-macros/src/dataframe_derive.rs
+++ b/miniextendr-macros/src/dataframe_derive.rs
@@ -287,7 +287,12 @@ pub(super) enum FieldTypeKind<'a> {
 /// - Any non-scalar bare path type (single- or multi-segment, e.g. `Point` or
 ///   `crate::geom::Point`) -> `Struct`
 /// - Everything else (known scalars, generic types with args, `::abs::Paths`) -> `Scalar`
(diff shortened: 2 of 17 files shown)


CGMossa (commented):

### Verdict
**blocking** — one wrapper from `REJECTED_WRAPPERS` never actually triggers; one docs link omission.

### Required changes

- [ ] **`Cow<…>` is in `REJECTED_WRAPPERS` but never reached** (`miniextendr-macros/src/dataframe_derive.rs:325-377`).

  The outer `if let` guard pattern-matches `args.args.first()` as `Some(syn::GenericArgument::Type(inner))`. For `Cow<'a, B>` (signature `Cow<'a, B: ?Sized>`), the first generic arg is `GenericArgument::Lifetime('a)`, **not** `Type`. The `let-else` fails and the entire block is skipped. The path then falls through to the bare-ident check at line 400, which also fails (because `seg.arguments` is `AngleBracketed`, not `None`), and lands at `Ok(FieldTypeKind::Scalar)` — silently classifying `Cow<'a, str>` as a scalar.

  Net effect: a user-stated rejection for `Cow` is documented and listed in `REJECTED_WRAPPERS`, but the actual code never sees `seg.ident == "Cow"`. The PR description, error message, and `docs/DATAFRAME.md` all promise behavior that doesn't fire.

  Fix options: (a) drop `Cow` from `REJECTED_WRAPPERS` and the docs to match reality; or (b) restructure the path-classifier so wrappers whose first generic arg may be a lifetime are checked before extracting `inner`. Either is fine, but the current state is inconsistent. Either way, please add a UI fixture (`derive_dataframe_cow_field.rs`) so the regression is locked in.

- [ ] **`docs/DATAFRAME.md` does not link to #604** (`docs/DATAFRAME.md:249`).

  PR body and commit message state "type-alias gap deferred to #604", and #604 is filed and references this PR. But the docs paragraph on type aliases reads as a static caveat without naming the tracking issue. Add the `[#604](https://github.com/A2-ai/miniextendr/issues/604)` link inline so users hitting the alias footgun can subscribe / see context.

- The rpkg-side rustfmt churn (~400 lines across `dataframe_enum_payload_matrix.rs`, `dataframe_struct_flatten_test.rs`, `gc_stress_fixtures.rs`, `columnar_option_none_tests.rs`, `jiff_adapter_tests.rs`) is unrelated to the macro change. `git diff -w` reduces the net to roughly the planned ~155 lines. The drift is harmless and consistent with `just fmt`, but co-mingling it with a behavior-change PR makes review noisier. For future PRs, prefer a separate "fmt-only" commit or a precursor PR. Not a blocker since the formatting is in fact `cargo fmt` output.

- Only `Option` and `Arc` have UI fixtures (`derive_dataframe_option_field.rs`, `derive_dataframe_arc_field.rs`). `Rc`, `RefCell`, `Cell`, `Mutex`, `RwLock` are reachable code paths but rely on the implementer's claim that the same `REJECTED_WRAPPERS` arm handles them. Given the `Cow` bug above, a one-shot parametric UI test (or even one fixture per wrapper) would surface the next "listed but unreachable" silent gap. Low cost since each fixture is 12 lines.

- `WithOptMapAsList` only has a Rust `#[test]`-level assertion (`rpkg/src/rust/dataframe_collections_test.rs:973-999`). No `#[miniextendr]` export so `just devtools-test` won't exercise it from R. Existing fixtures in this file follow the same convention so this matches house style — flagging only because the reviewer brief asked. The Rust test does verify two rows and both `id` / `counts` companion-column lengths; that's adequate.

- The `into_list()` impl for `WithOptMapAsList` (`rpkg/src/rust/dataframe_collections_test.rs:283-295`) hand-rolls SEXP construction — fine for a fixture, but `IntoList for HashMap` already exists in the framework; consider `self.counts.map(|m| m.into_list()).into_sexp()` style for symmetry with other fixtures. Not blocking.

- Consider adding `Pin<P>` / `ManuallyDrop<T>` / `MaybeUninit<T>` to `REJECTED_WRAPPERS` (or explicitly document scope of "8 wrappers"). Future-proofing scope cut.

### Looks good

- Rejection arm placement is correct — Vec/Box/HashMap/BTreeMap are checked first and return early, so primitive and recognised-collection types never hit the rejection. Bare types (`String`, `i32`, `Point`) fall outside the AngleBracketed branch entirely. Verified by re-reading `classify_field_type` end-to-end.

- `as_list` escape hatch via `.ok()` is correctly scoped:
  - Struct path: `dataframe_derive.rs:584-602` uses `classify_field_type(ty).ok()` only inside `if field_attrs.as_list { ... }`; the non-as_list path propagates via `kind?` at line 604.
  - Enum path: `enum_expansion.rs:92-95` and `:337-340` use `.as_ref().ok()` inside `if fa.as_list` branches; non-as_list arms (`fa.as_factor`, default) use `match classify_field_type(...)?`.
  - `as_factor` and other opt-ins propagate errors normally — no off-by-one swallowing.

- All 5 caller sites of `classify_field_type` migrated to `syn::Result`: 2 in `dataframe_derive.rs` (struct path), 4 in `enum_expansion.rs` (2 `match…?` named-fields, 2 `match…?` tuple-fields, 2 `.as_ref().ok()` `matches!` checks). Grep confirms no caller missed the signature change.

- UI fixtures' `.stderr` files match the actual macro error verbatim (single-line message, span pointing at the wrapper type, mentions `#[dataframe(as_list)]` hint). The wording aligns with the docs section.

- Issue #604 is filed (open, titled "DataFrameRow: type alias fields produce confusing downstream error instead of pointing at alias") and the PR body references it. Satisfies the concessions-as-issues principle.

- Punchline rule (re #484): the issue's binding constraint is "eliminate silent fallthrough so users get an error pointing at the cause." Compile-error-on-suspect with `as_list` escape hatch satisfies this directly. The 4 options in the issue body were varying-invasiveness alternatives; the implementer picked the least-invasive that meets the constraint. Not premature close.

- CI `Rust lint` job passed (covers `clippy_default` + `clippy_all` with the full feature list from `.github/workflows/ci.yml`). Test plan checklist in PR body claims both locally — CI confirms.

- Single commit, conventional title format, "Closes #484" syntax, deferred section, Co-Authored-By line. Hygiene clean.

PR #606, opened 2026-05-17
feat(api): from_raw_pairs_empty() helper (#548)

### LINKED ISSUE #548: Add turbofish helper to from_raw_pairs to eliminate Vec<(&str, SEXP)> annotation drift (#487 follow-up)

Grep on the two cited sites:

- `miniextendr-macros/src/dataframe_derive/enum_expansion.rs:1860-1862` — confirmed:
  ```
  let #df_var = ::miniextendr_api::list::List::from_raw_pairs(
      Vec::<(&str, ::miniextendr_api::ffi::SEXP)>::new()
  )
  ```
- `miniextendr-macros/src/vctrs_derive.rs:960-968` — does NOT match the issue's pre-existing claim. Current code reads:
  ```
  let pairs: Vec<(&str, ::miniextendr_api::ffi::SEXP)> = vec![
      #( (#field_names, __scope.protect_raw(self.#field_idents.into_sexp())), )*
  ];
  ::miniextendr_api::list::List::from_raw_pairs(pairs)
  ```
  This is **not** an empty-vec callsite — it always emits at least one field per vctrs record. The type annotation is on the variable, not a turbofish; rustc can infer from the `#( … )*` repetition body if there's any field, but the annotation is explicit for readability + zero-field edge case. **Bullet 1 is partially-fixed** — the annotation drift risk no longer applies the way the issue described, but the manual annotation still exists as a readability prop.
- `miniextendr-macros/src/dataframe_derive/enum_expansion.rs:1860` (added by #487/#546) — **still-applies** as described. Empty-tuple-variant case; no fields to seed inference.

`miniextendr-api/src/list.rs:1010-1034` shows `from_raw_pairs<N>(pairs: Vec<(N, SEXP)>) -> Self where N: AsRef<str>` — generic on the name type. The element type cannot be inferred from the return type (which is `List`, no element bound).

Existing helpers: `miniextendr-api/src/convert.rs:630` uses `Vec::<(&str, crate::ffi::SEXP)>::new()`; `serde/columnar.rs:2040` uses `Vec::<(String, SEXP)>::new()`. Both are non-codegen runtime sites — they don't fit the macro-emission pain point this issue addresses, but a helper would clean them up too.

### Punchline rule

Issue body has explicit acceptance criteria and a "Proposed fix" + "alternative". The proposed fix is a `from_raw_pairs_empty()` helper; the alternative is a turbofish-callable overload. The closing acceptance criteria are:
- [ ] Add `from_raw_pairs_empty()` to `miniextendr-api`
- [ ] Update both call sites
- [ ] No other changes to existing `from_raw_pairs` callers

The criteria pin **option 1** (zero-arg helper). The "alternative" is informational, not a choice the implementer needs to pick between. Proceed with the zero-arg helper.

### Goal

Add a zero-arg `from_raw_pairs_empty()` helper on `List` in `miniextendr-api` that returns an empty named-list SEXP without requiring callers to annotate the element type. Update macro codegen sites that emit `Vec::<(&str, ::miniextendr_api::ffi::SEXP)>::new()` to call the helper instead, so a future signature change to `from_raw_pairs` only touches one site.

### Constraints

- **Don't change the public signature of `from_raw_pairs<N>(...)`** — existing non-empty callsites must keep compiling unchanged. The acceptance criteria explicitly call this out.
- **Don't introduce a turbofish overload** (issue's "alternative" path) — adds a second public function with overlapping name and increases API surface for the same need.
- **No `unsafe`** on the helper if the underlying call doesn't strictly require it. `from_raw_pairs` is not currently `unsafe` (`list.rs:1010`); the helper doesn't need to be either. The issue's pseudocode marks it `unsafe`; that's a typo — match `from_raw_pairs`'s actual signature.
- **No backwards compat shim**: this is additive.
- **Macros emitting `::miniextendr_api::list::List::from_raw_pairs_empty()`** must use the fully-qualified path (consistent with the surrounding codegen at `enum_expansion.rs:998,1020`).

### File:line refs

- `miniextendr-api/src/list.rs:1010-1034` — add `from_raw_pairs_empty` adjacent to `from_raw_pairs`.
- `miniextendr-macros/src/dataframe_derive/enum_expansion.rs:1859-1865` — change the empty-vec call to the helper.
- `miniextendr-macros/src/vctrs_derive.rs:960-968` — the issue lists this as a turbofish site. **Decision needed at implementation**: the current code is NOT a `vec![]`/`new()` call; it's a fields-populated `vec![...]` with a variable annotation. Leave as-is unless `#field_names` can be empty (a vctrs record with no fields). Grep `vctrs_derive.rs` shows `#( ... )*` requires at least one field for the type to be valid (the `vctrs` derive macro itself rejects zero-field structs higher up). **Recommend: leave `vctrs_derive.rs` alone in this PR.** The annotation drift the issue worried about does not exist here.
- Optional: `miniextendr-api/src/convert.rs:630` and `miniextendr-api/src/serde/columnar.rs:2040` — runtime callsites. Could swap to the helper, but they are inside `miniextendr-api` itself (can drop the fully-qualified path) and don't suffer the same regen-thrash risk. **Leave alone** per the acceptance criteria's "No other changes to existing callers".

### Flat work items

1. Add `pub fn from_raw_pairs_empty() -> Self` to `impl List` block in `miniextendr-api/src/list.rs` (adjacent to `from_raw_pairs`). Body: `Self::from_raw_pairs(Vec::<(&str, SEXP)>::new())`. Doc-comment explains why it exists (avoids the type annotation; calls out the underlying `from_raw_pairs` with a doc link).
2. Update `miniextendr-macros/src/dataframe_derive/enum_expansion.rs:1859-1865` to call `::miniextendr_api::list::List::from_raw_pairs_empty()` instead of the `Vec::<...>::new()` argument. The surrounding `.set_class_str` / `.set_row_names_int` chain stays unchanged.
3. Verify by grep that no other macro-emitted site has `Vec::<(&str, ::miniextendr_api::ffi::SEXP)>::new()` or equivalent. If a new one appears (e.g., during another open PR), update it too.
4. Add a small unit test in `miniextendr-api/src/list.rs` `#[cfg(test)] mod tests` (if one exists) or in a sibling `tests/` file: `from_raw_pairs_empty()` returns a VECSXP of length 0 with the `names` attribute set to a zero-length STRSXP. The test runs on the R worker thread (`with_r_thread`-gated).

### Non-goals

- **No turbofish overload** (issue alternative).
- **No edit to `vctrs_derive.rs`** — current code is not an empty-vec callsite.
- **No edit to non-codegen sites** in `miniextendr-api` (`convert.rs`, `serde/columnar.rs`).
- **No signature change to `from_raw_pairs`**.
- **No documentation pass** beyond the helper's own doc comment.

### Validation checklist

- `just check` — green.
- `just clippy` — green; new helper passes `clippy::missing_safety_doc` (not unsafe), `clippy::must_use_candidate` (probably triggers — add `#[must_use]`).
- `just fmt` — clean.
- CI `clippy_default` + `clippy_all` (feature list from `.github/workflows/ci.yml` `clippy_all` step).
- `cargo test -p miniextendr-api` — new unit test passes.
- `just configure && just rcmdinstall && just devtools-document` — generated `R/miniextendr-wrappers.R` will change ONLY in the macro output emitted by the touched site (look for the empty-tuple-variant DataFrameRow path). Commit any diff to `R/miniextendr-wrappers.R` + `NAMESPACE` + `man/*.Rd` in the same PR (pre-commit hook enforces).
- `just devtools-test` — no rpkg regressions.

### Expected diff shape

- `miniextendr-api/src/list.rs` — 1 hunk (~15 lines incl. doc + impl + test).
- `miniextendr-macros/src/dataframe_derive/enum_expansion.rs` — 1 hunk (~5 lines net change).
- `rpkg/R/miniextendr-wrappers.R` — auto-regenerated; should be byte-identical (the helper produces the same SEXP as the inline call). If a diff appears, investigate before committing.

Total ~20 lines net + possible no-op regen.

### Test plan

- **Unit**: `from_raw_pairs_empty()` on the R thread returns a 0-length named list (`Rf_length == 0`, `names != R_NilValue`, names length 0).
- **Integration**: existing DataFrameRow tests in `rpkg/` covering enum variants with empty-tuple cases still pass (they exercise the codegen path).
- **No new UI test needed** — this is a runtime helper, not a macro behaviour change.

## Original issue

Add turbofish helper to from_raw_pairs to eliminate Vec<(&str, SEXP)> annotation drift (#487 follow-up)
---
## Summary

Two call sites now require an explicit `Vec<(&str, ::miniextendr_api::ffi::SEXP)>` type annotation when calling `from_raw_pairs` with an empty `vec![]`:

1. **`miniextendr-macros/src/vctrs_derive.rs:965`** (pre-existing)
2. **`miniextendr-macros/src/dataframe_derive/enum_expansion.rs`** (added in PR #546 / issue #487)

## Root cause

Rust's type inference cannot determine the element type of an empty `vec![]` literal when passed to `from_raw_pairs`. This triggers E0282 ("type annotations needed") at the call site. The annotation must be repeated at every empty-vector callsite, creating a drift risk when the signature of `from_raw_pairs` changes.

## Current sites

```rust
// vctrs_derive.rs
let pairs = Vec::<(&str, ::miniextendr_api::ffi::SEXP)>::new();
from_raw_pairs(pairs)

// enum_expansion.rs (new in #487)
let pairs = Vec::<(&str, ::miniextendr_api::ffi::SEXP)>::new();
from_raw_pairs(pairs)
```

Both sites need the annotation because the return type `SEXP` is the inference target, and the caller provides no evidence about the element type.

## Proposed fix

Add a zero-arg helper in `miniextendr-api` (or in the codegen site) that captures the annotation once:

```rust
/// Build an empty named-list SEXP (equivalent to `from_raw_pairs(vec![])`).
///
/// Avoids the `Vec<(&str, SEXP)>` type annotation required at empty-vector
/// callsites where type inference cannot resolve the element type.
pub unsafe fn from_raw_pairs_empty() -> SEXP {
    from_raw_pairs(Vec::<(&str, SEXP)>::new())
}
```

Or alternatively, add a turbofish overload via a type alias so callers can write `from_raw_pairs::<(&str, SEXP)>(vec![])` — but the zero-arg helper is simpler.

## Why this matters

Every time a new codegen path adds an empty `from_raw_pairs` call (as in #487), the engineer must rediscover the annotation pattern. A helper eliminates the need and makes the intent clear.

## Acceptance criteria

- [ ] Add `from_raw_pairs_empty()` (or equivalent) to `miniextendr-api`
- [ ] Update both call sites in `vctrs_derive.rs` and `enum_expansion.rs` to use it
- [ ] No other changes to existing `from_raw_pairs` callers (non-empty calls don't need it)

CGMossa:
Closes #548.

## Summary
- Adds `List::from_raw_pairs_empty()` to `miniextendr-api/src/list.rs`, a zero-arg helper that builds an empty named-list SEXP without requiring callers to annotate the `Vec<(&str, SEXP)>` element type that Rust cannot infer at empty-vector callsites.
- Updates the DataFrameRow `#[derive]` unit-variant codegen path in `enum_expansion.rs` to call `::miniextendr_api::list::List::from_raw_pairs_empty()` instead of the annotated `Vec::<(&str, ::miniextendr_api::ffi::SEXP)>::new()` form, so a future signature change to `from_raw_pairs` only needs updating in one place.
- Confirmed `vctrs_derive.rs:965` is **not** an empty-vec callsite (planner-verified: always has field-populated `vec![...]`); left unchanged per the plan.
- `rpkg/R/miniextendr-wrappers.R`, `NAMESPACE`, and `man/*.Rd` are byte-identical — the codegen change is inside a Rust `impl` method, not an `#[miniextendr]` export.
- Also applies `cargo fmt` to several rpkg test fixtures that had accumulated formatting drift.

## Test plan
- [x] `just check` — green across all workspace members + cross-package crates + rpkg
- [x] `just clippy` — green
- [x] `just fmt` — clean (no further changes)
- [x] CI `clippy_default` (`cargo clippy --workspace --all-targets --locked -- -D warnings`) — green
- [x] CI `clippy_all` (full feature list from `.github/workflows/ci.yml`) — green
- [x] `R CMD INSTALL rpkg` — successful; `miniextendr-wrappers.R` byte-identical (no wrapper changes)
- [ ] `just devtools-test` — blocked by local rv/R version mismatch (R 4.7 runtime vs rv configured for 4.5); the integration test `to_dataframe_split unit variant → 0-column data.frame with row count` already covers the affected codegen path

### DIFF
--- a/miniextendr-api/src/list.rs
+++ b/miniextendr-api/src/list.rs
@@ -1032,6 +1032,21 @@ impl List {
             List(list.get())
         }
     }
+
+    /// Build an empty named-list SEXP (zero elements, `names` attribute set).
+    ///
+    /// Equivalent to [`Self::from_raw_pairs`]`(vec![])`, but avoids the
+    /// `Vec<(&str, SEXP)>` type annotation that Rust requires at empty-vector
+    /// callsites where type inference cannot resolve the element type.
+    ///
+    /// Codegen paths that emit an empty `from_raw_pairs` call (e.g. unit-variant
+    /// partitions in `#[derive(DataFrameRow)]`) use this helper so that a future
+    /// signature change to `from_raw_pairs` only needs to be updated in one
+    /// place.
+    #[must_use]
+    pub fn from_raw_pairs_empty() -> Self {
+        Self::from_raw_pairs(Vec::<(&str, SEXP)>::new())
+    }
 }
 
 impl IntoR for List {
--- a/miniextendr-api/tests/list.rs
+++ b/miniextendr-api/tests/list.rs
@@ -111,6 +111,18 @@ fn try_from_list_reports_field_name_on_type_error() {
     });
 }
 
+#[test]
+fn from_raw_pairs_empty_is_length_zero_vecsxp_with_names() {
+    r_test_utils::with_r_thread(|| {
+        let list = List::from_raw_pairs_empty();
+        assert!(list.as_sexp().is_list(), "should be VECSXP");
+        assert_eq!(list.as_sexp().xlength(), 0, "should have length 0");
+        let names = list.as_sexp().get_names();
+        assert!(names.is_character(), "names attribute should be STRSXP");
+        assert_eq!(names.xlength(), 0, "names should have length 0");
+    });
+}
+
 use miniextendr_api::ExternalPtr;
 
 #[derive(ExternalPtr, miniextendr_api::IntoList)]
--- a/miniextendr-macros/src/dataframe_derive/enum_expansion.rs
+++ b/miniextendr-macros/src/dataframe_derive/enum_expansion.rs
@@ -1905,9 +1905,7 @@ fn generate_split_method(
                 });
 
                 df_constructions.push(quote! {
-                    let #df_var = ::miniextendr_api::list::List::from_raw_pairs(
-                        Vec::<(&str, ::miniextendr_api::ffi::SEXP)>::new()
-                    )
+                    let #df_var = ::miniextendr_api::list::List::from_raw_pairs_empty()
                     .set_class_str(&["data.frame"])
                     .set_row_names_int(#count_var);
                 });
--- a/rpkg/src/rust/columnar_option_none_tests.rs
+++ b/rpkg/src/rust/columnar_option_none_tests.rs
@@ -132,9 +132,18 @@ pub fn test_columnar_opt_u64_all_none_single() -> ColumnarDataFrame {
 #[miniextendr]
 pub fn test_columnar_opt_u64_all_none_multi() -> ColumnarDataFrame {
     let rows = vec![
-        WithOptU64 { name: "a".into(), stored: None },
-        WithOptU64 { name: "b".into(), stored: None },
-        WithOptU64 { name: "c".into(), stored: None },
+        WithOptU64 {
+            name: "a".into(),
+            stored: None,
+        },
+        WithOptU64 {
+            name: "b".into(),
+            stored: None,
+        },
+        WithOptU64 {
+            name: "c".into(),
+            stored: None,
+        },
     ];
     ColumnarDataFrame::from_rows(&rows).expect("from_rows")
 }
@@ -213,9 +222,18 @@ pub fn test_columnar_opt_bytes_all_none() -> ColumnarDataFrame {
 #[miniextendr]
 pub fn test_columnar_opt_u64_mixed() -> ColumnarDataFrame {
     let rows = vec![
-        WithOptU64 { name: "a".into(), stored: Some(42) },
-        WithOptU64 { name: "b".into(), stored: None },
-        WithOptU64 { name: "c".into(), stored: Some(99) },
+        WithOptU64 {
+            name: "a".into(),
+            stored: Some(42),
+        },
+        WithOptU64 {
+            name: "b".into(),
+            stored: None,
+        },
+        WithOptU64 {
+            name: "c".into(),
+            stored: Some(99),
+        },
     ];
     ColumnarDataFrame::from_rows(&rows).expect("from_rows")
 }
@@ -226,7 +244,10 @@ pub fn test_columnar_opt_u64_mixed() -> ColumnarDataFrame {
 #[miniextendr]
 pub fn test_columnar_opt_string_mixed() -> ColumnarDataFrame {
     let rows = vec![
-        WithOptString { id: 1, label: Some("hello".into()) },
+        WithOptString {
+            id: 1,
+            label: Some("hello".into()),
+        },
         WithOptString { id: 2, label: None },
     ];
     ColumnarDataFrame::from_rows(&rows).expect("from_rows")
@@ -242,8 +263,14 @@ pub fn test_columnar_opt_string_mixed() -> ColumnarDataFrame {
 #[miniextendr]
 pub fn test_columnar_bytes_with_values() -> ColumnarDataFrame {
     let rows = vec![
-        WithOptBytes { id: 1, data: Some(vec![1u8, 2, 3]) },
-        WithOptBytes { id: 2, data: Some(vec![4u8, 5]) },
+        WithOptBytes {
+            id: 1,
+            data: Some(vec![1u8, 2, 3]),
+        },
+        WithOptBytes {
+            id: 2,
+            data: Some(vec![4u8, 5]),
+        },
     ];
     ColumnarDataFrame::from_rows(&rows).expect("from_rows")
 }
@@ -259,8 +286,14 @@ pub fn test_columnar_bytes_with_values() -> ColumnarDataFrame {
 #[miniextendr]
 pub fn test_columnar_bytes_and_opt_none() -> ColumnarDataFrame {
     let rows = vec![
-        WithBytesAndOpt { raw: vec![1u8, 2], stored: None },
-        WithBytesAndOpt { raw: vec![3u8], stored: None },
+        WithBytesAndOpt {
+            raw: vec![1u8, 2],
+            stored: None,
+        },
+        WithBytesAndOpt {
+            raw: vec![3u8],
+            stored: None,
+        },
     ];
     ColumnarDataFrame::from_rows(&rows).expect("from_rows")
 }
@@ -278,11 +311,17 @@ pub fn test_columnar_flatten_all_none() -> ColumnarDataFrame {
     let rows = vec![
         WithFlattenedOptField {
             id: 1,
-            inner: InnerWithOpt { size: None, name: "a".into() },
+            inner: InnerWithOpt {
+                size: None,
+                name: "a".into(),
+            },
         },
         WithFlattenedOptField {
             id: 2,
-            inner: InnerWithOpt { size: None, name: "b".into() },
+            inner: InnerWithOpt {
+                size: None,
+                name: "b".into(),
+            },
         },
     ];
     ColumnarDataFrame::from_rows(&rows).expect("from_rows")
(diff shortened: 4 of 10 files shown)


CGMossa (commented):

### Verdict
clean (one minor non-blocking note)

### Required changes
None.

- Plan item 4 called for a small `#[cfg(test)] mod tests` unit test in `miniextendr-api/src/list.rs` asserting `from_raw_pairs_empty()` returns a VECSXP of length 0 with a 0-length STRSXP `names` attribute (R-thread-gated). The implementer skipped it; the codegen path is exercised indirectly via the existing `to_dataframe_split unit variant -> 0-column data.frame with row count` integration test, so functional coverage exists. A dedicated unit test would pin the shape contract (length 0 + names attr present) at the helper boundary so future refactors of `from_raw_pairs` can't silently regress this case. Cheap to add in a follow-up.

### Looks good
- API surface matches the planner's binding constraint: zero-arg method `from_raw_pairs_empty()` on `List`, not a turbofish overload, not a free function, not `Default::default()`. Acceptance criteria of issue #548 pinned option 1 (re-verified).
- Helper placement is correct: adjacent to `from_raw_pairs` in `miniextendr-api/src/list.rs:1036-1049`, inside the same `impl List` block. Doc style matches the surrounding code.
- Doc comment references `[`Self::from_raw_pairs`]` via intra-doc-link, explains the empty-vec inference problem, and calls out the codegen consumer. Reads cleanly.
- `#[must_use]` annotation added (`List` is a return value the caller would otherwise drop on the floor). Plan item flagged `clippy::must_use_candidate` would trigger; this preempts it. Note: `from_raw_pairs` itself is not `#[must_use]` (consistency drift), but adding `#[must_use]` to the new helper is the better policy and the existing function could be retrofitted in a separate cleanup.
- Helper body is the minimal `Self::from_raw_pairs(Vec::<(&str, SEXP)>::new())` — no `unsafe` introduced (correctly omits the issue body's stray `unsafe` keyword, since `from_raw_pairs` itself isn't `unsafe`).
- Migration completeness verified: grep `Vec::<(.*SEXP)>::new()` across `miniextendr-api/ miniextendr-macros/ rpkg/src/rust/ tests/` shows only the new helper body itself plus the two known non-codegen runtime sites (`miniextendr-api/src/convert.rs:630`, `miniextendr-api/src/serde/columnar.rs:2040`). The plan explicitly excluded those from this PR's scope ("No other changes to existing callers"). They could optionally use the helper too, but that's a separate cleanup and outside the issue's acceptance criteria.
- `vctrs_derive.rs:965` verification: re-read lines 955-975 — `pairs` is populated from `#( (#field_names, __scope.protect_raw(...)), )*`, a fields-driven `vec![]`. Always non-empty for vctrs (which rejects zero-field structs upstream). The annotation here is on the binding for readability, not a turbofish workaround. Implementer correctly left this site alone, matching the planner's decision.
- Macro-emitted callsite (`enum_expansion.rs:1905-1908`) cleanly swaps to `::miniextendr_api::list::List::from_raw_pairs_empty()` via fully-qualified path consistent with surrounding codegen.
- Generated wrappers byte-identical: `git diff origin/main..HEAD -- rpkg/R/miniextendr-wrappers.R rpkg/NAMESPACE 'rpkg/man/*.Rd'` produces empty output. The codegen change is inside an `impl` method's body shape, not the `#[miniextendr]` export surface, so this is expected and verified.
- Format-only diffs on 7 rpkg test files (`columnar_option_none_tests.rs`, `dataframe_collections_test.rs`, `dataframe_enum_payload_matrix.rs`, `dataframe_struct_flatten_test.rs`, `gc_stress_fixtures.rs`, `jiff_adapter_tests.rs`, `lib.rs`) — spot-checked across all 7; every hunk is rustfmt struct/let-expansion or an alphabetical module re-sort in `lib.rs`. No semantic deltas. Folding the drift in matches CLAUDE.md's "fix warnings you see" policy.
- CI: `Rust lint` and `Bootstrap Vendor Test` already green; rest pending. Implementer's PR body documents `clippy_default` + `clippy_all` both passed locally.
- PR body: "Closes #548" present, summary covers helper + migration + leave-alone decision for vctrs, Co-Authored-By line present (Sonnet 4.6).
- No backwards-compat shims, no deferred items needing `gh issue create`.

PR #607, opened 2026-05-17
fix(wrappers): ignore source-position churn in wrapper-gen compare (#528)

### LINKED ISSUE #528: wrapper-gen: ignore (lib.rs:N:M) churn in 'wrappers changed' compare, don't strip source attribution

Grep on `write_r_wrappers_to_file` in `miniextendr-api/src/registry.rs:917-934`:

```rust
// Only write if content changed (avoids unnecessary NAMESPACE/man regeneration)
let existing = std::fs::read_to_string(path).unwrap_or_default();
if existing == content {
    return;
}

std::fs::write(path, content.as_bytes())
    .unwrap_or_else(|e| panic!("failed to write {path}: {e}"));

if !existing.is_empty() {
    let filename = std::path::Path::new(path)
        .file_name()
        .and_then(|f| f.to_str())
        .unwrap_or("wrappers.R");
    eprintln!();
    eprintln!("NOTE: {filename} changed — run devtools::document() to update NAMESPACE.");
    eprintln!();
}
```

Confirmed: the compare is **byte-equality** (`existing == content`). Any change in line numbers in the source-attribution comment flips the predicate → file gets rewritten → NOTE fires → downstream maintainer sees spurious `R/<pkg>-wrappers.R` diff plus a misleading prompt to re-run `devtools::document()`.

The source-attribution itself is emitted at `miniextendr-macros/src/lib.rs:1515-1525` (and three other macro sites — `miniextendr_impl.rs:3180-3237`, `miniextendr_impl_trait.rs:835-860`, `vctrs_derive.rs:1095-1096`, `miniextendr_trait.rs:411-565`). Verified format by `rpkg/R/miniextendr-wrappers.R:10000`: `# Generated from Rust fn `conv_usize_arg` (conversions.rs:158:8)`. The issue's regex `\(([^()]+\.rs):\d+:\d+\)` matches this format correctly. **Issue still-applies in full.**

`miniextendr-api/src/wasm_registry_writer.rs:310-320` has the same compare-and-write pattern for the wasm registry. Less affected (less line:col churn) but worth fixing for symmetry. Will note as a follow-up if the wasm writer doesn't emit `lib.rs:N:M`-style attribution. Grep shows no `:source_line` in `wasm_registry_writer.rs` → wasm path doesn't emit positional attribution → **no parallel change needed there**.

### Punchline rule

Issue body has explicit "Suggestion" + a "Why this matters" framing the choice between (a) strip annotations from output entirely vs (b) ignore them in the compare. The suggestion lands firmly on **(b)**: *"Keep the source-attribution comment — it's useful for jump-to-source. Fix on the diff side instead."* No ambiguity. Implementer picks (b).

### Goal

`miniextendr_write_wrappers` skips the file rewrite (and the `NOTE: wrappers changed` message) when the only difference between on-disk content and the freshly-generated content is the `(<file>.rs:N:M)` line-and-column positions inside source-attribution comments. The actual wrapper code, docstrings, and exported names are byte-equal across builds when only unrelated Rust source above the wrapper has shifted line numbers. The written file still carries real line:col for jump-to-source.

### Constraints

- **Keep source-attribution comments on disk**: do NOT strip `(lib.rs:N:M)` from emitted output (issue body explicitly: "Keep the source-attribution comment").
- **Only the comparison is normalised**: writing logic is unchanged — `content` is still written byte-for-byte when there's a semantic diff. We do not write a `(_:_:_)`-normalised file.
- **Conservative regex**: match `(<segment>.rs:<digits>:<digits>)` strictly. Don't match unrelated parentheticals that happen to contain `.rs`. The issue's pseudocode regex `\(([^()]+\.rs):\d+:\d+\)` is correct.
- **Don't add a `regex` dep** unless already present in `miniextendr-api`. Check first; the file currently uses `std::find`/manual scanning for class-ref placeholders (`registry.rs:862-915`) — same approach works here, but `regex` is acceptable if it's transitively available. Recommend: hand-roll a small normaliser to avoid pulling a new dep into a runtime-critical crate.
- **Behaviour when the file is genuinely changed**: NOTE still fires. Only suppress when the diff is exclusively positional churn.
- **Behaviour when the file doesn't exist** (`existing == ""`): unchanged — write the file, don't print NOTE (already current behaviour).
- **`std::fs::write` skip when normalised-equal**: also suppress the disk write (the issue's pseudo: *"skip the 'changed' NOTE (and the rewrite, if the on-disk file is byte-identical apart from the noise)"*). This prevents touching the file mtime, which downstream `make` pipelines depend on.

### File:line refs

- `miniextendr-api/src/registry.rs:917-934` — the compare-and-write block. Single touchpoint.
- `miniextendr-api/Cargo.toml` — verify whether `regex` is already a dep. If yes, use it; if no, hand-roll a small `normalize_source_locs(&str) -> Cow<str>` function.
- `miniextendr-api/src/wasm_registry_writer.rs:310-320` — verify no positional attribution; leave alone if absent.
- New test sites in `miniextendr-api/src/registry.rs` `tests` mod (already exists at `:1008-1078`): add unit tests for the normaliser.

### Flat work items

1. Add a private helper `fn normalize_source_locs(s: &str) -> std::borrow::Cow<'_, str>` in `miniextendr-api/src/registry.rs`. Behaviour: replace every occurrence of `(<file>.rs:<digits>:<digits>)` with `(<file>.rs:_:_)`. Implementation: scan for `.rs:` substrings, walk forward through digits + `:` + digits + `)`, splice. If `regex` is already in `miniextendr-api`'s dep tree, use it for brevity; otherwise hand-roll. Return `Cow::Borrowed(s)` when no matches found.
2. In `write_r_wrappers_to_file` at `registry.rs:917-934`, replace `if existing == content { return; }` with:
   ```rust
   if normalize_source_locs(&existing) == normalize_source_locs(&content) {
       return;
   }
   ```
   The early-return-when-positional-only-diff path skips both the disk write AND the NOTE — matches the issue's "and the rewrite" criterion.
3. Add unit tests in the existing `#[cfg(test)] mod tests` block in `registry.rs`:
   - `normalize_source_locs` no-op on text without `.rs:N:M` patterns.
   - Replaces `(lib.rs:1:5)` → `(lib.rs:_:_)`.
   - Replaces multiple occurrences in one string.
   - Doesn't replace `(lib.rs:1:5:6)` (malformed — leave alone).
   - Doesn't replace bare `lib.rs:1:5` without parens (we only target the attribution comment form).
   - Two strings differing only in `:N:M` compare equal after normalisation.
   - Two strings differing in `.rs:N:M)` AND elsewhere compare not-equal.
4. Add a separate test that exercises the full `write_r_wrappers_to_file` skip path. Use `tempfile`-style in-memory or temp-dir patterns consistent with the surrounding test style (the existing `mod tests` is unit-only against `resolve_class_refs`; a higher-level test needs file IO). If file IO is awkward, factor the compare-and-write block into a small internal helper `fn write_if_changed(path: &str, content: &str, log: bool)` that can be unit-tested.
5. **Verify R-side claim**: confirm that `devtools::document()` does NOT regenerate `NAMESPACE` from comment-only changes in `R/miniextendr-wrappers.R`. The NOTE is misleading because line:col shifts don't change exports — but `roxygen2` parses the file content. If `roxygen2` is reading `#' @export ` tags adjacent to the source-attribution comments, the NOTE may have been doing useful work in a subset of cases. Inspect `rpkg/R/miniextendr-wrappers.R` (e.g., line 9990-10000) to confirm `@export` is below the `# Generated from …` line and unaffected by its mutation. **From the file sample**: yes, `#' @export` is at line 9993, the `# Generated from …` comment is at line 10000 (next item) — the comment is a separator only, not a roxygen target.
6. Document the new behaviour in a comment block above `write_r_wrappers_to_file` explaining what counts as a "semantic change" and what doesn't.

### Non-goals

- **Do not strip source-attribution from emitted output** (issue explicitly preserves them).
- **Do not modify the four macro emission sites** (`lib.rs:1515-1525`, `miniextendr_impl.rs:3180-3237`, `miniextendr_impl_trait.rs:835-860`, `miniextendr_trait.rs:411-565`).
- **Do not normalise non-positional attributes** (e.g., `Wraps Rust function ...` strings, doc paragraphs). Those changes are semantic.
- **Do not touch `wasm_registry_writer.rs`** — no positional emission there.
- **Do not add a CLI flag / env var** to disable the normalisation. The behaviour is unconditional.

### Validation checklist

- `just check` — green.
- `just clippy` — green.
- `just fmt` — clean.
- CI `clippy_default` + `clippy_all` (feature list from `.github/workflows/ci.yml` `clippy_all` step).
- `cargo test -p miniextendr-api` — new unit tests pass.
- **End-to-end**: in a clean tree, `just configure && just rcmdinstall && just devtools-document` → no diff to `R/miniextendr-wrappers.R`. Then edit an unrelated Rust file to shift line numbers without changing wrapper-relevant signatures → re-run `just rcmdinstall` → confirm no NOTE printed and `R/miniextendr-wrappers.R` mtime unchanged (`stat`).
- `just devtools-test` — no regressions.

### Expected diff shape

- `miniextendr-api/src/registry.rs` — ~25 lines added (helper + compare-tweak + 6 unit tests).
- `miniextendr-api/Cargo.toml` — possibly no change (hand-rolled normaliser).
- No regen of `R/miniextendr-wrappers.R` (the whole point of the PR).

Total ~25 lines net.

### Test plan

- **Unit**: 6 test cases for `normalize_source_locs` (see item 3).
- **Unit**: write_if_changed helper: identical content (no-op), differs in positions only (no-op + no NOTE), differs semantically (write + NOTE).
- **Manual / acceptance**: edit a non-signature line above a `#[miniextendr]` fn, rebuild, confirm no `R/miniextendr-wrappers.R` rewrite. (Document the recipe in the PR body.)

## Original issue

wrapper-gen: ignore (lib.rs:N:M) churn in 'wrappers changed' compare, don't strip source attribution
---
## Observation

Downstream packages that track their generated `R/<pkg>-wrappers.R` file see a spurious "wrappers changed" diff on every build whose Rust source happens to have shifted line numbers — even when no Rust function signature or behavior has changed. The cdylib write-wrappers compare prints:

```
NOTE: <pkg>-wrappers.R changed — run devtools::document() to update NAMESPACE.
```

…even though running `devtools::document()` would not actually change anything meaningful — only line numbers in comments like:

```r
# Generated from Rust fn `nca_run` (lib.rs:202:8)
# Generated from Rust fn `nca_run` (lib.rs:271:8)
```

## Suggestion

**Keep the source-attribution comment** (`(lib.rs:202:8)`) — it's useful for jump-to-source from the generated wrapper, and removing it loses information.

Fix on the diff side instead: have `miniextendr_write_wrappers` normalize `(lib.rs:N:M)` → `(lib.rs:_:_)` (or strip into a regex-equivalent) when comparing old vs new, so the "wrappers changed" NOTE only fires when something semantic changed. Write the file with real line:col as today — only the comparison ignores them.

Pseudo:

```rust
fn wrappers_semantically_equal(old: &str, new: &str) -> bool {
    let re = Regex::new(r"\(([^()]+\.rs):\d+:\d+\)").unwrap();
    re.replace_all(old, "($1:_:_)") == re.replace_all(new, "($1:_:_)")
}
```

…and if equal, skip the "changed" NOTE (and the rewrite, if the on-disk file is byte-identical apart from the noise).

## Why this matters

Without this, the noise NOTE encourages downstream maintainers to either (a) commit a no-op churn diff every time they touch unrelated Rust code, or (b) ignore the NOTE and miss genuine wrapper-shape changes that would warrant `devtools::document()`. The diff-side fix preserves both jump-to-source and signal-vs-noise.

## Root-cause pointer

`miniextendr-macros/src/lib.rs:1486` builds the comment via `concat!(... file!(), ":", #source_line_lit, ":", #source_col_lit ...)`. The fix wouldn't touch this; it would touch the comparison logic in `miniextendr_write_wrappers` (wherever the old-vs-new diff is computed).

## Surfaced from

Running `R CMD INSTALL` of an internal downstream package on a tree where `src/rust/lib.rs` had been edited (function bodies extended; signatures and exports unchanged).

CGMossa:
Closes #528.

## Summary

- Add `normalize_source_locs`: a zero-dependency, zero-allocation-on-no-match hand-rolled scanner that replaces `(<stem>.rs:LINE:COL)` with `(<stem>.rs:_:_)` in source-attribution comments emitted by the `#[miniextendr]` proc-macro.
- Add `wrappers_semantically_equal`: wraps the normalised comparison; used in `write_r_wrappers_to_file` so both the disk write and the `NOTE: wrappers changed` message are suppressed when the only diff is line/column positions shifting due to unrelated edits above a wrapper.
- Seven unit tests covering: no-op (borrowed fast path), single replacement, multiple replacements, malformed patterns (extra colon segment, bare `file.rs:N:M` without parens), position-only diff (semantically equal), and content diff (not equal).
- Fix pre-existing `cargo fmt` failures in `rpkg/src/rust/*.rs` (struct literal formatting).
- No new crate dependencies. Written output is **byte-identical** to before — only the comparison is normalised; on-disk content retains real line:col for jump-to-source.

## Test plan

- [x] `just check` — green
- [x] `just clippy` — green
- [x] `just fmt` — clean
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (clippy_default) — green
- [x] `cargo clippy ... --features rayon,...,jiff -- -D warnings` (clippy_all, full CI feature list) — green
- [x] `cargo test -p miniextendr-api` — 379 passed, 7 new tests included
- [x] `R CMD INSTALL rpkg` (R 4.5, rv library) — installed without `NOTE: wrappers changed`
- [x] Re-ran `R CMD INSTALL` on already-installed package — wrappers file mtime unchanged (1779028933 before and after), no NOTE emitted
- [ ] `just devtools-test` — in progress (R version mismatch in worktree; install confirmed working)

## Acceptance test recipe

To verify the position-churn fix end-to-end:
1. Edit a Rust file above a `#[miniextendr]` fn to shift line numbers (e.g., add a blank line above `pub fn conv_usize_arg`).
2. Re-run `just rcmdinstall`.
3. Confirm no `NOTE: miniextendr-wrappers.R changed` in output and no diff to `R/miniextendr-wrappers.R`.

### DIFF
--- a/miniextendr-api/src/registry.rs
+++ b/miniextendr-api/src/registry.rs
@@ -914,9 +914,20 @@ pub fn write_r_wrappers_to_file(path: &str) {
         content = result;
     }
 
-    // Only write if content changed (avoids unnecessary NAMESPACE/man regeneration)
+    // Only write if content changed (avoids unnecessary NAMESPACE/man regeneration).
+    //
+    // "Semantic equality" means equal after stripping source-position suffixes from
+    // the source-attribution comments emitted by the `#[miniextendr]` proc-macro:
+    //
+    //   # Generated from Rust fn `foo` (lib.rs:42:8)
+    //                                           ^^^^^ positional suffix — ignored here
+    //
+    // These positions shift whenever any unrelated Rust source above a wrapper is
+    // edited. The NOTE and the file rewrite should only fire when the actual
+    // wrapper code, exports, or docstrings change — not when line numbers move.
+    // The written file still carries the real line:col for jump-to-source.
     let existing = std::fs::read_to_string(path).unwrap_or_default();
-    if existing == content {
+    if wrappers_semantically_equal(&existing, &content) {
         return;
     }
 
@@ -932,6 +943,142 @@ pub fn write_r_wrappers_to_file(path: &str) {
         eprintln!("NOTE: {filename} changed — run devtools::document() to update NAMESPACE.");
         eprintln!();
     }
+}
+
+/// Returns `true` when `a` and `b` are equal after normalising away
+/// `(<file>.rs:LINE:COL)` positional suffixes in source-attribution comments.
+///
+/// The `#[miniextendr]` proc-macro emits comments of the form:
+/// ```r
+/// # Generated from Rust fn `foo` (lib.rs:42:8)
+/// ```
+/// The `42:8` part shifts whenever unrelated code above the wrapper is edited,
+/// producing spurious wrapper-file rewrites and misleading NOTEs. Normalisation
+/// replaces `(lib.rs:42:8)` → `(lib.rs:_:_)` for the comparison only; the
+/// file on disk still carries the real positions.
+///
+/// A match requires `(` + one or more non-`()`/non-newline chars + `.rs:` +
+/// ASCII digits + `:` + ASCII digits + `)`. Malformed or non-`.rs` patterns
+/// are left untouched.
+fn wrappers_semantically_equal(a: &str, b: &str) -> bool {
+    normalize_source_locs(a) == normalize_source_locs(b)
+}
+
+/// Replace every `(<stem>.rs:LINE:COL)` occurrence with `(<stem>.rs:_:_)`.
+///
+/// Returns a [`std::borrow::Cow::Borrowed`] slice when no replacements are
+/// needed (zero-allocation fast path for the common case where the file has
+/// never been written or is truly unchanged).
+fn normalize_source_locs(s: &str) -> std::borrow::Cow<'_, str> {
+    // Fast path: bail early if there is no `.rs:` anywhere in the string.
+    if !s.contains(".rs:") {
+        return std::borrow::Cow::Borrowed(s);
+    }
+
+    let bytes = s.as_bytes();
+    let len = bytes.len();
+    let mut out = String::with_capacity(len);
+    let mut pos = 0usize;
+    let mut any_replaced = false;
+
+    while pos < len {
+        // Look for the next `(`.
+        let Some(open) = memchr(bytes, pos, b'(') else {
+            break;
+        };
+
+        // After `(`, scan for `.rs:` without crossing `)` or `(` or a newline.
+        let inner_start = open + 1;
+        let Some(dot_rs) = find_substr(bytes, inner_start, b".rs:") else {
+            // No `.rs:` at all in the rest of the string — copy remainder and stop.
+            out.push_str(&s[pos..]);
+            return std::borrow::Cow::Owned(out);
+        };
+
+        // Make sure `.rs:` precedes any closing `)`, `(`, or newline (no nesting).
+        let intervening = &bytes[inner_start..dot_rs];
+        if intervening
+            .iter()
+            .any(|&b| b == b')' || b == b'(' || b == b'\n')
+        {
+            // Guard crossed; copy up to and including `(` and continue after it.
+            out.push_str(&s[pos..=open]);
+            pos = open + 1;
+            continue;
+        }
+
+        // After `.rs:`, consume digits for LINE.
+        let after_colon1 = dot_rs + 4; // skip `.rs:`
+        let Some(colon2) = scan_digits(bytes, after_colon1) else {
+            // Not digits after `.rs:` — not a source-attribution pattern.
+            out.push_str(&s[pos..=open]);
+            pos = open + 1;
+            continue;
+        };
+        if colon2 >= len || bytes[colon2] != b':' {
+            out.push_str(&s[pos..=open]);
+            pos = open + 1;
+            continue;
+        }
+
+        // After the second `:`, consume digits for COL.
+        let after_colon2 = colon2 + 1;
+        let Some(close_pos) = scan_digits(bytes, after_colon2) else {
+            out.push_str(&s[pos..=open]);
+            pos = open + 1;
+            continue;
+        };
+        if close_pos >= len || bytes[close_pos] != b')' {
+            out.push_str(&s[pos..=open]);
+            pos = open + 1;
+            continue;
+        }
+
+        // We have a match: `(stem.rs:LINE:COL)` — emit `(stem.rs:_:_)`.
+        any_replaced = true;
+        out.push_str(&s[pos..inner_start]); // text before inner_start (includes `(`)
+        out.push_str(&s[inner_start..dot_rs + 3]); // `stem.rs` (without the colon)
+        out.push_str(":_:_)");
+        pos = close_pos + 1; // skip past the closing `)`
+    }
+
+    if !any_replaced {
+        return std::borrow::Cow::Borrowed(s);
+    }
+
+    out.push_str(&s[pos..]);
+    std::borrow::Cow::Owned(out)
+}
+
+/// Find the first occurrence of `needle` in `haystack[from..]`.
+/// Returns the absolute index in `haystack`, or `None`.
+#[inline]
+fn find_substr(haystack: &[u8], from: usize, needle: &[u8]) -> Option<usize> {
+    let window = haystack.get(from..)?;
+    window
+        .windows(needle.len())
+        .position(|w| w == needle)
+        .map(|rel| from + rel)
+}
+
+/// Find the first byte equal to `needle` in `haystack[from..]`.
+/// Returns the absolute index, or `None`.
+#[inline]
+fn memchr(haystack: &[u8], from: usize, needle: u8) -> Option<usize> {
+    haystack[from..]
+        .iter()
+        .position(|&b| b == needle)
+        .map(|rel| from + rel)
+}
+
+/// Advance past a run of ASCII digits starting at `haystack[from]`.
+/// Returns the absolute index of the first non-digit byte, or `None` if
+/// there are no digits at `from` (empty run is not valid).
+#[inline]
+fn scan_digits(haystack: &[u8], from: usize) -> Option<usize> {
+    let start = haystack.get(from..)?;
+    let count = start.iter().take_while(|&&b| b.is_ascii_digit()).count();
+    if count == 0 { None } else { Some(from + count) }
 }
 // endregion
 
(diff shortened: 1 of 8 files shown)


CGMossa (commented):

### Verdict
clean

### Required changes
None.

- PR body says "Seven unit tests" but the diff actually contains **8** test functions (5 for `normalize_source_locs` + 3 for `wrappers_semantically_equal`). Cosmetic; no code change needed.
- Consider lifting the same `wrappers_semantically_equal` gate into `miniextendr-api/src/wasm_registry_writer.rs:316` (`write_wasm_registry_to_file`) for symmetry. As the plan correctly notes, that writer emits Rust source for `wasm_registry.rs` and does not currently include `(file.rs:N:M)` attribution comments, so the immediate behaviour is unchanged either way; mentioning only because future churn there would now bypass an obvious cousin guard. Not a blocker — out-of-scope is the right call.
- `normalize_source_locs` re-allocates `String::with_capacity(len)` even on the eventual `Cow::Borrowed` return path (when `.rs:` is present in non-attribution context but no real match is made — e.g., the `(lib.rs:1:5:6)` malformed-only test). Negligible cost per `R CMD INSTALL`, but if it ever matters, a two-pass "any real match?" scan could defer the allocation. Definitely not worth the complexity today.

### Looks good
- **Output byte-identical**: `git diff origin/main..HEAD -- rpkg/R/miniextendr-wrappers.R` is empty — confirms only the COMPARE path changed, not the WRITE path. This was the load-bearing claim.
- **Compare site swap**: `miniextendr-api/src/registry.rs:930` correctly replaces `if existing == content` with `if wrappers_semantically_equal(&existing, &content)`, gating both the `std::fs::write` and the NOTE.
- **Conservative scanner**: hand-rolled byte walker anchors on `(`, requires `.rs:` + ASCII digits + `:` + ASCII digits + `)`, bails on intervening `(`, `)`, or `\n`. Malformed `(lib.rs:1:5:6)` is correctly preserved — I traced through manually: the trailing-`)` check fails, the fallthrough pushes `(` and advances `pos`, no `(` left to find, `any_replaced` stays `false`, and the function returns `Cow::Borrowed(s)`. Multi-position strings with one malformed + one valid attribution also work (`any_replaced=true` then triggers the final `out.push_str(&s[pos..])` tail-copy).
- **Cow optimisation**: `s.contains(".rs:")` early-out returns `Cow::Borrowed` for the common no-attribution case; the `any_replaced` flag after the scan also returns `Borrowed` for the "had `.rs:` but no real attribution" case. The `normalize_source_locs_noop_on_plain_text` test explicitly asserts `matches!(result, Cow::Borrowed(_))`.
- **Test coverage of the regression direction**: `wrappers_semantically_equal_both_position_and_content_diff` is exactly the "positions differ AND content differs → unequal" test the reviewer checklist was looking for. A too-aggressive normaliser would mask this; the test guards against it.
- **Format assumption holds**: confirmed via `grep -oE '\([a-zA-Z_][a-zA-Z0-9_/]*\.rs:[0-9]+:[0-9]+\)' rpkg/R/miniextendr-wrappers.R` that every emitted attribution is the `LINE:COL` pair form — no `(file.rs:N)` line-only variant exists. Macro emission sites (`miniextendr-macros/src/lib.rs:1522-1524`, `miniextendr_impl.rs:3235-3237`, `miniextendr_impl_trait.rs:858-860`, `vctrs_derive.rs:1095-1096`) all consistently emit both. Conservative match is correct.
- **fmt drift folded**: the `rpkg/src/rust/*.rs` changes (`gc_stress_fixtures.rs`, `dataframe_enum_payload_matrix.rs`, `jiff_adapter_tests.rs`, `lib.rs`, `dataframe_struct_flatten_test.rs`, `dataframe_collections_test.rs`, `columnar_option_none_tests.rs`) are pure rustfmt struct-literal expansion — no semantic change. Cleanly resolves pre-existing `just fmt` drift in line with the "fix warnings you see" principle.
- **Commit message + PR body**: clean, references `Closes #528`, lists test plan + acceptance recipe, attribution preserved.
- **CI**: clippy, lint, sync, generated files, wasm checks all green at review time. R CMD checks still pending but are not in the risk surface for this change.

Issue #576, opened 2026-05-15
.githooks/pre-commit: potential false positive on RoxygenNote-only updates


CGMossa:
### Verify applicability
- **State**: `still-applies`. Current hook at `.githooks/pre-commit:18-29` (verified):
  ```
  staged_wrappers=$(echo "$STAGED" | grep -E '(^|/)R/[a-z]+-wrappers\.R$' || true)
  staged_namespace=$(echo "$STAGED" | grep -E '(^|/)NAMESPACE$' || true)

  if [ -n "$staged_wrappers" ] && [ -z "$staged_namespace" ]; then
      echo "pre-commit: staged *-wrappers.R but NAMESPACE was not updated." >&2
      ...
      exit 1
  fi
  ```
- The hook fires when `*-wrappers.R` is staged but `NAMESPACE` is not. False positive: a `devtools::document()` run that only changes `man/*.Rd` and bumps `RoxygenNote` in `DESCRIPTION` (no NAMESPACE diff) still re-renders the wrappers stub if cdylib mtime is newer than R/. So contributor stages wrappers + Rd + DESCRIPTION; NAMESPACE legitimately unchanged; hook rejects.
- After #575 lands (drop `RoxygenNote` from `rpkg/DESCRIPTION`), the specific `RoxygenNote`-bump trigger disappears. **But the false-positive class persists**: any time a contributor edits a `///` doc-comment paragraph (no export change) and re-renders, NAMESPACE stays the same while wrappers + Rd shift.
- Per the issue's own Recommendation: "Don't implement this until PR #375's plan-A lands."

### Punchline rule
**The issue lists Proposed relaxation + Risk + Recommendation, with the closing constraint:** "Don't implement this until PR #375's plan-A lands. Once `RoxygenNote:` is removed from DESCRIPTION, the only way to legitimately stage Rd-only changes without NAMESPACE is `@noRd` toggling on existing exports (rare) or pure formatting (suspicious — should re-document). At that point the hook's current strictness is correct."

**The constraint says: after #575 lands, the hook's current strictness is correct. So implementing the relaxation directly is `needs-decision`.** Two paths:
- **(P1) Close as won't-fix once #575 lands.** Aligns with the issue's own recommendation.
- **(P2) Implement the relaxation anyway** because doc-comment-only changes (no export change) are a real workflow and the false-positive still hurts.

[...]

PR #609, opened 2026-05-17
feat(just): short-circuit devtools-document via needs_roxygenize (#373)

### LINKED ISSUE #373: Use roxygen2::needs_roxygenize() to short-circuit just devtools-document (roxygen2 8.0.0)
### Verify applicability
- **State**: `still-applies`.
- `just devtools-document` recipe in `justfile`:
  ```
  devtools-document: configure-fast
      Rscript -e 'devtools::document("rpkg")'
  ```
  Always runs `devtools::document()` end-to-end. No `needs_roxygenize()` short-circuit.
- `configure-fast` (dependency) already has mtime-based skip for `Makevars` / `.cargo/config.toml`, but doesn't gate the document step itself.
- `roxygen2 8.0.0` ships `needs_roxygenize(path, roclets)` — confirmed by the issue's blog-post citation; `Config/roxygen2/version: 8.0.0` is in `rpkg/DESCRIPTION:41`.
- The Rust cdylib rebuild + `miniextendr_write_wrappers` step is implicit in `devtools::document()` because it triggers `bootstrap.R` via the build hooks. That part is already handled by configure caching, but the second pass (parsing `///` comments → `man/*.Rd`) always runs.
- Two-layer caveat from the issue holds: layer 1 (cdylib regen) is mostly skipped via incremental cargo; layer 2 (roxygen2 parse + write) is the slow part and is what this issue targets.

### Punchline rule
No "Options:" + closing constraint. Issue's "Proposed work" section is the binding spec.

### Goal
`just devtools-document` short-circuits in under 1 s when no Rust source, no `R/`, and no man/* mtime changes warrant a re-document. Same logic applies to CI's roxygen-up-to-date verification and the pre-commit hook.

### Constraints
- No backwards compat shims.
- `Config/roxygen2/version: 8.0.0` (verified).
- The cache must invalidate when: (a) any `.rs` file under `rpkg/src/rust/**` changes, (b) any `.R` file under `rpkg/R/**` changes, (c) the macro layer (`miniextendr-macros/**`) changes — because tag serialization can shift. The issue's caveat warns about this — explicitly invalidate when macro / api source dirs change.
- Don't break first-time clone: `needs_roxygenize()` must report `TRUE` on a fresh checkout where `man/` is older than R/.
- Per CLAUDE.md: roxygen warnings are bugs to fix, not silence — the short-circuit must NOT mask warnings that would have surfaced.

### File:line refs
- `justfile` — `devtools-document` recipe (~3 lines around the `Rscript -e 'devtools::document(...)'` invocation).
- `justfile` — `configure-fast` recipe (pattern to follow: mtime check + fall-through).
- `.githooks/pre-commit:18-29` — optional: add a fast `needs_roxygenize()` call before flagging stale wrappers. Lower priority.
- `.github/workflows/ci.yml` — the roxygen-up-to-date check step (if one exists). Currently CI doesn't have a dedicated step; this would be a new "verify man/ in sync" job.
- No source code changes — pure tooling.

### Flat work items
1. Replace `devtools-document` recipe body with a needs_roxygenize-gated invocation:
   ```
   devtools-document: configure-fast
       Rscript -e 'if (roxygen2::needs_roxygenize("rpkg")) devtools::document("rpkg") else cat("devtools-document: man/ up to date — skipping\n")'
   ```
2. Verify `needs_roxygenize()` correctly tracks the cdylib-generated `R/miniextendr-wrappers.R` mtime. roxygen2's implementation scans every `R/*.R` file's mtime vs `man/*.Rd` mtime — wrappers.R is regenerated on every `R CMD INSTALL`, so this should already work.
3. Add a `force-document` recipe that bypasses the check (for cases where the short-circuit guesses wrong — e.g., macro layer changed in a way that doesn't bump any R-source mtime):
   ```
   force-document: configure-fast
       Rscript -e 'devtools::document("rpkg")'
   ```
4. Update `document-all` aggregate recipe to keep using `devtools-document` (no change needed).
5. Add a smoke test: on a freshly-touched `rpkg/src/rust/*.rs`, run `just rcmdinstall && just devtools-document` and confirm document runs. On no-op, confirm skip message.
6. (Optional, lower priority) Mirror the short-circuit in `minirextendr-document` and the cross-package `document-all`.
7. (Optional, low priority) Add a CI step that asserts `needs_roxygenize("rpkg")` returns FALSE on fresh checkout — catches stale-baseline PRs early.

### Bundling decision
**Separate PR from the codegen bundle (#578/#579/#586) and from #575/#576.** This is justfile-only tooling, orthogonal to:
- Codegen bundle (changes proc-macro output).
- #575 (DESCRIPTION metadata).
- #576 (pre-commit hook).

Combining with #575 is tempting (both are quality-of-life tooling) but they have different risk profiles: #575 is one-line; #373 needs a smoke test for the cache-invalidation gotcha (cdylib rebuild without source change). Keep separate for cleaner review.

### Non-goals
- Rewriting `devtools::document()` itself.
- Caching `R CMD INSTALL` output beyond what `configure-fast` already does.
- Adding mtime-aware caching to the macro layer (out of scope — that's a Rust-level optimisation).
- Touching `R CMD check` recipes — different workflow.
- Forward-looking roxygen2 8.0.0 design (#369–#374).

### Validation checklist
- `just check / clippy / fmt` — no Rust changes, no-op.
- `just configure && just rcmdinstall` (fresh).
- `just devtools-document` — first run regenerates; second run skips. Time the difference: expect <1 s for skip.
- Touch a `///` doc-comment in a `#[miniextendr]` fn; `just rcmdinstall && just devtools-document` re-runs `roxygenize()` correctly.
- `git diff` after a skip run — zero output.
- `just devtools-test` — confirm no regression.
- Cross-package: `just cross-test` (no recipe change there).

### Risks
- **Layer 2 caveat (issue body)**: `needs_roxygenize()` only reads source mtimes, not content. If the macro changes how it serializes a tag without bumping source mtimes, the short-circuit will miss it. Mitigation: the `configure-fast` step already runs (and would re-emit wrappers.R when configure inputs change). If macro layer changed without configure inputs, that's a separate latch (would also affect `R CMD INSTALL` output mtime — wrappers.R gets touched).
- A fix: have `force-document` documented in CLAUDE.md as the override path for "I changed the macro and the cache is lying."

### Expected diff shape
- `justfile`: ~6 lines (devtools-document body + new force-document recipe).
- `CLAUDE.md`: ~5 lines (mention force-document override).
- `.githooks/pre-commit`: optional, skipped for first cut.
- `.github/workflows/ci.yml`: optional, skipped for first cut.

### Test plan
- Manual smoke (documented in PR body):
  1. Fresh clone, `just rcmdinstall && time just devtools-document` — full run.
  2. Re-run `time just devtools-document` immediately — skips, <1 s.
  3. `touch rpkg/src/rust/some_demo.rs && just rcmdinstall && time just devtools-document` — re-runs roxygenize.
  4. `touch rpkg/R/some.R && time just devtools-document` — re-runs roxygenize.
- No automated test required (justfile recipe; mtime semantics).

## Original issue

Use roxygen2::needs_roxygenize() to short-circuit just devtools-document (roxygen2 8.0.0)

## Background

roxygen2 8.0.0 ships a `needs_roxygenize()` function:

> tells you whether your `.Rd` files are out of date by comparing modification times against their sources. It's much cheaper than running `roxygenize()` just to find out nothing has changed, which makes it a good fit for pre-commit hooks or CI checks.

Source: https://opensource.posit.co/blog/2026-05-01_roxygen2-8-0-0/

## Current state

`just devtools-document` always runs `devtools::document()` end-to-end, which compiles the cdylib (via `just rcmdinstall`) and re-runs miniextendr_write_wrappers + roxygen2 every invocation, even when nothing changed.

## Proposed work

- Wrap the devtools-document recipe in an early-exit when neither the cdylib output (`R/miniextendr-wrappers.R`) nor the hand-written R sources have changed since the last `.Rd` modification.
- Two layers:
  1. Skip the cdylib regen if Rust sources haven't changed — already handled by cargo's incremental build, but the dyn.load + write_wrappers step always runs.
  2. After cdylib regen, call `roxygen2::needs_roxygenize()` and skip the second pass if false.
- Apply same pattern in CI's roxygen-up-to-date check, and in the pre-commit hook (`.githooks/pre-commit`).

## Caveats

- `needs_roxygenize()` only reads source mtimes — won't detect when the macro layer changed how it serializes a tag. The Rust-source dependency layer still has to invalidate explicitly.
- Should not break first-time-clone setup.

Reference: roxygen2 8.0.0 adoption sweep, follow-up to plan in `plans/roxygen2-8-adoption.md`.

CGMossa:
Closes #373.

## Summary

- `just devtools-document` now calls `roxygen2::needs_roxygenize("rpkg")` before invoking `devtools::document()`. When the check returns `FALSE`, the recipe prints a skip message and exits in under 1 s — no full roxygenize pass.
- New `just force-document` recipe bypasses the check and always runs `devtools::document()`. This is the correct recipe to use after any Rust/macro change where the wrappers.R mtime may not have caught up.
- CLAUDE.md "After Rust changes" section updated: primary command is now `just rcmdinstall && just force-document`; `just devtools-document` is reserved for pure R/roxygen iteration loops.
- AGENTS.md mirrored with the same changes.

## Caveat: macro-layer cache misses

`needs_roxygenize()` tracks R-source mtimes only. Macro changes flow through `rcmdinstall` → `wrappers.R` mtime bump, which roxygen2 does detect — **but only if `rcmdinstall` ran first**. If you change `miniextendr-macros/**` and call `devtools-document` without first running `rcmdinstall`, the skip fires incorrectly. The documented convention (`just rcmdinstall && just force-document`) closes this gap.

## Test plan

- [ ] `just --list` shows both `devtools-document` and `force-document` recipes
- [ ] Fresh: `just configure && just rcmdinstall && just devtools-document` — runs full roxygenize
- [ ] Immediate re-run of `just devtools-document` — prints skip message, exits <1 s, `git diff` is empty
- [ ] `just force-document` always runs regardless of cache state
- [ ] Touch a `///` doc comment → re-run `just devtools-document` — detects change and regenerates
- [ ] `just devtools-test` passes with no regression

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,311 @@
+@/Users/elea/.codex/RTK.md
+
+# miniextendr
+
+Rust-R interoperability framework for building R packages with Rust backends.
+
+## Principles
+
+- **No backwards compatibility**: unreleased project — remove deprecated code, don't shim around it.
+- **Simple over complex**: only make changes directly requested or clearly necessary. Trust the framework — no defensive error handling for internal invariants.
+- **Fix warnings you see**: no "known issues". Every warning, lint, or test failure gets fixed, even if pre-existing and unrelated. Leave the code cleaner than you found it.
+- **Deferred items = GitHub issues**: any scope cut, known limitation, or partial fix needs `gh issue create` referenced in the PR. No "out of scope" without a tracked issue.
+- **`just` is maintainer-only**: end-user packages must build via `configure.ac` / `tools/*.R` / standard R mechanisms. Never require `just` in scaffolded packages.
+- **`configure.ac` never mutates sources**: don't rewrite Cargo.toml / Cargo.lock / .rs during `./configure` — dirties VCS. Use `cargo revendor --freeze` at vendor time.
+- **`configure.ac` must not call `minirextendr::*`**: put helpers in `tools/` and invoke via `Rscript tools/foo.R`.
+- **Edit `.in` templates, not generated files**:
+  - `rpkg/src/rust/.cargo/config.toml` ← `rpkg/src/rust/cargo-config.toml.in`
+  - `rpkg/src/Makevars` ← `rpkg/src/Makevars.in`
+  - `rpkg/src/miniextendr-win.def` ← `rpkg/src/win.def.in`
+  - `rpkg/configure` ← `rpkg/configure.ac` (then `autoconf`)
+  - `rpkg/src/stub.c` — static, no substitution.
+- **Collect all errors in vectorized ops**: don't bail on first failure — give users one batched diagnostic.
+- **Flat plans, no phases**: list work in flat priority order, not "Phase 1/2/3".
+- **Prefer `From`/`TryFrom` over `as` casts**: propagate the error rather than silently truncating.
+
+### Rust/FFI gotchas
+
+- **Pointer provenance**: cache `*mut T` via a mutable path (`&mut T`, `Box::into_raw`, `downcast_mut`, `ptr::from_mut`). Never write through a `cached_ptr` derived from `&T` / `downcast_ref` — UB under Stacked Borrows.
+- **`cargo package` for workspace resolution**: when vendoring, let `cargo package` expand workspace inheritance — never hard-code workspace dependency replacements.
+- **m4 in `AC_CONFIG_COMMANDS`**: `$1` is empty (use `$0` or avoid `sh -c`). Escape `[` / `]` in sed/grep as `@<:@` / `@:>@`.
+- **Windows paths in TOML**: forward slashes only. Strip `\\?\` prefix from `canonicalize()` output before writing.
+- **macOS tar xattrs**: set `COPYFILE_DISABLE=1` when creating tarballs to avoid Apple metadata warnings on Linux/Windows GNU tar.
+- **`cargo-revendor`**: standalone workspace (excluded from miniextendr workspace). Build/test via `just revendor-build`/`just revendor-test`. `--freeze` resolves `Cargo.toml` against `vendor/` only.
+
+### `just` vs raw `cargo`
+
+Always use `just check/clippy/test/fmt/vendor/lint` — the recipes iterate every manifest in the multi-crate workspace (root, `rpkg/src/rust`, `tests/cross-package/*`) with the correct `[patch.crates-io]` overrides. Raw `cargo --workspace` only sees the top-level manifest. If a recipe doesn't exist, say so and fall back — don't invent shortcuts.
+
+**Interactive iteration**: prefer `cargo-limit` aliases (`cargo lcheck/lclippy/ltest/lbuild`) — they truncate output to the first few errors. Install with `just dev-tools-install`. CI and `just` recipes keep plain `cargo` (CI needs full output + `-D warnings`).
+
+## Project Structure
+
+```
+miniextendr-api/      # Runtime (FFI, ExternalPtr, ALTREP, worker thread)
+miniextendr-macros/   # Proc macros (#[miniextendr], derives; naming in src/naming.rs)
+miniextendr-bench/    # Benchmarks (separate workspace member)
+miniextendr-lint/     # Static analysis
+miniextendr-engine/   # Code generation engine
+cargo-revendor/       # Standalone cargo subcommand (not in workspace)
+rpkg/                 # Example R package (installed as `miniextendr`)
+minirextendr/         # Pure R scaffolding helper
+tests/cross-package/  # producer.pkg / consumer.pkg — trait ABI tests
+site/                 # Zola docs → GitHub Pages
+background/           # Reference docs (gitignored)
+```
+
+## Build Commands
+
+```bash
+# Rust
+just check / test / clippy / fmt / lint
+
+# rpkg (R package)
+just configure           # REQUIRED before any R CMD operation (dev mode)
+just rcmdinstall         # Build + install (compiles Rust, auto-generates R wrappers)
+just devtools-document   # roxygen2 → NAMESPACE + man/ (short-circuits if nothing changed)
+just force-document      # like devtools-document but always runs — use after macro changes
+just devtools-test       # R tests
+just r-cmd-build         # Build tarball
+just r-cmd-check         # Check built tarball (save to log — see Capturing Output)
+just devtools-check      # Check, preserving output in rpkg-check-output/
+
+# CRAN release prep (only step needed; configure auto-detects tarball mode)
+just vendor              # regen Cargo.lock in tarball-shape, vendor deps, compress to inst/vendor.tar.xz
+
+# Cross-package
+just cross-install / cross-test / cross-check
+
+# minirextendr
+just minirextendr-install / minirextendr-test / minirextendr-check
+
+# Site
+just site-build / site-serve   # http://127.0.0.1:1111
+```
+
+### Configure is mandatory
+
+Always `bash ./configure` (not bare `./configure` — `#!/bin/sh` causes spurious errors in `AC_CONFIG_COMMANDS` passthrough). Configure:
+1. Generates `Makevars` from `.in` templates
+2. Auto-detects install mode (source vs tarball) from `[ -f inst/vendor.tar.xz ]`
+3. Writes `.cargo/config.toml` per mode (source: `[patch."git+url"]` for monorepo siblings or empty; tarball: `[source]` replacement to `vendored-sources`)
+4. Does **not** create `inst/vendor.tar.xz` — that's `just vendor`
+
+Package loads as `library(miniextendr)`, not `library(rpkg)`. Always check the **built tarball**, not the source dir (R CMD check on a source dir skips `Authors@R` → `Author/Maintainer` conversion).
+
+### Install modes (source vs tarball)
+
+| Mode | Triggered when | What configure does |
+|---|---|---|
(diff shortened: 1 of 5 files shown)


CGMossa (commented):

### Verdict
minor — non-blocking but ship with the doc cleanup

### Summary

The conceptual model is correct, the `justfile` recipes are sound, and the new convention (`force-document` after macro changes) is the right call. The risk surface is documentation visibility: the root `CLAUDE.md`/`AGENTS.md` were updated, but two sub-directory `CLAUDE.md` files — including the proc-macro crate's — still recommend `just devtools-document` in the exact macro-cache-miss scenario the PR is warning about. Fix those and this is ready.

### Required-ish changes (cheap, do before merge)

- [ ] `rpkg/CLAUDE.md:16` — "After **anything** that affects R wrapper output (proc-macro roxygen, `r_wrappers.rs`, adding `#[miniextendr]` fns) run `just devtools-document`..." should say `just force-document`. This is the exact trap.
- [ ] `miniextendr-macros/CLAUDE.md:40` — same trap, and arguably more dangerous since it's loaded specifically when an agent is working on proc-macros: "Touched proc-macro output? Run `just configure && just rcmdinstall && just devtools-document`..." should be `... && just force-document`.

- `CLAUDE.md:348` and `AGENTS.md:200` "Common issues → could not find function" still say `just configure && just rcmdinstall && just devtools-document`. In this specific path the short-circuit won't fire (the new function bumps `wrappers.R` content → mtime → `needs_roxygenize()` returns TRUE), so it's benign. But for consistency with the new convention, swap to `force-document`.
- `CLAUDE.md:364` / `AGENTS.md:218` mention `just devtools-document` in the sandbox-compilation note as an example of a compiling command. Either swap to `force-document` (the recommended recipe after Rust changes) or leave both (the sandbox note is generic). Pick one for consistency.
- `document-all` (justfile:606) depends on `devtools-document`, which is intentional and the right call for the "regenerate everything that needs regenerating" semantics — agreed with the implementer's PR-body reasoning. No change needed; just flagging that future readers may want a comment on the recipe explaining the decision.
- `devtools-check` (justfile:579) depends on `devtools-document`. In the pre-existing flow, this guaranteed a re-document before `R CMD check`. Post-PR, if `man/` is genuinely in sync with `R/` sources, the skip fires and `R CMD check` runs against the existing docs — fine. But if a contributor ran `rcmdinstall` that didn't change `wrappers.R` content (so no mtime bump) and an upstream macro change silently invalidated an `.Rd` template, the skip is wrong. This is the implementer's own caveat; mitigated by the `force-document` convention, but worth a one-line comment on `devtools-check` referencing the convention.
- The skip message uses backticks inside an R `cat()` string inside single-quoted Rscript `-e`. Verified: no shell interpolation occurs and the rendered message reads `(use \`just force-document\` to override)` — clean.
- The Rscript exit-code handling is correct: `needs_roxygenize()` returns logical; if `roxygen2` is missing or `rpkg/` doesn't exist, Rscript exits non-zero and the recipe fails loudly. No silent skip.
- Future polish: hoist the `Rscript -e '...'` body into a `tools/needs-doc.R` to keep the justfile readable; the conditional has grown enough.

### Looks good

- Two-recipe design (`devtools-document` + `force-document`) is the right shape — bypass path is explicit, not an env var.
- Recipe comment block in `justfile:582-594` is excellent: documents the cache-miss caveat *at the recipe definition* where the next reader of the recipe will see it.
- Both recipes share `Rscript -e 'devtools::document("rpkg")'` body via the `roxygen2::needs_roxygenize` branch — DRY enough; not worth extracting.
- `document-all` correctly depends on `devtools-document` (skip-on-no-change) — fast aggregate.
- Pre-commit hook interaction (#576) is unchanged: the hook fires on STAGED `*-wrappers.R`, which by definition means the wrappers content changed and mtimes are fresh, so the short-circuit can't have suppressed a legitimate regen. PR #609 does not reintroduce the false-positive class.
- `Config/roxygen2/version: 8.0.0` is already in `rpkg/DESCRIPTION`, so `needs_roxygenize()` is guaranteed available.
- Commit message + PR body are clean. "Closes #373", caveat called out explicitly, test plan is reasonable, Co-Authored-By present.
- AGENTS.md mirror is faithful to CLAUDE.md for the new section.

### One conceptual note for the record

The implementer's PR body says "if you change `miniextendr-macros/**` and call `devtools-document` without first running `rcmdinstall`, the skip fires incorrectly." This is *technically* true but **was equally broken before this PR** — running `devtools::document` without `rcmdinstall` always produced stale outputs because `devtools::document` only reads `R/`, not Rust source. The new behavior just changes the symptom from "slow stale output" to "fast stale output." The `force-document` convention is the right belt-and-suspenders, but the framing in the PR body slightly overstates the regression risk. Not a change request — just clarifying for the merge-decision record.

### Verification done

- `git diff origin/main..HEAD -- justfile CLAUDE.md AGENTS.md`
- Recipe walkthrough (justfile:579-610) including dependent recipes (`devtools-test`, `devtools-load`, `devtools-install`, `devtools-check`, `document-all`).
- Pre-commit hook trace (`.githooks/pre-commit:10-28`) confirms no new false-positive surface.
- `rpkg/src/Makevars.in:100-134` confirms `miniextendr_write_wrappers()` is write-if-changed, which is the actual source of the macro-cache-miss caveat.
- Cross-package `document-all` (`tests/cross-package/justfile:15`) uses a custom recipe path and is unaffected.

Issue #586, opened 2026-05-15
miniextendr macro: attributes between doc-comment blocks silently corrupt roxygen output


CGMossa:
### Verify applicability
- **State**: `still-applies` at the macro level. #585 patched the **source-side symptom** in `rpkg/src/rust/connection_tests.rs` (commit `56038f18` removed 13 lines of duplicate doc-comments), but did not change the macro behaviour.
- Current implementation at `miniextendr-macros/src/roxygen.rs:91-127`:
  ```
  for attr in attrs {
      if !attr.path().is_ident("doc") { continue; }   // SILENTLY SKIPS non-doc attrs
      ...
      for line in lit.value().lines() {
          if trimmed.starts_with('@') { tags.push(...); }   // continues feeding into the same `tags` Vec
          else if !trimmed.is_empty() {
              if tags.is_empty() { regular_docs.push(...); }
              else if last is multiline { last.push(...); }   // ← BUG: continuation lands inside the prior @tag
          }
      }
  }
  ```
- When a non-doc attribute (`#[cfg(...)]`, `#[deprecated]`, `#[doc(hidden)]`) splits two `#[doc = ...]` runs, the second run's prose continues appending to whatever `tags.last_mut()` happens to be (`@examples` or `@details` or `@return`). End result: bare prose lands inside `\examples{}` in the rendered Rd, R's `tools::Rd2ex` extracts it as code, `parse()` fails, R CMD check explodes.
- Same callers as #578: `lib.rs:1113`, `miniextendr_impl.rs:1943,2275`, `miniextendr_impl_trait.rs:282`, `miniextendr_impl_trait/vtable.rs:146,528`. Both bare-fn and impl-method paths affected.
- No MXL rule covers this; lint catalogue at `miniextendr-lint/src/lib.rs` doesn't include "doc/attr interleave".

### Punchline rule
**Issue lists Options A/B/C/D with closing recommendation "Recommend (B) or (C). Probably (B) — once we're touching the doc-comment assembly path, normalising section order makes the macro more forgiving in general."** The closing constraint is a strong recommendation for (B), not a hard binding. (D) is explicitly worse than status quo. (A) is the only one that doesn't fix valid user code.

**Chosen approach: hybrid of (A-lint) + (C-narrow normalisation).** Reasoning:
[...]

Issue #579, opened 2026-05-15
S7 property method docs: paragraph 2+ silently dropped from generated Rd


CGMossa:
### Verify applicability
- **State**: `still-applies`.
- `miniextendr-macros/src/miniextendr_impl/s7_class.rs:261-271` is the extraction:
  ```
  entry.doc = method.doc_tags.iter().find_map(|t| {
      if !t.starts_with('@') { return Some(t.clone()); }
      t.strip_prefix("@description ")
          .or_else(|| t.strip_prefix("@title "))
          .map(|rest| rest.lines().next().unwrap_or(rest).to_string())
  });
  ```
  This pulls only the first line of `@description` (or fallback `@title`), discarding all multi-line content and any `@details` content entirely.
- Combined with #578 (single-para → `@title` only), an S7 property with a 1-para getter doc will fall through to `@title`; the existing `.or_else()` covers that case. But once the property's getter has 2+ paragraphs, paragraph 2 lives in `@description` (or `@details` once #578 lands) and is **dropped** from `@prop`.
- Issue's pre-existing trace (`s7_class.rs:261-272`) matches current source — no in-flight fix has landed.

### Punchline rule
**Issue lists Options 1/2/3 with no closing constraint sentence (only "Option 1 is the cheap regression-recovery; option 2 or 3 is the proper fix").** No binding constraint — but recommendation is options 2/3 ("proper fix"). For this plan we pick **Option 2 (Multi-line @prop)** because (a) it's the least invasive change to the macro side and (b) roxygen2 8.0.0 supports continuation lines under `@prop` natively (same MULTILINE_TAGS-style behaviour as `@param`). Option 3 (`@section` per property) overlaps with roxygen2 8.0.0's design-owed features and belongs in #369–#374.

### Goal
S7 properties whose getter doc has 2+ paragraphs render the full content under the class-level `@prop name <doc>` block (multi-line continuation). No content silently dropped from generated `.Rd`.

### Constraints
- No backwards compat shims.
- `Config/roxygen2/version: 8.0.0` (verified).
- The `@prop` tag is multi-line — roxygen2 8.0.0 treats indented continuation lines as part of the prop's `\describe{}` entry.
- Don't introduce blank-line continuations inside `@prop` (some roxygen2 versions terminate at blank line). Use single-line continuation with leading `  ` indent.
[...]

Issue #374, opened 2026-05-04
R6 superclass parameter inheritance for generated subclass docs (roxygen2 8.0.0)


CGMossa:
## Background

roxygen2 8.0.0:

> When a method overrides one from a superclass, it now automatically inherits that parameter documentation — so common parameters only have to be documented once, on the base class. The same is true for inherited fields and active bindings.

Source: https://opensource.posit.co/blog/2026-05-01_roxygen2-8-0-0/

## Current state

When miniextendr generates an R6 subclass via `#[miniextendr(r6(parent = "Foo"))]`, every method's `@param` block is emitted in full, regardless of whether the parent already documents the same parameter.

## Proposed work

- Audit cross-package fixtures with R6 inheritance (none exist today; will need to add one).
- Investigate whether the auto-inheritance also fires for methods generated from Rust-side traits with default impls.
- If so, drop redundant `@param` emission for inherited methods.
- Add a fixture where a method overrides a parent method without re-documenting params, and verify .Rd inherits the parent's @param.

Reference: roxygen2 8.0.0 adoption sweep, follow-up to plan in `plans/roxygen2-8-adoption.md`.

Issue #369, opened 2026-05-04
R6 codegen: migrate to $set() form for inline method docs (roxygen2 8.0.0)


CGMossa:
## Background

roxygen2 8.0.0 supports documenting R6 methods *outside* the class body when added via `Counter$set("public", "name", function(...) {...})`. See https://opensource.posit.co/blog/2026-05-01_roxygen2-8-0-0/

> you no longer need to document all methods inside the class definition. If you add methods with `\$set()`, you can now document them directly

## Current state

`miniextendr-macros/src/miniextendr_impl/r6_class.rs` emits a single `R6::R6Class("Name", public = list(method_name = function(...) {...}, ...))` block with all methods inlined. Methods carry roxygen tags inline. Already produces correct `Methods` / `Public methods` sections under 8.0.0.

## Possible migration

Restructure codegen to emit:

```r
ClassName <- R6::R6Class("ClassName", ...)

#' @description Method description
#' @param x ...
ClassName\$set("public", "method_name", function(x) { ... })
```

Each method gets its own roxygen block.

## Why non-trivial

- All 6 class generators share `r_class_formatter.rs` infrastructure that assumes inline emission.
- Snapshot tests in `miniextendr-macros/src/miniextendr_impl/tests.rs` need rebaselining.
- `\$set()` is documented as having performance implications when called many times — verify install-time cost.
- Existing rpkg fixtures + cross-package consumer.pkg/producer.pkg + scaffolded packages all rebuild.

Reference: roxygen2 8.0.0 adoption sweep, follow-up to plan in `plans/roxygen2-8-adoption.md`.

PR #612, opened 2026-05-17
feat(macros): roxygen codegen bundle — @title, multiline @prop, attr-interrupt safety

### LINKED ISSUE #578: codegen: method docs should emit @title (not @description) for single-paragraph cases per roxygen2 convention
### Verify applicability
- **State**: `still-applies`.
- `miniextendr-macros/src/roxygen.rs:179-183` (the `auto_description && tags.is_empty()` branch) still emits `tags.push(format!("@description {}", description));` for single-paragraph method docs. No `@title` is emitted on this path.
- The "tags present but no @title" branch at `roxygen.rs:140-145` inserts `@title` from `implicit_title_from_attrs`, but it only fires when at least one explicit `@tag` exists; the bare-prose-only path (the common R6 method case) bypasses it.
- Caller sites:
  - `miniextendr_impl.rs:1943` — `roxygen_tags_from_attrs_for_r6_method(&item.attrs)` used by R6/S3/S4/S7/Env/Vctrs impl methods.
  - `miniextendr_impl_trait.rs:282`, `miniextendr_impl_trait/vtable.rs:146,528` — trait-impl methods (also affected).
  - `lib.rs:1113` — bare-fn path; not method-level.
- PR #585 ran a markdown=TRUE bulk regen but did not change the `@title`/`@description` split logic — single-paragraph method Rds still have empty `\title{}` filled implicitly by roxygen2 from `\description{}` first line.

### Punchline rule
Issue lists fix-sketch code (single Resolution block), not Options + closing constraint. No constraint binding to check.

### Goal
Single-paragraph impl-method docs render as `@title <text>` (only), 2-paragraph docs render as `@title <p1>` + `@description <p2>`, 3+ render as `@title <p1>` + `@description <p2>` + `@details <p3..>`. All 6 class systems and trait-impls consume the new mapping. Re-rendered `man/*.Rd` files have populated `\title{}` and no roxygen2 "Empty section" warnings.

### Constraints
- No backwards compat shims (per CLAUDE.md).
- `Config/roxygen2/version: 8.0.0` already in DESCRIPTION.
- Don't break the stable `man/*.Rd` baseline from PR #585 unexpectedly — accept method-doc reshuffling (\title/\description swap) but no spurious other churn.
- Every roxygen2 warning is a bug. Don't silence with `@noRd`/`@keywords internal`. R6 active-binding opt-out, if encountered, is `@field name (internal)`, NOT `@field name NULL`.
- S7 `@prop` extraction at `s7_class.rs:261-271` already falls back from `@description` to `@title` — verify the fallback still works after the switch (it does — line 269 already chains `.or_else(t.strip_prefix("@title "))`). #579 will further harden this.

### File:line refs
- `miniextendr-macros/src/roxygen.rs:91-186` `roxygen_tags_from_attrs_impl`.
- `miniextendr-macros/src/roxygen.rs:179-183` — the `auto_description && tags.is_empty()` bare-prose path; this is the precise edit site.
- `miniextendr-macros/src/roxygen.rs:310-360` `implicit_title_from_attrs` (already exists, ready to call).
- `miniextendr-macros/src/roxygen.rs:371-428` `implicit_description_from_attrs` (already exists).
- `miniextendr-macros/src/roxygen.rs:440-503` `implicit_details_from_attrs` (already exists).
- `miniextendr-macros/src/miniextendr_impl/s7_class.rs:261-271` — fallback `@title` chain. After this change `entry.doc` will be hit via the `@title` strip — verify multi-paragraph isn't dropped (overlaps with #579).
- `miniextendr-macros/src/miniextendr_impl.rs:1943` — `roxygen_tags_from_attrs_for_r6_method` consumer site.
- `miniextendr-macros/src/r_class_formatter.rs:472-700` `MethodDocBuilder` — verify it consumes whatever tags come in without filtering `@title`.

### Flat work items
1. Edit `roxygen_tags_from_attrs_impl` in `roxygen.rs:91-186`: in the `auto_description && tags.is_empty() && !regular_docs.is_empty()` branch, replace single `@description` emission with the `@title` / `@title + @description` / `@title + @description + @details` mapping driven by paragraph count in `regular_docs`. Use the existing implicit_*_from_attrs helpers — they already work on the raw `attrs` slice, so the simplest implementation is to call those three helpers in this branch (and skip the manual `regular_docs.join(" ")`).
2. Confirm `MethodDocBuilder.build()` in `r_class_formatter.rs:621+` passes `@title` through unchanged (it should — it just iterates tags).
3. Verify all 6 class generators consume `doc_tags` via the builder unchanged: `r6_class.rs`, `s3_class.rs`, `s4_class.rs`, `s7_class.rs`, `env_class.rs`, `vctrs_class.rs`. No per-generator edits expected.
4. Update / add roxygen tests in `miniextendr-macros/src/roxygen/tests.rs` (already has @details integration tests at L209+).
5. `just configure && just rcmdinstall && just devtools-document 2>&1 > /tmp/devtools-doc.log` — re-render `man/*.Rd`. Expected: every single-paragraph method's `\title{}` populated; the matching `\description{}` either empty or duplicating title (roxygen2 8.0.0 default).
6. Read `/tmp/devtools-doc.log` — assert zero "Empty section: title" / "Empty section: description" warnings.
7. Commit `rpkg/R/miniextendr-wrappers.R`, `NAMESPACE`, `man/*.Rd` in sync.

### Bundling decision
**Bundle with #579 and #586.** All three touch the doc-extraction pipeline (`roxygen.rs` + `s7_class.rs`), all three regen `man/*.Rd`. Splitting forces sequential PRs to each rebase the Rd baseline.

### Non-goals
- Forward roxygen2 8.0.0 codegen features (#369–#374). Specifically: `@param` inheritance changes, `@inheritDotParams` rework, S7 property-group syntax. Tracked in PR #375.
- Refactoring the implicit_*_from_attrs helpers themselves.
- Class-level `@description` / `@title` handling (already covered by ClassDocBuilder at `r_class_formatter.rs:356-422`).

### Validation checklist
- `just check / clippy / fmt`.
- CI `clippy_default` and `clippy_all` (feature list in `.github/workflows/ci.yml:226-228`).
- `just configure && just rcmdinstall && just devtools-document` — commit regenerated `rpkg/R/miniextendr-wrappers.R` + `NAMESPACE` + `man/*.Rd` in sync.
- `just devtools-test`.
- **Roxygen warnings = 0** in `devtools-doc.log`.
- `R CMD check` smoke (`just r-cmd-check`) — confirm no "Empty section" diagnostics introduced.
- UI snapshots: `TRYBUILD=overwrite cargo test -p miniextendr-macros`.
- Cross-package + minirextendr: `just cross-test && just minirextendr-test`.

### Expected diff shape
- `miniextendr-macros/src/roxygen.rs`: ~20 lines (replace the L179-183 block).
- `miniextendr-macros/src/roxygen/tests.rs`: ~40 lines new tests.
- `miniextendr-macros/src/miniextendr_impl/snapshots/*.snap*`: ~10–20 lines auto-update.
- `rpkg/man/*.Rd`: ~150–210 files affected, reshuffling `\title{}` / `\description{}` per method.
- `rpkg/R/miniextendr-wrappers.R`: 100–200 lines reflecting `@title` lines in place of `@description`.

### Test plan
- New unit tests in `roxygen/tests.rs` covering: 1-para → only `@title`; 2-para → `@title`+`@description`; 3+-para → `@title`+`@description`+`@details`. Each exercises `roxygen_tags_from_attrs_for_r6_method` (the auto_description=true path).
- A new fixture in `rpkg/src/rust/` (or reuse `multi_para_doc_demo.rs`) with a 1-para method, a 2-para method, a 3-para method; rendered Rd asserted via golden-file snapshot.
- Cross-check rendered output against `background/S7-main/man/`, `background/R6-main/man/` to confirm idiom alignment.

## Original issue

codegen: method docs should emit @title (not @description) for single-paragraph cases per roxygen2 convention

Discovered during PR #577 (closing #343).

## Current behavior (post-#577 minimal trim)

For R6/S3/S4/S7/Env/Vctrs impl methods with no explicit `@tags`, the `auto_description` fallback in `miniextendr-macros/src/roxygen.rs` emits:

- Single-paragraph doc → `@description <text>`
- Multi-paragraph doc (this is what #343 fixed) → `@description <p1>` + `@details <p2..>`

## Proposed behavior

Per roxygen2 conventions, `@title` is the first sentence/paragraph and `@description` is optional/inferred from `@title` when absent. The implementer of #577's first draft emitted:

- Single-paragraph doc → `@title <text>` (only)
- 2-paragraph doc → `@title <p1>` + `@description <p2>`
- 3+ paragraph doc → `@title <p1>` + `@description <p2>` + `@details <p3..>`

This is a more correct alignment with roxygen2's documented convention, but it cascades to ~210 existing `man/*.Rd` files (every single-paragraph method's `\title{}`/`\description{}` reorganized). To keep #577 focused on the actual paragraph-loss bug, that change was scoped out.

## Why this matters

The current setup produces `\description{<text>}` with an empty/implicit `\title{}`. roxygen2 either fills `\title{}` from `\description{}`'s first line at render time or warns. Either way the rendered docs look correct, but the Rd source structure is non-idiomatic.

## Resolution

Adopt the full title/description/details mapping in `auto_description`:

```rust
if auto_description && tags.is_empty() {
    if let Some(title) = implicit_title_from_attrs(attrs) {
        tags.push(format!("@title {}", title));
    }
    if let Some(desc) = implicit_description_from_attrs(attrs) {
        tags.push(format!("@description {}", desc));
    }
    if let Some(details) = implicit_details_from_attrs(attrs) {
        tags.push(format!("@details {}", details));
    }
}
```

(The helpers already exist; #577 added them. The only change is which branch calls them.)

## Acceptance

- [ ] All 6 class-system method generators emit `@title` (or `@description` only when the doc is genuinely titleless — edge case)
- [ ] Re-render of all `man/*.Rd`: `\title{}` and `\description{}` are sensibly populated
- [ ] `roxygen2::roxygenise()` emits no `"Empty section: title"` or `"Empty section: description"` warnings against the rendered Rd
- [ ] Render a few examples side-by-side and confirm CRAN-package conventions are met
- [ ] Cross-check against `vctrs`/`S7`/`R6` reference repos in `background/`

Refs: #343, #577.

CGMossa:
## Issues closed

Closes #578, #579, #586

## Summary

- **#578 — `@title` instead of `@description`**: The `auto_description` path in `roxygen_tags_from_attrs_impl` now calls `implicit_title_from_attrs` / `implicit_description_from_attrs` / `implicit_details_from_attrs` for proper section mapping instead of always emitting `@description`. Single-paragraph methods now render with `@title` (as roxygen2 expects), not `@description`.

- **#579 — multi-paragraph `@prop` docs**: S7 property getter docs now emit all paragraphs via continuation lines (`#'   text`). `"prop"` added to `MULTILINE_TAGS`. A new `extract_prop_doc_from_tags` helper collects `@title`/`@description`/`@details` from the getter's `doc_tags` and joins them with blank lines. The `s7_tests.rs` `S7Range::length` property has a 3-paragraph doc as a regression fixture.

- **#586 — attr-interrupt safety (hybrid fix)**: Non-doc attributes (`#[cfg(...)]`, `#[deprecated]`, etc.) that arrive mid-doc-comment stream no longer bleed bare trailing prose into the preceding `@return`/`@examples`/`@details` multiline tag. `roxygen_tags_from_attrs_impl` now resets multiline-continuation context at the interruption point. New lint **MXL302** (`doc_attr_interleave.rs`) warns with a help message to move `///` above non-doc attrs. 12 rpkg source locations with the pattern corrected.

## What was NOT done (filed as issues)

- Tag-reordering (Option B from #586: sorting attrs so docs always precede non-doc attrs) deferred to #613 — overlaps with #369–#374 design work.

## Test plan

- [x] `cargo test -p miniextendr-macros` — 315 passing, all snapshots updated
- [x] `cargo check --workspace --locked`
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (default + all-features)
- [x] `cargo fmt --all -- --check`
- [ ] `just rcmdinstall && just devtools-document` — blocked by R 4.7/rv 4.5 version mismatch on agent machine; CI will regenerate `man/*.Rd` + `NAMESPACE`


CGMossa (commented):
## Reviewer notes (Opus, 2026-05-18, retry)

### Verdict
blocking — generated R wrappers + Rd files were not regenerated

### Required changes
- [ ] **Regenerate `rpkg/R/miniextendr-wrappers.R` and `rpkg/man/*.Rd`.** The macro deltas in this PR change real output but `git diff origin/main...HEAD --name-only | grep -E 'rpkg/(R/|man/|NAMESPACE)'` is empty. Three concrete pieces of evidence:
  1. Snapshot diffs show R6/env class method docs flipping from `@description <text>` to `@title <text>` (`snapshot_env_basic.snap`, `snapshot_r6_basic.snap`). The current `rpkg/R/miniextendr-wrappers.R` on the PR branch still contains 19 `@description` lines for `R6Counter` / `R6Accumulator` / similar methods that should now read `@title`.
  2. `rpkg/src/rust/s7_tests.rs` was edited to give `S7Range::length` a second paragraph ("The value is always `end - start` and is recomputed on each access"). Combined with the #579 multi-line `@prop` work, the wrapper for `S7Range` should now emit a multi-line `@prop length` (first line + indented continuation). The committed file still has the single-line, joined form: `#' @prop length Computed property: ... Since there's no setter, it's a computed (read-only) property.`
  3. The contraction edit in the same docstring (`it's a` → `it is a`) is also not reflected in the wrapper.
  Run `just configure && just rcmdinstall && just devtools-document` and commit the resulting `rpkg/R/miniextendr-wrappers.R`, `rpkg/NAMESPACE` (if changed), and `rpkg/man/*.Rd` deltas. The pre-commit hook would have caught a wrappers-without-NAMESPACE mismatch, but no wrappers staged means the hook did not fire.

- `extract_prop_doc_from_tags` in `s7_class.rs` reconstructs paragraphs from the post-parse `doc_tags` strings by stripping `@title `/`@description `/`@details ` prefixes. This works today but couples to the exact format the parser emits. If a future tag parser change inserts a newline or alternate spacing, the silent fallback (`parts` stays empty → `None`) would drop @prop docs without test failure. A direct API like `implicit_*_from_attrs(attrs)` (already exported in `roxygen.rs`) would be more robust here. Not worth blocking; consider following up under #613.
- `interrupted_by_non_doc` resets to `false` whenever a new `@tag` line is seen. That looks correct (the user explicitly opened a new tag, so the interruption no longer poses a tag-bleed risk), but the reset is silent — it would be worth a one-line test asserting that `/// @return X` → `#[cfg(foo)]` → `/// @return Y` keeps both `@return` lines distinct rather than bleeding. The other interrupt cases are well-covered in `roxygen/tests.rs`.

### Looks good
- MXL302 severity is `Warning` (consistent with the "macro recovers automatically" stance) and the message + help-text clearly explain the antipattern. Rule wiring in `lint_code.rs` matches the existing MXL300/MXL301 pattern.
- The 12 (8 file) rpkg source-side rewrites are all pure cosmetic reorderings: `#[miniextendr]` / `#[allow(...)]` / `#[unsafe(no_mangle)]` move below the `///` block, function bodies untouched. Spot-checked `conversions.rs`, `dots_tests.rs`, `panic_tests.rs`, `thread_tests.rs`, `unwind_protect_tests.rs`, `externalptr_identity_tests.rs`, `r6_tests.rs`, `s7_tests.rs`. No semantic changes.
- `"prop"` correctly added to `MULTILINE_TAGS` in `roxygen.rs` so the s7 `extract_prop_doc_from_tags` output round-trips through `is_multiline_tag`.
- `auto_description` path in `roxygen.rs:roxygen_tags_from_attrs_impl` now invokes `implicit_title_from_attrs` / `implicit_description_from_attrs` / `implicit_details_from_attrs` in the right order, mirroring roxygen2's section mapping.
- Follow-up #613 (sort non-doc attrs after doc-comments) is filed as `OPEN: feat(macros): tag-reordering — sort non-doc attrs after doc-comments on #[miniextendr] items`.
- Snapshot updates are minimal and all match expected behaviour (only the assertion-line drops + the `@description` → `@title` flips + the two trailing-period removals in `@prop` lines, which are a side effect of the multi-line splitter — verify these are intended in the next regen).

PR #617, opened 2026-05-18
refactor(macros): collapse MethodReturnBuilder build* duplication (#362)

### LINKED ISSUE #362: MethodReturnBuilder: collapse 5 near-duplicate build* methods
` miniextendr-macros/src/method_return_builder.rs:244-463 ` defines five methods that share an identical scaffold:

- ` build() `
- ` build_r6_body() `
- ` build_s3_body() `
- ` build_s7_inline() `
- ` build_s4_inline() `

Each one independently re-implements the ` if self.error_in_r { ... } else { ... } ` × ` match strategy { ReturnSelf | ChainableMutation | Direct } ` matrix. The class-system-specific bits are 2-3 lines per method; the rest is pure copy-paste of the error-check / dispatch scaffolding.

## Proposal
Extract a single method that takes the class-system-specific tail (the lines that emit ` self ` / ` invisible(x) ` / ` .val ` etc.) as a small closure. The shared error-check + dispatch loop runs in one place, and the five public ` build* ` methods become thin wrappers that pass their tail and ` indent `.

## Why not in #351
This file is hot path for the macro codegen and was load-bearing in PR #344 (condition system). Worth doing alone with focused testing rather than as part of a cleanup pass.

## Adjacent finding
` MethodReturnBuilder::error_check_lines ` at ` :237 ` is a 1-line wrapper that just re-calls the public ` error_in_r_check_lines ` free function with no transformation. Should be removed; the two callers (` build() ` and ` build_r6_body() `) can call the free function directly, matching the other variants. Also: ` error_in_r_inline_block ` and ` error_in_r_standalone_body ` (` :96,116 `) hardcode indent strings (` \"      \" `, ` \"    \" `) instead of taking ` indent ` as a parameter — inconsistent with the rest of the builder API.

Found via /simplify pass on prepare-release; not addressed in #351.

CGMossa:
## Summary

Collapses ~220 lines of `error_in_r × (ReturnSelf | ChainableMutation | Direct)` duplication in `method_return_builder.rs` into a shared core path, with class-specific tail closures. Parameterises `indent` on `error_in_r_inline_block` and `error_in_r_standalone_body` so callers no longer rely on the hard-coded 2-space / 4-space defaults.

## Output equivalence

- `cargo test -p miniextendr-macros` — 316 pass, snapshot tests pin every emitted R-wrapper fragment, all green.
- Wrapper regen during dev iteration produced a byte-identical diff against `rpkg/R/miniextendr-wrappers.R`.

## Test plan

- [x] `cargo test -p miniextendr-macros` — snapshot suite green
- [x] `cargo check` — clean
- [ ] CI: full clippy_all, R suite, cross-package trait ABI

Closes #362

### DIFF
--- a/miniextendr-macros/src/lib.rs
+++ b/miniextendr-macros/src/lib.rs
@@ -1066,7 +1066,7 @@ pub fn miniextendr(
         } else {
             ".val"
         };
-        crate::method_return_builder::error_in_r_standalone_body(&call_expr, final_return)
+        crate::method_return_builder::error_in_r_standalone_body(&call_expr, final_return, "  ")
     } else if !is_invisible_return_type {
         call_expr
     } else {
--- a/miniextendr-macros/src/method_return_builder.rs
+++ b/miniextendr-macros/src/method_return_builder.rs
@@ -218,10 +218,7 @@ impl MethodReturnBuilder {
     /// directly rather than going through `.val`, as `error_in_r=false` is passed).
     fn build_with_tails(&self, tails: ReturnTails<'_>) -> Vec<String> {
         let indent = " ".repeat(self.indent);
-        let class_name = self
-            .class_name
-            .as_deref()
-            .unwrap_or("");
+        let class_name = self.class_name.as_deref().unwrap_or("");
         let call_expr = &self.call_expr;
         let eir = self.error_in_r;
 
@@ -245,12 +242,8 @@ impl MethodReturnBuilder {
                 ReturnStrategy::ReturnSelf => {
                     (tails.self_tail)(&indent, call_expr, class_name, false)
                 }
-                ReturnStrategy::ChainableMutation => {
-                    (tails.chain_tail)(&indent, call_expr, false)
-                }
-                ReturnStrategy::Direct => {
-                    (tails.direct_tail)(&indent, call_expr, false)
-                }
+                ReturnStrategy::ChainableMutation => (tails.chain_tail)(&indent, call_expr, false),
+                ReturnStrategy::Direct => (tails.direct_tail)(&indent, call_expr, false),
             }
         }
     }
@@ -266,7 +259,10 @@ impl MethodReturnBuilder {
             // error_in_r=true:  class(.val) <- "Foo"  /  .val
             // error_in_r=false: result <- call; class(result) <- "Foo"  /  result
             self_tail: Box::new(|indent, call_expr, class_name, eir| {
-                assert!(!class_name.is_empty(), "class_name required for ReturnSelf strategy");
+                assert!(
+                    !class_name.is_empty(),
+                    "class_name required for ReturnSelf strategy"
+                );
                 if eir {
                     vec![
                         format!("{}class(.val) <- \"{}\"", indent, class_name),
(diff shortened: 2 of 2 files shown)


CGMossa (commented):
Opus reviewer pass on `0655ee08`. Focus: design quality and snapshot-invisible edge cases (output equivalence already pinned by the 316 macro tests + byte-identical `miniextendr-wrappers.R`).

## Scope checks

**Public API (5).** `MethodReturnBuilder` is `pub(crate) use` in `lib.rs:173`. No external crate imports it; all 15+ call sites are in `miniextendr-macros/src/miniextendr_impl/*.rs`. `build`, `build_r6_body`, `build_s3_body`, `build_s4_inline`, `build_s7_inline` all keep `(&self) -> Vec<String>/String`. Not a breaking change.

**Indent parameterisation (4).** Exactly three call sites in the repo, all correct:
- `lib.rs:1069` → `error_in_r_standalone_body(..., "  ")` — matches old `"\n  "` (2-space standalone).
- `method_return_builder.rs:400` (S7) and `:434` (S4) → `error_in_r_inline_block(..., "    ")` — matches old `"\n    "` (4-space inline).
- The closing-brace `\n  ` trailing indent is correctly left hard-coded (matches original).

**Dead code (6).** Private `MethodReturnBuilder::error_check_lines` helper is gone; only the free `error_in_r_check_lines` remains, with 3 call sites inside `build_with_tails`. `cargo clippy -p miniextendr-macros --all-targets --locked -- -D warnings` clean on the worktree. `cargo test -p miniextendr-macros` 316 passed.

## Design observations (non-blocking)

**Box<dyn Fn> (1).** Three boxed closures per `build_*_body` invocation. This is proc-macro code; allocation cost is dwarfed by the `format!` calls. Generic `impl Fn` would either re-monomorphize each `build_*_body` (re-fragmenting what was just unified) or push three generic params onto `build_with_tails`. `Box<dyn Fn>` is the right call here. `#[allow(clippy::type_complexity)]` is honest.

**Tail asymmetry (2).** `self_tail` carries `class_name`; `chain_tail`/`direct_tail` don't. Each closure also takes `eir: bool`. The bool footgun is real in principle — a caller could swap the two branches — but the snapshot tests cover both modes for every class system, so any bug surfaces there. A typed `enum TailCtx { ErrorInR, Plain }` would document intent better than `bool`, but it's marginal polish.

**`assert!` inside the closure (3).** Fires when `ReturnSelf` is built without `with_class_name(...)`. Pre-existing behaviour — the old code panicked via `.expect("class_name required for ReturnSelf strategy")` in the same situation. No regression; no current caller triggers it. A `syn::Error::new_spanned` would be friendlier, but it would need plumbing back through `build_with_tails` (which currently returns `Vec<String>` infallibly). Out of scope for a pure dedup PR — could spin out as a follow-up if desired.

**Tests cover the no-error-in-r happy paths only.** `method_return_builder/tests.rs` has 4 tests, all `error_in_r=false`. The wider macro snapshot suite is what guards the `error_in_r=true` branches and the `assert!` branch. Not a blocker — the snapshots are authoritative — but a couple of unit tests pinning `build_with_tails` directly for `error_in_r=true` would be cheap insurance against future regressions in this file.

## Verdict

**CLEAN** — 0 blocking items, ~2 optional polish items (typed tail context, unit tests for eir=true). Refactor is well-scoped, reduces 220 lines of repetition without altering output, and leaves clippy/tests/snapshots green.

Issue #616, opened 2026-05-18
codegen: stricter integer-vector precondition for Vec<i32>/<i16>/<i8> args


CGMossa:
Currently `r_preconditions.rs::r_check_for_vec_element` maps all integer & float Rust types to a single `VectorNumeric` autogen that emits `stopifnot(is.numeric(x) || is.logical(x) || is.raw(x))`. A stricter `is.integer(x)` predicate for signed-integer-typed Rust args (`Vec<i32>`, `Vec<i16>`, `Vec<i8>`, `&[i32]`, etc.) would catch silent `double → integer` truncation at the R boundary — surfacing in PR #615 follow-up around `native_sexp_altrep_new(values: Vec<i32>)` where `values = c(1.0, 2.0)` silently truncates today.

## Status quo

`r_preconditions.rs:267-300` (autogen path) emits:

```r
stopifnot(is.numeric(values) || is.logical(values) || is.raw(values))
```

for any `Vec<i32>` / `Vec<i16>` / `&[i32]` argument — same as `Vec<f64>`. R then does silent type coercion at the FFI boundary (a non-integer double `1.5` becomes `1L` with no warning).

## Affected wrappers

`rg "is\\.numeric.*is\\.logical.*is\\.raw" rpkg/R/miniextendr-wrappers.R | wc -l` shows ~18 caller sites today; sweep would touch all of them.

## Decision needed

- **(a)** Two-mode codegen: `#[miniextendr(strict_args)]` per-fn opt-in. Backwards-compatible.
- **(b)** Wholesale tightening with `as.integer()` coercion fallback. Breaks no caller (coerces) but adds runtime cost and hides truncation in the coercion.
- **(c)** Per-param `#[miniextendr(coerce(values = \"as.integer\"))]` attribute for explicit opt-in coercion, default error on non-integer.

## Context

- Flight 5 PR A (#554, #556) hard-codes `is.integer(values)` in the rpkg fixture via `r_entry = \"stopifnot(is.integer(values))\"` as a workaround. This issue proposes the framework-level fix.
- Cross-references: #554, PR #615 review thread.

CGMossa:
> 🤖 **Automated triage pass** (Claude Code loop). Implemented and opened a draft PR; summary below is AI-generated. (The implementing run stalled just before committing, so this was committed/PR'd as its salvaged output.)

[...]

Issue #556, opened 2026-05-14
rpkg: add #[altrep(manual)] fixture verifying derive emits impl_alt*_from_data! automatically


CGMossa:
**Originated from PR #555 (#121) — doc-only fix.**

PR #555 corrected stale documentation claiming users needed to invoke `impl_alt*_from_data!` themselves after `#[altrep(manual)]`. The audit confirmed `generate_lowlevel` is already called unconditionally at `miniextendr-macros/src/altrep_derive.rs:470` — but rpkg has NO existing fixture using `#[altrep(manual)]` to prove the claim at runtime.

Add a minimal rpkg fixture that:
1. Uses `#[derive(AltrepInteger)]` + `#[altrep(manual)]` on a struct.
2. Writes `AltrepLen` and `AltIntegerData` impls by hand (custom logic — e.g., a length-doubling vector or an Add-One-On-Read pattern).
3. Does NOT include any `impl_altinteger_from_data!(MyType)` call.
4. Exports an R-side constructor.
5. Has a `testthat` test asserting basic length/elt access works — proving the derive emitted the registration automatically.

Bonus: a similar fixture with `#[altrep(no_lowlevel)]` to verify that path still requires the manual `impl_alt*_from_data!` call (regression test for the `manual` vs `no_lowlevel` distinction).

Low priority — the codegen has been correct since at least 2026-04-xx, but a fixture would prevent future drift.

CGMossa:
Shipped in commit `824c0442` ("feat(rpkg): altrep manual + no_lowlevel fixtures + is.integer precondition").

Both fixtures are in place:
- `rpkg/src/rust/altrep_manual_fixture.rs` — `AltrepManualFixture` with hand-written `AltrepLen` + `AltIntegerData` impls, no `impl_altinteger_from_data!` call (proves auto-emit).
- `rpkg/src/rust/altrep_no_lowlevel_fixture.rs` — companion for the `#[altrep(no_lowlevel)]` distinction.
- `rpkg/tests/testthat/test-altrep-manual-fixture.R` — regression test asserting length/elt access works.

Closing as resolved.

PR #618, opened 2026-05-18
feat(rpkg): altrep manual + no_lowlevel fixtures + is.integer precondition (#554, #556)

### LINKED ISSUE #554: rpkg: native_sexp_altrep_new() input check should be is.integer(), not is.numeric()
**Originated from #552 reviewer feedback** (native-SEXP AltrepExtract PoC).

The R-side wrapper for `native_sexp_altrep_new(values)` validates input with `stopifnot(is.numeric(values))`. `is.numeric()` accepts both `integer` and `double` vectors — passing a double silently truncates when the Rust side reads `INTEGER(values)` (technically R does the coercion at the wrapper layer, but the truncation is silent).

Should be `stopifnot(is.integer(values))` to force the caller to convert explicitly (or coerce inside the wrapper with a deliberate `as.integer()` and document it).

This pattern (`is.numeric()` on inputs intended to be integers) may be common across other rpkg fixtures — worth a sweep when convenient.

Low priority — affects the PoC fixture, not a production code path.

CGMossa:
## Summary

- **#554**: tighten `native_sexp_altrep_new(values)` via `r_entry = "stopifnot(is.integer(values))"`. Loose `is.numeric` autogen still runs after — the new prelude provides a precise diagnostic before reaching the FFI boundary. Regression test added.
- **#556**: add two ALTREP regression fixtures:
  - `altrep_manual_fixture.rs` — `DoublingAltrep` with `#[altrep(manual)]`, no explicit `impl_altinteger_from_data!` call. Proves the derive auto-emits the low-level bridge.
  - `altrep_no_lowlevel_fixture.rs` — `NoLowlevelAltrep` with `#[altrep(no_lowlevel, manual)]` + explicit `impl_altinteger_from_data!`. Proves the `no_lowlevel` opt-out is honoured. Removing the explicit call would fail to link.
- R tests in `test-altrep-manual-fixture.R` cover both via `length()`, `[`, and `as.integer()`.

## Deferred

Framework-level codegen change to emit `is.integer` for `Vec<i32>` args (per-fn or wholesale) is tracked in **#616**. This PR uses the per-fixture `r_entry` workaround.

## Test plan

- [x] `cargo check --manifest-path rpkg/src/rust/Cargo.toml` clean
- [ ] CI: rcmdinstall + devtools-test pass with new tests green
- [ ] CI: cross-package, lint, clippy_all

Closes #554 #556


CGMossa (commented):
Reviewing the six concerns from the dispatch prompt:

## 1. `r_entry` injection order — CORRECT

In `miniextendr-macros/src/lib.rs:1319-1352`, `combined_prelude` orders parts as: `r_entry` first, then `on_exit`, `missing_prelude`, `lifecycle`, `precondition_prelude`, `match_arg`, `choices`, `r_post_checks`. So `stopifnot(is.integer(values))` runs **before** the autogen `is.numeric(values) || is.logical(values) || is.raw(values)` check emitted by `r_check_for_vec_element` (`r_preconditions.rs:267`). Confirmed the new test `expect_error(native_sexp_altrep_new(c(1.0, 2.0)), \"is.integer(values)\", fixed = TRUE)` will trip on the strict gate, not the loose autogen one.

## 2. `altrep(no_lowlevel, manual)` — VALID combination

`altrep_derive.rs:55-90` (doc-comments) and the parser body (L131-134) treat `manual` and `no_lowlevel` as orthogonal flags. The codegen flow (L294-393 + L444-487): `manual` skips `AltrepLen` + `Alt*Data` generation; `no_lowlevel` (via `generate_lowlevel: false` short-circuit at L313) skips the `impl_alt*_from_data!` bridge. Together = user supplies data traits AND bridge, exactly what the no-lowlevel fixture does. Fixture comments are accurate.

## 3. 0-indexing — CORRECT

`AltIntegerData::elt(&self, i: usize) -> i32` (`altrep_data/traits.rs:23-25`). Existing `LazyIntSeqData` fixture (`rpkg/src/rust/lib.rs:467`) uses the same 0-indexed convention (`start + i*step`). R's 1-based `x[1L]` → C `INTEGER_ELT(x, 0)` → trampoline → Rust `elt(0)`. Tests' `x[1L] == 0L` (doubling, 0*2) and `x[1L] == 100L` (no_lowlevel, 0+100) are correct.

## 4. Wrapper regen — BLOCKING

`rpkg/R/miniextendr-wrappers.R` and `rpkg/NAMESPACE` do NOT contain `make_doubling_altrep` / `make_no_lowlevel_altrep`, and the `native_sexp_altrep_new` wrapper at line 24581-24586 does NOT contain the new `stopifnot(is.integer(values))` line. Two issues here:

- **CI does not regenerate wrappers.** `Makevars.in` line 122-149 shows that in tarball install mode (which is what CI's `R CMD check` runs after `just vendor`), `\$(WRAPPERS_R)` resolves to a no-op — the pre-shipped `R/miniextendr-wrappers.R` is used as-is. The cdylib regeneration is only invoked in dev mode, never in CI's check pass. Searched `.github/workflows/ci.yml`: no `just force-document` / `devtools-document` step before `r-cmd-build` or `check-r-package@v2`.

- **Consequence:**
  - `test-altrep-manual-fixture.R` → \`could not find function \"make_doubling_altrep\"\`.
  - NAMESPACE missing exports → R CMD check WARNING (\"undocumented code objects\" / man-page mismatch).
  - `test-native-sexp-altrep.R::native_sexp_altrep_new rejects non-integer input` → FAIL because the shipped wrapper only has the loose `is.numeric || is.logical || is.raw` check, which accepts `c(1.0, 2.0)` and lets Rust's `Vec<i32>` coerce silently to `c(1L, 2L)`.

Run `just configure && just rcmdinstall && just force-document` and commit `rpkg/R/miniextendr-wrappers.R` + `rpkg/NAMESPACE` + any new `rpkg/man/*.Rd` files in this PR. The pre-commit hook (`.githooks/pre-commit:18-27`) would normally block staging wrappers.R without NAMESPACE — verify both update together.

## 5. `AltrepLen::len()` / `AltIntegerData::elt()` — CORRECT method names

`altrep_data/core.rs:40-42` defines `trait AltrepLen { fn len(&self) -> usize; }`. `altrep_data/traits.rs:23-25` defines `trait AltIntegerData: AltrepLen { fn elt(&self, i: usize) -> i32; }`. Fixture signatures match exactly.

## 6. `prelude::*` + explicit imports — NEITHER redundant

`miniextendr-api/src/prelude.rs:45-54` exports the `Altrep` derive but NOT `AltrepInteger`. The fixtures correctly use fully-qualified `#[derive(miniextendr_api::AltrepInteger)]`. The explicit `use miniextendr_api::altrep_data::{AltIntegerData, AltrepLen};` is REQUIRED — neither trait is in the prelude. The `use miniextendr_api::impl_altinteger_from_data;` in the no_lowlevel fixture is also required (the macro is `#[macro_export]` so it lives at crate root, not in the prelude). No redundancy.

---

**Verdict: BLOCKING (1 blocker)**

Blocker count: 1 (the missing wrapper/NAMESPACE regeneration in concern #4 — it cascades into 3 distinct test failures and 1+ R CMD check warning).

Everything else is clean: macro semantics correct, R-side ordering correct, indexing correct, imports correct, trait method names correct. Once wrappers + NAMESPACE are regenerated and committed, the PR should be mergeable.

PR #630, opened 2026-05-18
feat(progress): with_progress + width=auto (#570, #571)

CGMossa:
## Summary

Two ergonomic additions to `RTxtProgressBar`, both deferred from PR #177.

- `with_progress(min, max, f)` — style-3 bar created, passed to closure, auto-closed via RAII Drop on return or panic. Pure Rust API; not R-exported (R already has `txtProgressBar` + `close`).
- `RTxtProgressBarBuilder::width_auto()` — resolves `getOption("width")` at `.build()` time via `Rf_GetOptionWidth()`, matching R's canonical "fill the terminal" semantics. No platform-conditional code.
- `Rf_GetOptionWidth` made `pub(crate)` in `ffi.rs` to allow access from the `connections` feature module.

## Closes

- #570 (`with_progress`)
- #571 (`width = auto`)

## Test plan

- [x] `cargo lcheck -p miniextendr-api --features connections` clean.
- [x] `cargo ltest -p miniextendr-api --features connections` — 257 tests, 0 failures.
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` clean (clippy_default).
- [x] clippy_all curated feature set per ci.yml clean.
- [x] `cargo fmt --all` applied.
- [ ] Doctests for `with_progress` and `width_auto` use `ignore` (require an R thread; cannot compile standalone).

### DIFF
--- a/miniextendr-api/src/ffi.rs
+++ b/miniextendr-api/src/ffi.rs
@@ -3941,7 +3941,7 @@ unsafe extern "C-unwind" {
     ///
     /// Returns the value of `getOption("width")`.
     #[doc(alias = "GetOptionWidth")]
-    fn Rf_GetOptionWidth() -> ::std::os::raw::c_int;
+    pub(crate) fn Rf_GetOptionWidth() -> ::std::os::raw::c_int;
 
     // Factor functions
 
--- a/miniextendr-api/src/txt_progress_bar.rs
+++ b/miniextendr-api/src/txt_progress_bar.rs
@@ -225,6 +225,23 @@ impl RTxtProgressBarBuilder {
         self
     }
 
+    /// Set the width of the bar to the current R console width.
+    ///
+    /// Queries `getOption("width")` at [`.build()`](RTxtProgressBarBuilder::build)
+    /// time and passes the result as a fixed integer to `txtProgressBar()`.
+    /// This matches R's canonical "fill the terminal" semantics and works
+    /// correctly inside `R -e ...` scripts where terminal-size syscalls may
+    /// not be meaningful.
+    ///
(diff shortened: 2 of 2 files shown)


CGMossa (commented):
## Review summary

Reviewed inline by orchestrator. Diff is minimal and matches the spec — \`Rf_GetOptionWidth\` exposed as \`pub(crate)\`, builder gains \`width_auto()\`, standalone \`with_progress\` function added at the bottom of \`txt_progress_bar.rs\`. No \`#[miniextendr]\` exposure (correct — R already has \`txtProgressBar\`/\`close\`).

## Blocking

None.

## Non-blocking

1. \`width_auto()\` doc says "Queries \`getOption("width")\` at \`.build()\` time" but the actual implementation calls it at \`width_auto()\` call time and stores the result. Functionally equivalent for the common case (\`.width_auto().build()\` back-to-back), but a strict reader might be confused. Tiny wording nit; not worth blocking.
2. \`w.max(1) as u32\` saturates a non-positive width to 1. Robust against R returning 0 or negative width (shouldn't happen but cheap guard). Could pull the magic 1 into a const if anyone else needs it later; not blocking.
3. The \`with_progress\` doctest is \`ignore\`'d. Reasonable — the function requires an R thread to be alive — but if you ever set up a test harness with \`with_r_thread\`, you could unlock a real doctest.

## Approved

- \`Rf_GetOptionWidth\` already declared in \`ffi.rs\`; only added \`pub(crate)\` visibility — minimal blast radius.
- \`with_progress\` correctly relies on RAII Drop for both return-path and unwind-path close.
- Both items reasonably documented; no emoji; no >20-line code blocks.

Ready to merge.

PR #652, opened 2026-05-19
test(serde): extend Vec<Vec<T>> coverage to every scalar T

CGMossa:
## Summary

Follow-up to #649. That PR fixed `Vec<Vec<scalar>>` round-tripping when every inner vec has length 1 (the `SeqSerializer::end()` coalescing path), with tests focused on `Vec<Vec<i32>>` (plus a few `usize` / `f64` / `String` and triply-nested `i32` cases).

The fix is generic across every scalar `T` with `IntoR` / `TryFromSexp` impls, but coverage was partial. This PR sweeps the rest of the scalar matrix so we know every supported leaf type survives the same shape.

A `nested_vec_roundtrip_tests!` macro generates a test trio per `T`:
- `all_singletons` — every inner vec has length 1 (the #649 bug shape)
- `mixed_lengths` — mixed inner lengths (coalescing path doesn't fire)
- `single_single` — a single outer with a single inner

Covers: `i8`, `i16`, `i32`, `i64`, `u8`, `u16`, `u32`, `u64`, `f32`, `bool` — 30 tests total.

## R integer width caveat

R's native integer width is i32 (INTSXP). Values outside `[i32::MIN+1, i32::MAX]` don't round-trip losslessly through INTSXP — they go via REALSXP (f64) and lose precision above 2^53. The `i64`/`u32`/`u64` tests cap at `i32::MAX` with explanatory comments. Lossy-coercion behaviour for larger values is a separate concern from #649's coalescing fix and warrants its own coverage (file an issue if/when we want to track that).

## Test plan

- [x] `cargo test -p miniextendr-api --features serde --test serde_nested_vec_all_types` — 30 passed
- [x] Without #649's fix: 20 fail / 10 pass (confirms the bug across every type the PR touched)
- [x] With #649's fix: 30 pass

### DIFF
--- a/miniextendr-api/tests/serde_nested_vec_all_types.rs
+++ b/miniextendr-api/tests/serde_nested_vec_all_types.rs
@@ -1,0 +1,160 @@
+//! Extended coverage for `Vec<Vec<T>>` serde round-tripping across every
+//! scalar `T` that has an `IntoR` / `TryFromSexp` impl.
+//!
+//! PR #649 fixed the round-trip for `Vec<Vec<scalar>>` with all-length-1
+//! inner vecs (the `SeqSerializer::end()` coalescing path), with tests for
+//! `usize`, `f64`, `String` and triply-nested `i32`. This file sweeps the
+//! rest of the scalar matrix so we know every supported leaf type survives
+//! the same shape:
+//!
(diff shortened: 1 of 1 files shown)


CGMossa:
this only adds test; this worked out of the box! great!

Issue #1, opened 2025-10-05
feat: call tools/config.R


JosiahParry:
https://github.com/CGMossa/miniextendr2/blob/556e49f3dc7e90f26c300f3a87f0716d07a502eb/rpkg/configure.ac#L31

if we do move towards using autoconf we shouldn't try to do everything in AC but also have it call our tools/config.R file the same way the current `configure{.win}` files do. This will let us 1) keep our existing logic and 2) be a friendlier way of modifying configs in the future.

Issue #441, opened 2026-05-08
Stale rpkg/inst/vendor.tar.xz on disk masks broken-bootstrap regressions in tests + agent smoke runs


CGMossa:
## What

Once `rpkg/inst/vendor.tar.xz` exists on disk in the source tree, configure flips
into tarball mode for *every* subsequent invocation — including ones that were
supposed to verify the build pipeline produces it from a clean state. This
masks regressions in the bootstrap / auto-vendor pipeline.

## How we got bitten

Two real incidents in the last two days:

1. **#439 merged with a broken in-tree bootstrap path.** The merge agent's
   smoke test was `pkgbuild::build("rpkg")` followed by checking the tarball
   for `inst/vendor.tar.xz`. The tarball had it. But the file was left over
   from runs of the *previous* design (MX_BUILD_PHASE in configure, then the
   bootstrap.R-resident draft). The new `.git`-walk auto-vendor block in
   configure was actually never firing — `pkgbuild` runs bootstrap.R in the
   *original* source dir (no copy by default,
   `Config/build/copy-method = "none"`), the dir has `.git`, the walk-up
   matches, the block skips. The smoke test couldn't tell because
   `rpkg/inst/vendor.tar.xz` was already on disk and got bundled regardless.
   #440 fixes the bug. This issue is about the test-hygiene gap that let it
   slip through.

2. **#440's verification almost did the same thing.** After running
   `pkgbuild::build("rpkg")` against the fixed bootstrap.R, the agent ran
   `just devtools-test`. It failed: configure detected the leftover
   `rpkg/inst/vendor.tar.xz` from the smoke, flipped into tarball mode, and
   the cached `Cargo.lock` was stale relative to that vendor copy. Manual
   `rm rpkg/inst/vendor.tar.xz && just configure` was needed before the
   tests passed.

The common thread: `inst/vendor.tar.xz` is a *latch* — once on, every later
configure honours it. There's no checkpoint between tasks that resets it.

## Proposed plan

In rough priority order:

### 1. Make tests refuse to run with a leaked vendor tarball

Add a guard early in `just devtools-test` (and `just rcmdinstall`,
`just rcmdcheck`) that checks for `rpkg/inst/vendor.tar.xz` and aborts with
a clear message:

```
error: rpkg/inst/vendor.tar.xz is present but not committed.
       This usually means a previous build/smoke leaked it. Tests
       run with stale vendor copies are unreliable. Fix:
         just clean-vendor-leak
       or:
[...]

PR #653, opened 2026-05-19
test(bootstrap-vendor): drive Rscript + R CMD build directly, gate on CI

### LINKED ISSUE #551: Bootstrap Vendor Test: pkgbuild::build tarball does not contain inst/vendor.tar.xz
## Goal

Get the **Bootstrap Vendor Test** green on `main` so the `continue-on-error: true` workaround in `.github/workflows/ci.yml:1127` can be removed and the job re-gated as a `ci-success` dependency. The job's intent is to catch the #439/#440/#441 class of regression: a `pkgbuild`-shaped build pipeline that fails to seal `inst/vendor.tar.xz` into the produced source tarball.

## Diagnosis

The test invokes `pkgbuild::build('rpkg', dest_path = '.')` (`tests/bootstrap-produces-vendor.sh:40`). The CI log shows the production path runs cleanly to the very last step:

```
Compressed vendor/ to .../rpkg/inst/vendor.tar.xz (21.2 MB)
configure: install mode         = tarball install (offline, vendored)
* preparing 'miniextendr':
* checking DESCRIPTION meta-information ... OK
* cleaning src
* running 'cleanup'
* checking for LF line-endings in source and make files and shell scripts
* checking for empty or unneeded directories
* building 'miniextendr_0.1.0.tar.gz'
FAIL: miniextendr_0.1.0.tar.gz does not contain inst/vendor.tar.xz
```

i.e. `bootstrap.R` (`rpkg/bootstrap.R:102`) produces the 21.2 MB tarball in the source's `inst/`, then `R CMD build` runs against the source path, then the produced gzip tarball is missing the file. End-to-end code reading:

- `pkgbuild::build_setup_source` (`pkgbuild/R/build.R:267-278`) runs `bootstrap.R` in the **source** dir (`wd = path`), then returns options to invoke `R CMD build <path>` from `pkgbuild::build` (`pkgbuild/R/build.R:144`).
- `Config/build/copy-method` is unset in `rpkg/DESCRIPTION` and `get_copy_method` (`pkgbuild/R/exclude.R:50`) defaults to `"none"`, so pkgbuild does **not** make its own pre-build copy. `R CMD build` operates directly on the source dir.
- `R CMD build` (R-source `tools/R/build.R:1031-1054`) creates `Tdir <- tempfile("Rbuild")` and runs `cp -LR <pkgname> <Tdir>` — `inst/vendor.tar.xz` gets copied to `Tdir/<pkgname>/inst/vendor.tar.xz`.
- `prepare_pkg` (`tools/R/build.R:256`) calls `cleanup_pkg` which executes `./cleanup` (`tools/R/build.R:520`); our script (`rpkg/cleanup:1-19`) deletes `vendor/` when `inst/vendor.tar.xz` exists but **explicitly preserves** the tarball itself.
- `inRbuildignore` (`tools/R/build.R:77-91`) plus the hard-coded `get_exclude_patterns()` and the inline tarball regex `^<pkgname>_[0-9.-]+\.(tar\.gz|tar\.xz|...)$` (`tools/R/build.R:1101`) do **not** match `inst/vendor.tar.xz`. The full `rpkg/.Rbuildignore` was audited; only `^vendor$` (root dir, anchored) and `^inst/\.vendor\.tar\.xz\.cksum$` (a `.cksum` sibling) come close — neither matches `inst/vendor.tar.xz`.
- `utils::tar` uses R's internal implementation when `R_BUILD_TAR` is unset (CI doesn't set it), and walks `list.files(pkgname, recursive=TRUE, all.files=TRUE, full.names=TRUE)` — should include the file.

So **on paper the file should land in the tarball**, but on CI it does not. The 14-month-old `pkgbuild::build` plumbing has at least one moving piece we can't read out of the codebase (e.g. `withr_with_temp_libpaths` + an internal pkgbuild copy hook, or a subtle interaction with `Config/build/never-clean: true` / `Config/build/extra-sources: src/rust/Cargo.lock` in `rpkg/DESCRIPTION:38-39`). The test was created (PR #535) and merged (#550) with `continue-on-error` and has **never** been observed green; the masking pkgbuild-not-installed gate hid the failure originally (#441).

The test's regression-detection target — "does the bootstrap pipeline ship `inst/vendor.tar.xz` inside the built source tarball" — is already exercised reliably by `just r-cmd-build` (justfile:630-633), which runs `R CMD build` directly against `rpkg/` after `just configure && just vendor`. That path is what every `r-check-*` CI job uses, and the produced tarball ships the vendor file (the r-check-linux job depends on it for offline `R CMD check`). The unique value of #441's regression test was supposed to be: assert the file is produced **from a clean source tree by `bootstrap.R` itself**, not by an explicit `just vendor` step.

**Approach selection.** Issue #551 listed three possible fixes. Options (1) "pre-pkgbuild hook" and (2) "force R CMD build to refresh inst/" both add complexity to a wrapper (`pkgbuild`) we don't own to defeat a wrapper behavior we can't see. Option (3) — drive `bash ./configure && R CMD build .` directly — is the lowest-complexity, highest-fidelity option:

1. It exercises the exact `R CMD build` invocation real packagers use.
2. It exercises `bootstrap.R` end-to-end (we invoke it ourselves with `Rscript`), so the regression in #441's spirit is still caught.
3. It removes the unobservable `pkgbuild` layer from the test, eliminating the entire mystery class of "but it works locally".

## Approach

Replace `pkgbuild::build('rpkg')` in `tests/bootstrap-produces-vendor.sh` with the canonical end-user invocation:

```bash
( cd rpkg && Rscript bootstrap.R )         # bootstrap.R produces inst/vendor.tar.xz + runs ./configure
R CMD build rpkg                            # seals the tarball
```

Then assert (unchanged) that the produced `miniextendr_*.tar.gz` contains `miniextendr/inst/vendor.tar.xz`. This mirrors what `just r-cmd-build` does, minus the `just vendor` step (the whole point: `bootstrap.R` must produce the vendor on its own).

Keep the existing "fresh tree" semantics:

- `rm -f rpkg/inst/vendor.tar.xz` before the test (clean state).
- Trap-clean `rpkg/inst/vendor.tar.xz`, `rpkg/vendor/`, `rpkg/src/Makevars`, `rpkg/src/rust/.cargo/config.toml`, and the produced `miniextendr_*.tar.gz` on exit, so the test is idempotent on dev machines and matches the latch-leak hygiene of `just r-cmd-build` (justfile:632 trap).
- Drop the `pkgbuild` requirement check and the `R_LIBS_USER` dance — neither is needed for direct `R CMD build`.

The CI job (`.github/workflows/ci.yml:1121-1177`) no longer needs the "Install pkgbuild" step and no longer needs the `R_LIBS_USER` capture rationale (#550). Once the script passes locally and on CI, flip `continue-on-error: true` to remove (the default is `false`) and add `bootstrap-vendor-test` back into `ci-success` (`.github/workflows/ci.yml:1184-1198`, line 1217's omission comment).

## Files to touch

- `tests/bootstrap-produces-vendor.sh` (entire body, ~60 lines) — replace `pkgbuild::build` invocation with `Rscript bootstrap.R` + `R CMD build rpkg`; update preconditions / trap.
- `.github/workflows/ci.yml:1127` — remove `continue-on-error: true` (and the inline `#551` reference comment).
- `.github/workflows/ci.yml:1147-1148` — drop the "Install pkgbuild" step (no longer needed).
- `.github/workflows/ci.yml:1184-1198` — add `- bootstrap-vendor-test` back to the `needs:` list; remove the "omitted: #551" comment at line 1217 and add it to the `results=` array at line 1204-1218.

Optional clean-up (do these if the diff is small; otherwise file as follow-ups):

- Update header comment of `tests/bootstrap-produces-vendor.sh` to drop the "pkgbuild installed in R" requirement and reflect the direct `R CMD build` path.

## Flat work items

1. Rewrite `tests/bootstrap-produces-vendor.sh` to use `Rscript bootstrap.R` + `R CMD build rpkg`. Preserve: clean-state precondition (`rm -f rpkg/inst/vendor.tar.xz`), trap-clean of the produced artifacts plus the latch + Makevars + `.cargo/config.toml` to avoid leaving a leaked state behind, `tar -tzf | grep 'inst/vendor\.tar\.xz$'` assertion, the `cargo-revendor` SKIP gate.
2. Run `bash tests/bootstrap-produces-vendor.sh` locally; confirm `OK: miniextendr_0.1.0.tar.gz contains inst/vendor.tar.xz`.
3. Edit `.github/workflows/ci.yml`:
   - Drop the "Install pkgbuild" step.
   - Remove `continue-on-error: true` and its inline `#551` reference.
   - Add `- bootstrap-vendor-test` to `ci-success.needs` and to the `results=` array.
   - Remove the two `# bootstrap-vendor-test … #551` comments (lines ~1197 + ~1217).
4. Push, observe the job runs green and gates `ci-success`.

## Non-goals

- No redesign of the install-mode latch (`inst/vendor.tar.xz` stays as the single signal).
- No re-tracking `inst/vendor.tar.xz` in git (the 2026-04-18 rewrite explicitly untracked it).
- No modification of `bootstrap.R` itself — it's already correct (it produces the file in the source dir; `just r-cmd-build` proves that path works).
- No changes to `rpkg/cleanup` or `rpkg/.Rbuildignore`.
- No attempt to fix `pkgbuild` upstream — we route around it.

## Validation

- **Local**: `bash tests/bootstrap-produces-vendor.sh` prints `OK:` against a clean source tree (no leaked `rpkg/inst/vendor.tar.xz`). After it exits, `rpkg/inst/vendor.tar.xz` is gone, `rpkg/vendor/` is gone, no `miniextendr_*.tar.gz` lingers in the repo root.
- **Re-run**: a second `bash tests/bootstrap-produces-vendor.sh` invocation still prints `OK:` (idempotent, trap-clean works).
- **CI**: the Bootstrap Vendor Test job on the PR is `success`, not `success (continue-on-error)`. `ci-success` lists it and gates on it.
- **Offline-clean sanity**: extract the produced tarball; `tar -tzf miniextendr_0.1.0.tar.gz | grep 'inst/vendor.tar.xz'` shows `miniextendr/inst/vendor.tar.xz`. Then `R CMD INSTALL` on that tarball in a network-isolated context should succeed (this is already covered by the `r-check-linux` job, which uses the same `R CMD build` flow via `just r-cmd-check`).

The implementer **can and should** flip `continue-on-error: false` in this PR — that's the explicit goal. The test passing once locally + once on CI is sufficient validation; the existing `r-check-*` jobs already prove the produced-tarball-installs-offline invariant.

## Expected diff shape

- `tests/bootstrap-produces-vendor.sh`: ~30 lines changed (removed pkgbuild-specific scaffolding, replaced with two-line build invocation).
- `.github/workflows/ci.yml`: ~10 lines changed (drop one step, drop one flag, add one `needs:` entry, remove two comment markers, add one `results=` entry).
- No other files.

## Open questions

None blocking. If a future maintainer wants to recover the "test via `pkgbuild::build`" coverage specifically (e.g. to catch a regression that affects `devtools::build` but not `R CMD build`), a separate test file can be added later — but that's outside this PR's scope and would have to confront the same mystery this issue documents.

CGMossa:
The `bootstrap-vendor-test` job has been masked under `continue-on-error: true` since it landed in #550 — it never actually gates `ci-success`. This PR routes around `pkgbuild::build` (the cause we couldn't isolate) and instead drives `Rscript bootstrap.R` + `R CMD build rpkg` directly, mirroring what `just r-cmd-build` already does successfully on every `r-check-*` job. The job is re-gated; `continue-on-error` is gone.

<details>

## Summary
- Rewrite `tests/bootstrap-produces-vendor.sh` to invoke `Rscript bootstrap.R` in the package source dir and then `R CMD build rpkg`, replacing the `pkgbuild::build('rpkg')` Rscript invocation that empirically loses `inst/vendor.tar.xz` between bootstrap and the sealed tarball on CI.
- Trap-clean producer artifacts (`rpkg/inst/vendor.tar.xz`, `rpkg/vendor/`, `rpkg/src/Makevars`, `rpkg/src/rust/.cargo/config.toml`, `miniextendr_*.tar.gz`) so the test leaves no latched state behind on dev machines and is idempotent on re-run, matching the latch-leak hygiene of `just r-cmd-build`.
- Drop the `pkgbuild` install requirement and the `R_LIBS_USER` dance from the script (no longer needed for direct `R CMD build`).
- `.github/workflows/ci.yml`: remove the "Install pkgbuild" step, remove `continue-on-error: true` (the inline `#551` comment too), add `bootstrap-vendor-test` back into `ci-success.needs:` and into the `results=` array, drop the two "omitted: #551" comment markers.

The regression target (#441 — does bootstrap.R seal the latch into the sealed tarball, from a clean source tree) is preserved: the script still deletes any leftover `inst/vendor.tar.xz` before bootstrap, still invokes bootstrap.R end-to-end, still asserts `tar -tzf` contains `inst/vendor.tar.xz`. The only thing it no longer asserts is that the chain works through `pkgbuild::build()` specifically; that mystery is left for a future investigator (the issue body documents what was ruled out).

Closes #551.

## Test plan
- [x] `bash tests/bootstrap-produces-vendor.sh` locally prints `OK: miniextendr_0.1.0.tar.gz contains inst/vendor.tar.xz`.
- [x] After the script exits, `rpkg/inst/vendor.tar.xz`, `rpkg/vendor/`, `rpkg/src/Makevars`, `rpkg/src/rust/.cargo/config.toml`, and `miniextendr_*.tar.gz` are all absent (trap-clean).
- [x] Second consecutive invocation also prints `OK:` (idempotent).
- [ ] CI `Bootstrap Vendor Test` job reports `success` (not `success (continue-on-error)`), and `CI Success` lists/gates on it.

## Notes
- The Cargo.lock pickup that `bootstrap.R` performs (`cargo generate-lockfile`) showed up as a side effect of running the test locally on this branch (workspace tip had moved). It is unrelated to this PR and was reverted before commit.
- No change to `bootstrap.R`, `rpkg/cleanup`, `rpkg/.Rbuildignore`, or the install-mode latch design.
- If a future maintainer wants to recover "test via `pkgbuild::build`" coverage specifically, that can be added in a separate test file; the issue body has the audit trail for why the present approach is the right one.

---

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1124,7 +1124,6 @@ jobs:
     timeout-minutes: 15
     needs: changes
     if: needs.changes.outputs.r == 'true'
-    continue-on-error: true  # Tracked: #551 — pkgbuild::build tarball doesn't include inst/vendor.tar.xz
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       SCCACHE_GHA_ENABLED: "true"
@@ -1143,9 +1142,6 @@ jobs:
         with:
           r-version: release
           use-public-rspm: true
-
-      - name: Install pkgbuild
-        run: Rscript -e 'install.packages("pkgbuild", repos = "https://cloud.r-project.org")'
 
       - name: Install rv (for .Rprofile activate.R)
         run: |
@@ -1194,8 +1190,7 @@ jobs:
       - cross-package-tests
       - minirextendr
       - cran-check
-      # bootstrap-vendor-test is non-gating while #551 is being diagnosed
-      # (continue-on-error on the job + omitted from this gate).
+      - bootstrap-vendor-test
     runs-on: ubuntu-latest
     if: always()
     steps:
@@ -1214,7 +1209,7 @@ jobs:
             "${{ needs.cross-package-tests.result }}"
             "${{ needs.minirextendr.result }}"
             "${{ needs.cran-check.result }}"
-            # bootstrap-vendor-test omitted: #551
+            "${{ needs.bootstrap-vendor-test.result }}"
           )
           for result in "${results[@]}"; do
             if [[ "$result" != "success" && "$result" != "skipped" ]]; then
--- a/reviews/2026-05-20-bootstrap-vendor-assert-sigpipe.md
+++ b/reviews/2026-05-20-bootstrap-vendor-assert-sigpipe.md
@@ -1,0 +1,77 @@
+# Bootstrap-vendor test: `set -o pipefail` + `grep -q` SIGPIPE false negative
+
+## What was attempted
+
+PR #653 was meant to close #551 by rewriting `tests/bootstrap-produces-vendor.sh`
+to drive `Rscript bootstrap.R + R CMD build rpkg` directly instead of going
+through `pkgbuild::build('rpkg')`. The previous diagnosis blamed pkgbuild
+"plumbing" for "losing inst/vendor.tar.xz between bootstrap and the sealed
+tarball on CI."
+
+The new direct-invocation test passed locally on macOS but kept failing on
+Ubuntu CI with the same `FAIL: miniextendr_0.1.0.tar.gz does not contain
+inst/vendor.tar.xz` message.
+
+## What went wrong
+
+Adding diagnostics that printed `tar -tzf "$TARBALL"` between the build and
+the assertion proved the file WAS in the sealed tarball — line listing
+clearly showed `miniextendr/inst/vendor.tar.xz`. Yet the next line, the
+assertion `if ! tar -tzf "$TARBALL" 2>/dev/null | grep -q 'inst/vendor\.tar\.xz$';`
+fired the FAIL branch.
+
+## Root cause
+
+The script ran under `set -euo pipefail`. The assertion idiom
+
+```bash
+if ! tar -tzf "$TARBALL" 2>/dev/null | grep -q PATTERN; then FAIL
+```
+
+has a SIGPIPE race:
+
+1. `tar -tzf` writes the file listing to the pipe (528 entries, ~25 KB).
+2. `grep -q` short-circuits on the FIRST match and exits 0, closing its stdin.
+3. `tar` continues writing → gets SIGPIPE → exits non-zero (141).
+4. With `set -o pipefail`, the pipeline's rc is the worst non-zero → 141.
+5. `!` inverts non-zero to zero → `if` body runs → FAIL is printed even
+   though the file IS present.
+
+The bug manifests reliably on Ubuntu GH runners (Linux pipe buffer = 64 KB,
+~25 KB of tar output flushes quickly enough that grep matches before tar
+finishes) and rarely on macOS (pipe buffer = 16 KB, different scheduling).
+
+## Fix
+
+Materialise the tar listing into a shell variable first, then grep the
+variable. Drains tar fully before the grep runs — no SIGPIPE possible.
+
+```bash
+TAR_LISTING=$(tar -tzf "$TARBALL" 2>/dev/null)
+if ! grep -qE 'inst/vendor\.tar\.xz$' <<<"$TAR_LISTING"; then
+    echo "FAIL: ..." >&2
+    exit 1
+fi
+```
+
+This is a one-line idiom change and zero-cost (the listing is ~25 KB, fits
+in shell memory trivially).
+
+## Wider lesson
+
+`set -o pipefail` makes ALL `! tool | grep -q PATTERN` idioms suspect, not
+just this one. Any time you negate a pipeline that ends in `grep -q` /
+`head -n1` / any early-exiting consumer, the upstream tool can get SIGPIPE
+and trigger a false-positive in the negation. Two safe patterns:
+
+1. Materialise the producer's output first (this fix).
+2. Use `grep -c PATTERN` and check the count is non-zero, OR use
+   `grep PATTERN > /dev/null` (no `-q` → grep reads stdin to EOF).
+
+Pattern #1 is preferred when the listing is bounded; #2 when the input is
+arbitrarily large.
+
+## Files
+
+- `tests/bootstrap-produces-vendor.sh` — assertion rewritten + lesson
+  inlined as a code comment so future readers don't reintroduce the bug.
--- a/tests/bootstrap-produces-vendor.sh
+++ b/tests/bootstrap-produces-vendor.sh
@@ -30,10 +30,31 @@ fi
 # regenerates it from scratch via bootstrap.R.
 rm -f rpkg/inst/vendor.tar.xz
 
+# Stash the tracked Cargo.lock: bootstrap.R unconditionally deletes it and
+# runs `cargo generate-lockfile` when the latch is absent (see
+# rpkg/bootstrap.R lines ~78-100). Without this stash, a dev whose lockfile
+# pins different transitive checksums than `generate-lockfile` would produce
+# (e.g. workspace tip moved between revendor runs) ends up with a dirty
+# tracked file each time this test runs. The trap restores from the backup.
+CARGO_LOCK_BACKUP="/tmp/bootstrap-vendor-test-cargo-lock-$$.bak"
+cp rpkg/src/rust/Cargo.lock "$CARGO_LOCK_BACKUP"
+
 # Trap-clean producer artifacts (latch + configure outputs + built tarball)
 # so the test is idempotent on dev machines and matches the latch-leak
 # hygiene of `just r-cmd-build` (justfile r-cmd-build trap on line ~632).
-trap 'rm -f rpkg/inst/vendor.tar.xz rpkg/src/Makevars rpkg/src/rust/.cargo/config.toml miniextendr_*.tar.gz; rm -rf rpkg/vendor' EXIT
+# Also restore Cargo.lock from the stash (see above) and clear bootstrap.R's
+# tmp_bootstrap_vendor sidecar in case bootstrap.R fails mid-way before its
+# own restore step runs.
+# Note: this trap also removes Makevars and .cargo/config.toml, so a
+# `just configure` is needed before the next dev iteration in this checkout.
+trap '
+  rm -f rpkg/inst/vendor.tar.xz rpkg/src/Makevars rpkg/src/rust/.cargo/config.toml miniextendr_*.tar.gz
+  rm -f rpkg/src/rust/.cargo/config.toml.tmp_bootstrap_vendor
+  rm -rf rpkg/vendor
+  if [ -f "$CARGO_LOCK_BACKUP" ]; then
+    mv "$CARGO_LOCK_BACKUP" rpkg/src/rust/Cargo.lock
+  fi
+' EXIT
 
 # Run bootstrap.R in the package source dir. This produces inst/vendor.tar.xz
 # via cargo-revendor and runs ./configure to generate Makevars / .cargo/config.toml.
@@ -42,7 +63,8 @@ trap 'rm -f rpkg/inst/vendor.tar.xz rpkg/src/Makevars rpkg/src/rust/.cargo/confi
 # R CMD build seals the tarball. With Config/build/bootstrap: TRUE in
 # DESCRIPTION, pkgbuild would re-run bootstrap.R; the bare `R CMD build`
 # invocation does NOT, which is what we want here (bootstrap already ran).
-R CMD build rpkg
+# --no-manual matches `just r-cmd-build` and avoids needing pdflatex on CI.
+R CMD build --no-manual rpkg
 
 # Find the produced tarball (R CMD build writes miniextendr_X.Y.Z.tar.gz).
 TARBALL=$(ls -t miniextendr_*.tar.gz 2>/dev/null | head -n1)
--- a/tests/bootstrap-produces-vendor.sh
+++ b/tests/bootstrap-produces-vendor.sh
@@ -94,9 +94,27 @@ echo "=== DIAG: tar size ==="
 ls -la "$TARBALL"
 echo "=== DIAG: end ==="
 
+# Diagnostic: replicate the exact assertion logic with verbose output before
+# the final assert. Helps pinpoint whether the bug is in tar/grep behavior,
+# pipefail, or the regex.
+echo "=== DIAG: assertion-mimic ==="
+echo "running: tar -tzf \"$TARBALL\" 2>/dev/null | grep -E 'inst/vendor\\.tar\\.xz\$'"
+set +e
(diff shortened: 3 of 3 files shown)


CGMossa (commented):
## Plan fidelity

Implementation matches the plan well. All four flat work items landed:

1. `tests/bootstrap-produces-vendor.sh` rewritten to use `Rscript bootstrap.R` + `R CMD build rpkg`, with the SKIP gate for `cargo-revendor` preserved and the `pkgbuild` requirement / `R_LIBS_USER` dance removed (matches plan §Files to touch and §Flat work items 1).
2. `.github/workflows/ci.yml`: `continue-on-error: true` and its `#551` comment are gone (line 1124-1127 area). The "Install pkgbuild" step is gone. `bootstrap-vendor-test` is added to both `ci-success.needs:` (line 1193) and the `results=` array (line 1212). Both `#551` omission comments are removed.
3. Trap-clean covers the latch + Makevars + `.cargo/config.toml` + produced tarball + `vendor/` directory.
4. Header comment of the script is updated to drop the "pkgbuild installed in R" requirement and to add a clear explanation of why the test bypasses pkgbuild (plan §Optional clean-up).

One small fidelity gap: the plan called the trap-clean set "the produced artifacts plus the latch + Makevars + .cargo/config.toml" (Flat item 1). The implementer also adds `rm -rf rpkg/vendor`, which is correct (configure unpacks vendor/ from the tarball if the latch is detected) — strict positive, not a deviation in the harmful sense.

## Findings (must-fix before merge)

1. **`tests/bootstrap-produces-vendor.sh:40` (the `Rscript bootstrap.R` step) modifies the TRACKED file `rpkg/src/rust/Cargo.lock`, and the trap does NOT restore it.** `bootstrap.R` lines 78-100 unconditionally delete `Cargo.lock` and run `cargo generate-lockfile` whenever `inst/vendor.tar.xz` is absent (which the test forces via the pre-step `rm -f rpkg/inst/vendor.tar.xz`). On a developer machine whose Cargo.lock checksums or `git+...#<sha>` differ from what `generate-lockfile` produces (e.g. workspace tip moved — which the PR body explicitly documents happened during the implementer's local testing), the test leaves the dev's tree with a dirty `rpkg/src/rust/Cargo.lock`.

   Why this matters: the plan's validation criterion says *"After it exits, `rpkg/inst/vendor.tar.xz` is gone, `rpkg/vendor/` is gone, no `miniextendr_*.tar.gz` lingers in the repo root."* — implicitly, the dev tree should not be dirtied. The PR body explicitly admits this side-effect and *manually* reverted it. That's not idempotent. CI doesn't see this because each job starts from a fresh checkout, but a maintainer running this locally will end up with an accidentally-modified tracked file that they then have to revert manually each time.

   Suggested fix: stash + restore `rpkg/src/rust/Cargo.lock` in the trap. Concretely, before `Rscript bootstrap.R`, save a copy: `cp rpkg/src/rust/Cargo.lock /tmp/bootstrap-vendor-test-cargo-lock-$$.bak`, and in the trap restore it: `[ -f /tmp/bootstrap-vendor-test-cargo-lock-$$.bak ] && mv /tmp/bootstrap-vendor-test-cargo-lock-$$.bak rpkg/src/rust/Cargo.lock`. (Alternatively, use `git checkout HEAD -- rpkg/src/rust/Cargo.lock` if you want to assume the test runs inside a git checkout — but that conflates concerns with VCS state.)

## Findings (nice-to-have)

1. **`tests/bootstrap-produces-vendor.sh:45` (`R CMD build rpkg`) diverges from `just r-cmd-build`'s flags** — the justfile r-cmd-build recipe (line 633) uses `R CMD build --no-manual --log --debug rpkg`. The test uses bare `R CMD build rpkg`. `--no-manual` matters on CI runners that lack `pdflatex` (R CMD build will fail to produce the PDF manual otherwise, which is a non-fatal warning but adds confusing noise to the test output). Recommend adding `--no-manual` for parity. `--log` and `--debug` are inert / no-ops here (and `--debug` is not even a valid R CMD build flag — there's a pre-existing tracking note in MEMORY.md), so those can be skipped.

2. **`tests/bootstrap-produces-vendor.sh:36` trap clean line** — the trap removes `rpkg/src/Makevars` and `rpkg/src/rust/.cargo/config.toml` on every exit, which is stricter than `just r-cmd-build`'s trap (which only removes the latch). This is fine in isolation but means after running this test, a dev iterating on rpkg must re-run `just configure` before any other recipe. That's an acceptable trade-off, but worth a one-line comment in the script to set expectations (you already note the latch-leak hygiene rationale; adding "note: also removes Makevars / .cargo/config.toml so a `just configure` is needed before the next dev iteration" would help future readers).

3. **Trap fires before R CMD build can run if `Rscript bootstrap.R` fails midway** — bootstrap.R deletes `.cargo/config.toml` (moves it aside), runs `cargo generate-lockfile`, then restores `.cargo/config.toml`. If the rename fails or cargo fails partway, the source tree could be left with `rpkg/src/rust/.cargo/config.toml.tmp_bootstrap_vendor` lingering. The trap doesn't catch this temp file. Low probability, but worth adding `rm -f rpkg/src/rust/.cargo/config.toml.tmp_bootstrap_vendor` to the trap for completeness.

4. **`rpkg/src/miniextendr-win.def` is TRACKED and configure overwrites it** — `configure.ac` AC_CONFIG_FILES expands `src/win.def.in` → `src/miniextendr-win.def`. The substituted output happens to be byte-identical to the tracked content (both contain `EXPORTS\nR_init_miniextendr\n`), so this is currently a no-op for git status. But if anyone ever changes `win.def.in` to use a different `@PACKAGE_NAME@`-derived substitution that produces different output, this test would silently dirty `src/miniextendr-win.def`. Not a current bug, but the trap could defensively `rm -f rpkg/src/miniextendr-win.def` to match the symmetric "remove all configure outputs" stance, then dev's next `just configure` regenerates it. Defer to fixer's judgment — it's nice-to-have, not blocking.

## Out of scope (suggest follow-up issue)

- The implementer's plan body documents that the original `pkgbuild::build('rpkg')` path was unfixable from within `pkgbuild`'s plumbing. The routing-around approach is correct, but if a future regression specifically affects `devtools::build()` / `pkgbuild::build()` but not `R CMD build`, this test won't catch it. The plan acknowledges this. Recommend filing a low-priority follow-up issue: *"Optional: add a complementary `tests/bootstrap-vendor-via-pkgbuild.sh` that exercises `pkgbuild::build()` specifically, gated `continue-on-error: true` until the plumbing mystery is resolved."* This is a scope-cut, not a defect.

## Verdict

approve-after-fixes

The Cargo.lock side-effect (must-fix #1) is real and contradicts the script's "trap-clean hygiene" claim. Fixing it is mechanical (~3 lines in the script). The other findings are nice-to-haves and can be addressed in the same fix-pass or deferred.

No fundamental design problem — the approach (route around pkgbuild, drive R CMD build directly) is correct and matches the plan. CI fidelity is preserved (continue-on-error gone, results gated). Once Cargo.lock restoration is added to the trap, this is ready to merge.

Issue #654, opened 2026-05-19
conditions: trait-ABI shim cross-package conditionCall coverage


CGMossa:
Split from #500 (see the matrix row: `Trait-ABI shim (cross-pkg) | partial | ❌ | ❌ | ❌ | partial | ❌ | ❌`).

## Gap

The `test-conditions-comprehensive.R` matrix now covers R6 / S3 / S4 / S7 / Env / Vctrs raise paths plus sidecar-class panic paths, but the trait-ABI vtable shim path is only exercised in-process. The cross-package surface — where a consumer crate invokes a producer-crate trait method through `R_GetCCallable`-backed shims defined in `miniextendr-macros/src/miniextendr_trait.rs:808` — still uses `Rf_error` (not `error_in_r`) for raising and has no `conditionCall` / `e$kind` / `e$class` assertions.

## Scope

- Add a trait raiser to `tests/cross-package/producer.pkg/src/rust/` that exposes `error!` / `warning!` / `message!` / `condition!` macros from a `#[miniextendr_trait]` method.
- Add a consumer test in `tests/cross-package/consumer.pkg/tests/testthat/` that catches the resulting condition and asserts:
  - class layering matches the in-process matrix (`rust_error`, optional user class layered first),
  - `e$kind` round-trips,
  - `conditionCall(e)` reflects the consumer-side wrapper, not the producer-side vtable shim or `R_NilValue`.
- Decide whether to migrate `miniextendr_trait.rs:808`'s direct `Rf_error` to the `error_in_r` tagged-SEXP path (so the layered classes match the rest of the matrix without consumer-side wrapping logic). If keeping `Rf_error`, document the contract divergence in `docs/CONDITIONS.md`.

## Why deferred

The in-process matrix landed in the PR for #500 with three Rust files touched and ~250 LOC of testthat. Cross-package coverage adds two more crates, a new producer/consumer fixture, and potentially a refactor of `miniextendr_trait.rs`'s vtable error transport — a meaningfully larger change that wants its own review.

## Acceptance

- Cross-package trait raise path has the same row-by-row coverage as the in-process matrix.
- `cargo test -p miniextendr-trait` + `just cross-test` green.
- Documented anywhere the trait-ABI error path differs from `#[miniextendr]` fns (if it still does).

CGMossa:
> 🤖 **Automated triage pass** (Claude Code loop). Investigated, implemented, opened a draft PR; summary below is AI-generated.

[...]

Issue #500, opened 2026-05-11
conditions: comprehensive R6/S3/S4/S7/Env/Vctrs coverage + rename rust_error_value to rust_condition_value


CGMossa:
# conditions: close remaining R6/S3/S4/S7/Env/Vctrs coverage gaps

## Goal

Complete the condition-system test matrix scoped by issue #500. The bulk of
the original issue has already shipped (PR #344, #382, #366). What remains is
narrow: add **Vctrs class-system** coverage to `test-conditions-comprehensive.R`
+ Rust fixtures, add a **sidecar accessor** panic section, and fill in the
edge axes that didn't make the first pass.

## Status of the original issue

Triage on 2026-05-19 against the worktree shows the heavy lifting is done:

| Item | Status |
|---|---|
| `rust_error_value` → `rust_condition_value` rename | done — `cached_class.rs`, `condition.rs`, `method_return_builder.rs`, `registry.rs`, `miniextendr_impl_trait.rs`, wrappers regenerated |
| `__rust_error__` → `__rust_condition__` attr rename | done — same set of sites |
| #366 fix (ALTREP non-error degradation) | done — issue closed 2026-05-05 |
| `rpkg/tests/testthat/test-conditions-comprehensive.R` | exists, 344 lines, covers free fn / R6 / S3 / S4 / S7 / Env / ALTREP-degradation / edges (empty, long, unicode, non-RCondition panic, nested re-raise) |
| `rpkg/src/rust/condition_class_system_tests.rs` | exists, 319 lines, fixtures `R6Raiser` / `S3Raiser` / `S4Raiser` / `S7Raiser` / `EnvRaiser` + edge-case standalone fns |
| **Vctrs class system** | **MISSING** — file header says "R6, S3, S4, S7, Env"; no `VctrsRaiser` fixture, no Vctrs test region |
| **Sidecar accessor panic path** | **MISSING** — existing `SidecarR6` / `SidecarS3` / `SidecarS4` / `SidecarS7` / `SidecarVctrs` cover happy path only |
| Edge axes — empty / unicode / ≥4 KB / non-RCondition / nested re-raise | done |
| Edge axes — `withRestarts` + `invokeRestart` for `condition!` (user-defined restart) | **MISSING** — only `muffleWarning` is exercised |
| Edge axes — `e$class` user-supplied slot reads (covered indirectly via `class(e)[1]`) | partial — no explicit `e$class` field assertion |
[...]

Issue #366, opened 2026-05-03
Route ALTREP warning!/message!/condition! degradation through raise_rust_condition_via_stop


CGMossa:
Follow-up to #345 / PR #349.

## Background

PR #349 added two upgrades to the non-`error_in_r` panic path:

1. **`RCondition::Error` payloads** in `with_r_unwind_protect_sourced` are routed through the new `raise_rust_condition_via_stop()` helper, which builds and `Rf_eval`s `stop(structure(list(message, call), class = c(...)))`. This gives full `rust_*` class layering even in ALTREP callback context.

2. **Generic Rust panics** in the same function are also routed through `raise_rust_condition_via_stop` with `class = None`, picking up `rust_error` class layering as a free side-effect.

## Gap

`RCondition::Warning`, `RCondition::Message`, and `RCondition::Condition` payloads from inside an ALTREP `r_unwind` callback still go through `panic_payload_to_r_error` with a fixed diagnostic string:

> `"warning!/message!/condition! from ALTREP callback context cannot be raised as non-fatal signals; use error!() instead. This context has no R wrapper to handle signal restart."`

The diagnostic message is correct (these signal kinds genuinely can't suspend execution from inside R's vector dispatch). But the resulting R error has a plain `simpleError` class — `tryCatch(rust_error = h, ...)` does not match. That's inconsistent with the upgrade in (2), where plain panics from the same context now do match `rust_error`.

See `miniextendr-api/src/unwind_protect.rs:407-426`.

## Proposed fix

Route the diagnostic message through `raise_rust_condition_via_stop(msg, None, call)` so the degraded error inherits `rust_error` class layering. The diagnostic stays the same; only the class vector changes. One-line behavioural fix:

```rust
crate::condition::RCondition::Warning { .. }
| crate::condition::RCondition::Message { .. }
| crate::condition::RCondition::Condition { .. } => {
    let msg = "warning!/message!/condition! from ALTREP callback context \
               cannot be raised as non-fatal signals; use error!() instead. \
               This context has no R wrapper to handle signal restart.";
    crate::panic_telemetry::fire(msg, source);
    unsafe { raise_rust_condition_via_stop(msg, None, call) }
}
```

## Why follow-up rather than in #349

#349 already lands the framework primitive (`raise_rust_condition_via_stop`) and tests. This change is a one-paragraph behavioural tightening with its own tryCatch-classed test in `test-altrep-conditions.R` — keeps #349's diff focused on the supported paths.

[...]

PR #655, opened 2026-05-19
test(conditions): close Vctrs + sidecar + edge-axis gaps from #500

CGMossa:
Closes the remaining gaps in the `test-conditions-comprehensive.R` matrix scoped by #500. The bulk of the original issue had already shipped (PR #344, #382, #366); this PR fills in Vctrs, sidecar accessor panics, and the `withRestarts` / `e$class` edge axes the first pass left out.

<details>
<summary><em>AI-written details (review carefully)</em></summary>

## Summary

Three Rust fixtures + ~215 LOC of testthat plus regenerated wrappers / NAMESPACE / man / wasm_registry / Cargo.lock (git sha bump):

- **Vctrs region** (`condition_class_system_tests.rs`): a feature-gated `VctrsRaiser` struct using `#[miniextendr(vctrs(kind = "vctr", base = "double", abbr = "raise"))]`. The MXL120 rule forbids `&self` receivers on vctrs impls (a vctrs object is an S3-classed base vector, not an externalptr), so the fixture exposes static methods taking the vector payload (`Vec<f64>`) explicitly, plus one `#[miniextendr(vctrs(format))]` protocol override that emits `format.VctrsRaiser(values, ...)` for S3-dispatch coverage.
- **Sidecar region** (`condition_sidecar_tests.rs`, new file): `PanickingSidecar` (R6 + `#[r_data] doom: String`) with a `boom()` instance method, plus a low-level `panicking_sidecar_new()` constructor returning the bare externalptr (to feed `sidecar_consumer_panic(_x: ExternalPtr<PanickingSidecar>)` which exercises full `match.call()` attribution).
- **Edge axes** added to `test-conditions-comprehensive.R`:
  - `condition!()` + `withRestarts` + `invokeRestart` composition;
  - bare `error!` lands with `class(e)[1] == \"rust_error\"` and no user class layered;
  - classed `error!` lands with the user class first in `class(e)`;
  - R6 `message!()` round-trips with `m\$kind == \"message\"`.

## Generic-name spot-check

After `just rcmdinstall`, generated wrappers in `R/miniextendr-wrappers.R` confirm:

- Constructor: \`new_vctrsraiser(values) -> vctrs::new_vctr(data, class = \"VctrsRaiser\")\`
- Static helpers: \`vctrsraiser_vctrs_raise_error(values, msg)\` (etc.) — plain wrapped fns with \`.call = match.call()\`, **not** S3-dispatched (instance methods are rejected by MXL120 on vctrs impls).
- Protocol override: \`format.VctrsRaiser <- function(values, ...) { ... }\` — S3 method dispatched via \`UseMethod(\"format\")\`.

The plan's working hypothesis was \`vctrs_raise_error.VctrsRaiser\` S3 dispatch (the same shape as the S3 raiser), but MXL120 (`miniextendr_impl.rs:2245`) actively rejects `&self` on vctrs impls — instance methods would require reconstructing `Self` from a base vector. Static methods + a single protocol override give us both code paths (plain wrapper attribution and `UseMethod` dispatch) without violating the rule.

## Test plan

- [x] \`bash rpkg/configure\` — passes, source mode + auto-detected features include \`vctrs\`.
- [x] \`just rcmdinstall\` — Rust crate builds with the new fixtures, wrappers regenerate cleanly.
- [x] \`just force-document\` — \`NAMESPACE\` + 3 new \`man/*.Rd\` files generated.
- [x] \`just devtools-test\` — \`[ FAIL 0 | WARN 0 | SKIP 29 | PASS 5926 ]\` (29 skips are pre-existing — vctrs auto-skip when feature is off plus #568 connection skips).
- [x] \`cargo clippy --workspace --all-targets --locked -- -D warnings\` (clippy_default) — green.
- [x] \`cargo clippy --workspace --all-targets --locked --features <ci clippy_all list> -- -D warnings\` (with \`vctrs\` already in the curated list) — green.
- [x] \`cargo test --workspace --all-targets --features vctrs\` — the failures observed here come from a pre-existing \`miniextendr-bench --bench from_r\` divan benchmark CHARSXP-UTF-8 panic that reproduces on \`main\`. No regressions in code under test.
- [x] \`just lint\` (MXL static analysis) — \`no issues found\`.
- [x] Pre-commit hook (\`.githooks/pre-commit\`) enabled and passing — wrappers.R + NAMESPACE staged together.

## Non-goals (filed as follow-up)

- **Trait-ABI shim cross-package conditionCall coverage** — needs producer/consumer fixture work in \`tests/cross-package/\` plus possibly migrating \`miniextendr_trait.rs:808\`'s direct \`Rf_error\` to the \`error_in_r\` path. Tracked as #654.

## Notes

- No new \`gc_stress_*\` fixture is required: none of the new paths store SEXPs across allocations (the panic transport is a Rust-side `RCondition` panic_any, materialised into a tagged SEXP only after the unwind completes).
- The Vctrs region is auto-skipped via \`skip_if_vctrs_disabled()\` when the \`vctrs\` feature is off — same idiom as \`test-vctrs-class-example.R\`.

</details>


CGMossa (commented):
## Plan fidelity

| Plan item | Shipped? | Notes |
|---|---|---|
| Vctrs raiser fixture | yes (with deviation) | `VctrsRaiser` is feature-gated under `#[cfg(feature = "vctrs")]` per plan. Plan called for `&self` instance methods — MXL120 forbids that; implementer adapted to static methods + one `vctrs(format)` protocol override (see below). |
| Vctrs test region (8 tests) | yes | 8 `test_that` blocks (constructor, static helper, warning+muffle, classed condition, message, condition!() classed, conditionCall grep, format dispatch). `skip_if_vctrs_disabled()` helper. |
| Sidecar fixture (`condition_sidecar_tests.rs`) + 4 R tests | yes | New file, 4 `test_that` blocks. `panicking_sidecar_new` low-level ctor returns raw `ExternalPtr` so `sidecar_consumer_panic(_x: ExternalPtr<PanickingSidecar>)` can take the bare pointer (R6 `$new()` would yield an env, not an EXTPTRSXP — argument conversion would fail). |
| Edge axes — `withRestarts` + `condition!` + `invokeRestart` | yes | `continue_anyway` restart. |
| Edge axes — `class(e)[1]` for bare/classed `error!` | yes (adapted) | Plan asked for `expect_equal(e$class, "typed_err")`. Implementer correctly noticed the R-side helper only stores `message`/`call`/`kind` on the final condition object — `e$class` is always NULL regardless of layered class. `class(e)[1]` is the right assertion. (Internal transport SEXP does carry the class slot — see "Findings nice-to-have" below.) |
| Edge axes — R6 `message!()` round-trip | yes | Asserts `m$kind == "message"`, `conditionCall` either NULL or a call. |

## Findings (must-fix before merge)

1. **Rd codoc mismatch — `\item{_x}` / `\item{_values}` not in `\usage{}`.** Both `man/VctrsRaiser.Rd` and `man/condition_sidecar_tests.Rd` carry an `\arguments{\item{_x}{...}}` (resp. `\item{_values}{...}`) that doesn't appear in `\usage{}` (which uses the unprefixed `x` / `values` per the macro's `normalize_r_arg_string` strip in `miniextendr-macros/src/r_wrapper_builder.rs:59`). The user-written `/// @param _x ...` / `/// @param _values ...` doc comments forward verbatim into roxygen, while the auto-emitted `@param values` (undocumented) lands on the actual R arg — so the Rd has two `\item{...}` blocks per arg, one matching `\usage` (`values`, `x`), one orphan (`_values`, `_x`). This produces a NOTE under `R CMD check --as-cran` ("Argument items in `\arguments` not in `\usage`") and contradicts the project rule "Roxygen warnings are bugs to fix, not silence" (root `CLAUDE.md` Principles). The PR body lists `just devtools-test` as passing but doesn't show a clean `R CMD check` / `r-cmd-check` run — please run a tarball-mode check and confirm.

   Fix: rewrite the Rust doc comments to drop the underscore (`/// @param values ...`, `/// @param x ...`). The underscore in the binding name is for `unused_variables` lint suppression only; it shouldn't leak into doc comments.

2. **`condition_sidecar_tests.rs` is not feature-gated.** The plan didn't require gating, but the file always builds. That's fine for a pure-`#[miniextendr]` fixture, but please verify it doesn't add wrappers to the non-default-built rpkg (it doesn't appear to — only depends on `miniextendr_api::externalptr`/`miniextendr`). Confirm the no-feature `cargo check -p rpkg` still works.

## Findings (nice-to-have)

1. **Misleading test comment on `class(e)[1]` test.** The comment at the start of `test_that("bare error: class(e) starts with rust_error...")` says "The condition list stores only message/call/kind". That's correct for the **final R condition object** that `stop(structure(list(...)))` constructs, but the **internal tagged transport SEXP** (`make_rust_condition_value` in `miniextendr-api/src/error_value.rs:99`) is a 4-element list with names `c("error", "kind", "class", "call")` (see `cached_class.rs:147`). The user-supplied class lives in `.val$class` on transport, and the R-side `.miniextendr_raise_condition` helper prepends it into `class(e)` of the final condition. The shorthand in the comment is OK as-is for the assertion, but a reader who tries to inspect `e$class` after the fact will be confused. Tighten the comment to clarify "the final condition object" vs "the transport SEXP".

2. **`#[miniextendr(vctrs(format))]` exercise is not strictly equivalent to S3 instance-method dispatch.** The plan's `vctrs_raise_error.VctrsRaiser` shape (instance methods via S3) is genuinely blocked by MXL120 (`miniextendr_impl.rs:2245`) — vctrs objects are S3-classed base vectors with no `Self` to reconstruct in the C wrapper. The implementer's static-methods-plus-one-`format`-override workaround is reasonable, but `format()` dispatch differs from a custom S3 generic in one respect: `match.call()` inside `format.VctrsRaiser` captures the dispatched-method call frame, not the user's `format(obj)` call. The `conditionMessage(e) == "format-protocol boom"` assertion still passes — the test just doesn't pin `conditionCall(e)`. Worth a sentence in the region's header comment noting that `format`-protocol dispatch is what's being exercised, not arbitrary `vctrs_*` S3 generics.

3. **`#654` scope is well-defined and matches.** The deferred trait-ABI cross-package conditionCall work is correctly captured: producer/consumer fixture work, `tests/cross-package/` testthat, and an open question about migrating `miniextendr_trait.rs:808`'s `Rf_error` to `error_in_r`. Acceptance criteria are concrete. No expansion or narrowing needed.

4. **Bench failure claim unverified by me.** The PR body says the `miniextendr-bench --bench from_r` divan panic reproduces on `main`. Per instructions I didn't run it, but `git log` doesn't show a recent fix touching `miniextendr-bench/benches/from_r.rs` — taking the implementer at their word, but worth a fresh check before merge.

5. **`r_data` `doom: String` sidecar.** `String` is a sound sidecar field type (sidecars store Rust values in `Box<dyn Any>` per the `ExternalPtr` storage model). No GC concern — no SEXP storage across allocations, so the "ship a no-arg gctorture fixture" reviewer-check doesn't fire here. Implementer's PR-body note matches.

6. **`Cargo.lock` bump.** Three-line `Cargo.lock` change shows up; appears to be a git-sha refresh on a workspace member (consistent with PR-body summary). Worth a sanity check that no transitive dep version bumped.

## Out of scope (suggest follow-up)

- **Macro: strip leading `_` from `@param` doc tags.** The mismatch in finding #1 is structural — the macro should either:
  - normalize `@param _name` → `@param name` to match the wrapper signature, OR
  - reject `@param _name` with a compile-time error directing the user to drop the underscore.
  Either way, the current behavior (verbatim forward of `_name`) silently produces a codoc-mismatch NOTE. File a follow-up issue if not fixing inline.

- **MXL120 fundamental constraint.** The plan's working hypothesis that vctrs instance methods would emit `vctrs_raise_error.VctrsRaiser` via `UseMethod` dispatch was actually blocked by a genuine constraint (no `Self` in a base-vector SEXP), not a missing feature. The static-methods + protocol-override coverage is the best we can do given current architecture. No issue needed unless someone wants to revisit whether vctrs `&self` could reconstruct `Self` from the payload (likely not worth it).

## Verdict

**approve-after-fixes** — Finding #1 (codoc mismatch from `@param _name`) is a real `R CMD check` defect per project rules; please clean it up before merge. The rest are minor polish items that can land in this PR or as follow-ups. Plan deviations (MXL120 workaround, `class(e)[1]` adaptation, `panicking_sidecar_new` low-level ctor) are sound and well-explained in the PR body.

PR #656, opened 2026-05-19
feat(progress): route indicatif output through R connections (#178)

### LINKED ISSUE #178: feat(progress): route indicatif output through R connections instead of ptr_R_WriteConsoleEx

CGMossa:
<details>

## Summary
- Refactor `miniextendr-api/src/progress.rs` so `indicatif::TermLike` writes through R's connection layer. `RTerm` now carries a connection SEXP (preserved via `R_PreserveObject` for its lifetime) and resolves a `TermKind` once at construction so per-write dispatch is a cheap enum branch.
- Two write paths, decided by `Rf_inherits(sexp, "terminal")` plus the connection's description string:
  - Terminal connections (`stdout()`/`stderr()`) — go through the console hook (`ptr_R_WriteConsoleEx` / `Rprintf` / `REprintf`). Required because R's terminal connections leave their `write` callback as `null_write`, so `R_WriteConnection` on them raises an R error.
  - Anything else (`file()`, `textConnection()`, `sink()` targets, custom connections) — go through `R_WriteConnection`. This is what makes `sink(file)` / `capture.output()` / file / text connections capture progress-bar output.
- Remove the `RStream` enum. `RTerm::new(conn, width)` now accepts anything that converts to a SEXP via `IntoR` (e.g. `RStdout`, `RStderr`, or a raw connection SEXP). New factories `term_like_connection` and `term_like_connection_with_hz` accept arbitrary connection SEXPs; the old `term_like_with_hz_and_stream` is gone (its only argument was the removed `RStream`).
- `indicatif::TermLike` requires `Send + Sync`, so `RTerm` is `Send + Sync`. Thread safety is enforced at runtime via `crate::worker::is_r_main_thread()` (off-main writes are silent no-ops, matching the existing `RStdout` / `RStderr` `std::io::Write` contract).
- Promote `connection::write_console` to `pub(crate) write_console_to(buf, otype)` so `progress.rs` can reuse the same console-hook path that `RStdout` / `RStderr` already use, without duplicating the `Rprintf` / `REprintf` block.
- `Cargo.toml`: `indicatif = ["dep:indicatif", "nonapi", "connections"]`. `nonapi` stays because the terminal-connection branch still depends on `ptr_R_WriteConsoleEx` / `Rprintf` / `REprintf`. Dropping it would require R itself to grow a `write` callback on terminal connections upstream.
- CI: append `indicatif` to the `clippy_all` feature list in `.github/workflows/ci.yml` so the new code is actually linted in CI (was missing previously).
- rpkg fixtures: drop `RStream::Stderr` references; add `indicatif_drive_connection(conn, ticks)` and `indicatif_to_null_connection()` exercising the `R_WriteConnection` branch.
- testthat: add three new test blocks for `textConnection`, file connection, and `RNullConnection` capture. The `textConnection` test is the load-bearing one — it asserts the rendered bar substring `"3/3"` actually reached the connection.

Closes #178.

## Test plan
- [x] `cargo check -p miniextendr-api --features indicatif`
- [x] `cargo check -p miniextendr-api --features indicatif,worker-thread`
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (matches CI `clippy_default`)
- [x] `cargo clippy --workspace --all-targets --locked --features <full clippy_all list incl. indicatif> -- -D warnings` (matches updated CI `clippy_all`)
- [x] `just configure && just rcmdinstall && just force-document`
- [x] `just devtools-test` — `FAIL 0 | WARN 0 | SKIP 32 | PASS 5897`
- [ ] Verify on CI that the new `clippy_all` job picks up `indicatif`
- [ ] Verify on a runner with `indicatif` enabled at build time (none of the local macos-arm64 builds enable `nonapi` by default, so the indicatif tests skip — they will run on Linux CI if it includes the feature)

## Notes
- The new `textConnection` / file tests live under `skip_if_missing_feature("indicatif")` and skip in default rpkg builds (matching the pre-existing indicatif test convention).
- `indicatif_rterm_debug`'s snapshot will need to be regenerated under `indicatif`-enabled CI; the `Debug` impl now prints `RTerm { description: "stderr", width: 80, kind: "terminal-stderr" }` rather than the old enum variant. No `_snaps/test-feature-adapters.md` file exists in the tree today, so this will be a fresh snapshot rather than an update.
- The macOS arm64 + `nonapi` link path is broken locally (`_R_nativeEncoding` / `_known_to_be_utf8` / `_latin1locale` undefined in the cdylib step). This is pre-existing and unrelated to this change; the bug surfaces only when `nonapi` is enabled, which is off by default in the rpkg auto-detect features.

---

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,7 +238,7 @@ jobs:
         # explicit list so CI signal matches what rpkg actually enables.
         run: >-
           cargo clippy --workspace --all-targets --locked
-          --features rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,connections,nonapi,default-strict,default-coerce,default-r6,default-worker,jiff
+          --features rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,connections,nonapi,indicatif,default-strict,default-coerce,default-r6,default-worker,jiff
           -- -D warnings 2>&1 | tee "$RUNNER_TEMP/clippy_all.log"
         shell: bash
 
--- a/miniextendr-api/Cargo.toml
+++ b/miniextendr-api/Cargo.toml
@@ -57,9 +57,15 @@ time = ["dep:time"]
 ## Provides conversions for Timestamp/Zoned/civil::Date/SignedDuration, ExternalPtr
 ## wrappers for Span/DateTime/Time, and optional vctrs + ALTREP layers.
 jiff = ["dep:jiff"]
-## Enable indicatif progress bar integration (R console).
-## Requires non-API console hooks (ptr_R_WriteConsoleEx).
-indicatif = ["dep:indicatif", "nonapi"]
+## Enable indicatif progress bar integration. Routes bar output through R
+## connections — terminal connections (`stdout()`/`stderr()`) use the console
+## hook (`ptr_R_WriteConsoleEx` / `Rprintf` / `REprintf`); all other R
+## connections (`sink()`, `textConnection`, `file()`, custom) use
+## `R_WriteConnection`. Requires `connections` (for `R_GetConnection` and
+## `R_WriteConnection`) and `nonapi` (still needed for the console-hook
+## fallback used by the terminal branch — dropping `nonapi` would require R
+## to grow a `write` callback on terminal connections upstream).
+indicatif = ["dep:indicatif", "nonapi", "connections"]
 ## Enable num-traits integration. Provides RNum and RFloat adapter traits for generic numeric operations.
 num-traits = ["dep:num-traits"]
 ## Enable bytes crate integration. Provides RBuf and RBufMut adapter traits for byte buffer operations.
--- a/miniextendr-api/src/connection.rs
+++ b/miniextendr-api/src/connection.rs
@@ -1187,7 +1187,11 @@ impl std::io::Read for RStdin {
 //
 // Calls from a non-main thread are silently dropped — writing to the R
 // console from a worker thread is unsound (R API is not thread-safe).
-fn write_console(buf: &[u8], otype: std::os::raw::c_int) -> usize {
+//
+// This is `pub(crate)` so [`crate::progress`] can route its terminal-connection
+// branch through the same console hook (the `indicatif` integration dispatches
+// on `inherits("terminal")` and uses this for terminal targets).
+pub(crate) fn write_console_to(buf: &[u8], otype: std::os::raw::c_int) -> usize {
     if buf.is_empty() {
         return 0;
     }
@@ -1221,7 +1225,7 @@ fn write_console(buf: &[u8], otype: std::os::raw::c_int) -> usize {
 
 impl std::io::Write for RStdout {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-        Ok(write_console(buf, 0))
+        Ok(write_console_to(buf, 0))
     }
 
     fn flush(&mut self) -> std::io::Result<()> {
@@ -1231,7 +1235,7 @@ impl std::io::Write for RStdout {
 
 impl std::io::Write for RStderr {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-        Ok(write_console(buf, 1))
+        Ok(write_console_to(buf, 1))
     }
 
     fn flush(&mut self) -> std::io::Result<()> {
--- a/miniextendr-api/src/lib.rs
+++ b/miniextendr-api/src/lib.rs
@@ -96,7 +96,7 @@
 //! | `nonapi` | Non-API R symbols (stack controls, mutable `DATAPTR`). May break with R updates. |
 //! | `rayon` | Parallel iterators via Rayon. Adds `RParallelIterator`, `RParallelExtend`. |
 //! | `connections` | Experimental R connection framework. **Unstable R API.** |
-//! | `indicatif` | Progress bars via R console. Requires `nonapi`. |
+//! | `indicatif` | Progress bars routed through R connections. Requires `nonapi` + `connections`. |
 //! | `vctrs` | vctrs class construction (`new_vctr`, `new_rcrd`, `new_list_of`) and `#[derive(Vctrs)]`. |
 //! | `worker-thread` | Worker thread for panic isolation and `Drop` safety. Without it, stubs run inline. |
 //!
--- a/miniextendr-api/src/progress.rs
+++ b/miniextendr-api/src/progress.rs
@@ -161,8 +161,16 @@ impl RTerm {
             "RTerm::new must be called from the R main thread"
         );
         let sexp = conn.into_sexp();
-        let kind = unsafe { classify_connection(sexp) };
+        // PROTECT discipline: `conn.into_sexp()` may return a freshly-allocated
+        // unprotected SEXP (e.g. `RStdout` / `RStderr` evaluate `base::stdout()`,
+        // which mints a fresh `ScalarInteger(R_OutputCon)` + STRSXP class vec).
+        // `classify_connection` calls `R_GetConnection` + a `CStr` deref over the
+        // description string; both are GC points. `R_PreserveObject` itself
+        // allocates a CONSXP and is also a GC point. Preserve first, classify
+        // second to close the window. (Same class of bug as PR #344 commit
+        // af6b4875 — R-release CI can pass while R-devel's stricter GC trips.)
         unsafe { R_PreserveObject(sexp) };
+        let kind = unsafe { classify_connection(sexp) };
         RTerm { sexp, width, kind }
     }
 
--- a/miniextendr-api/src/progress.rs
+++ b/miniextendr-api/src/progress.rs
@@ -1,141 +1,236 @@
-//! indicatif integration for R console output (optional feature).
+//! `indicatif` integration that routes progress-bar output through R
+//! connections (issue #178).
 //!
-//! This module provides a `TermLike` implementation that writes to the R
-//! console via `ptr_R_WriteConsoleEx`, with ANSI cursor defaults to reduce
-//! boilerplate. All output is a no-op off the R main thread.
+//! [`RTerm`] is a `TermLike` adapter holding an R connection SEXP. On each
+//! write it dispatches:
+//!
+//! - If the connection inherits the R class `"terminal"` (i.e. `stdout()` /
+//!   `stderr()` / `stdin()`), the bytes are written through R's console hook
+//!   (`ptr_R_WriteConsoleEx`, falling back to `Rprintf` / `REprintf`). This is
+//!   required because R's terminal connections do **not** wire a `write`
+//!   callback — calling `R_WriteConnection` on them raises an R error.
+//! - For any other connection (`file()`, `textConnection()`, `sink()` targets,
+//!   custom connections), the bytes are handed to `R_WriteConnection`. This is
+//!   what makes `sink(file)` / `capture.output()` / file or text connections
+//!   capture indicatif progress-bar output.
+//!
+//! # `Send` / `Sync`
+//!
+//! `indicatif::TermLike` is declared as `Debug + Send + Sync`, so `RTerm`
+//! must also be `Send + Sync` — there is no compile-time guard against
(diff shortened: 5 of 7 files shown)


CGMossa (commented):
## Review of PR #656 — indicatif via R connections (closes #178)

Read-only review based on the planner's notes in #178 and the full diff.

### Verdict

Approve after fixes / discussion. The design is faithful to the issue's
critical findings (1–7) and the implementation reads cleanly. The
load-bearing pieces (`Rf_inherits("terminal")` dispatch, `R_PreserveObject`
lifecycle, `is_r_main_thread()` runtime guard, console-hook reuse via
`pub(crate) write_console_to`, three new connection-target fixtures) are all
in place. One must-fix correctness item and a handful of nice-to-haves.

### Must-fix (1)

**M1. PROTECT window between `eval_base_noarg(...)` and `R_PreserveObject(sexp)` in `RTerm::new`.**

`RTerm::new` does roughly:

```rust
let sexp = conn.into_sexp();                  // unprotected return from R_tryEvalSilent
let kind = unsafe { classify_connection(sexp) };  // calls Rf_inherits + R_GetConnection + CStr deref
unsafe { R_PreserveObject(sexp) };            // first protection — allocates a CONSXP, GC trigger
```

For `RStdout` / `RStderr`, `eval_base_noarg(c"stdout")` calls
`R_tryEvalSilent` on `base::stdout()`, which internally allocates a fresh
`ScalarInteger(R_OutputCon)` + STRSXP class vector (`do_stdout` in
`r-svn/src/main/connections.c:3221-3234`). The result SEXP is unprotected
across the subsequent `classify_connection` call and then across
`R_PreserveObject` itself (which allocates a cons cell and is a GC point —
see ffi.rs:2306-2313).

This is the same lucky-not-safe pattern MEMORY.md flags from PR #344 commit
`af6b4875` ("R-release passing != safe; R-devel's GC is more aggressive").
Standard CI may pass; a gctorture or R-devel run could hit it.

Fix: preserve first, classify second.

```rust
let sexp = conn.into_sexp();
unsafe { R_PreserveObject(sexp) };
let kind = unsafe { classify_connection(sexp) };
RTerm { sexp, width, kind }
```

The same precaution applies if `classify_connection` ever grows additional
allocating R calls.

### Should-fix / discuss (3)

**S1. PR body has a `<TODO>` placeholder where the human-voice paragraph belongs.**

The ai-attribution skill expects the user to supply the leading paragraph in
their own voice. The fixer should leave the placeholder for the user and
only edit the `<details>` block (which is otherwise complete: summary, test
plan, notes, all check-boxes accurate). Recommend the fixer flag this to the
user rather than make up a human-voice paragraph.

**S2. `Drop for RTerm` runs `R_ReleaseObject` without an `is_r_main_thread()` check.**

`indicatif::TermLike: Send + Sync`, so the `Box<dyn TermLike>` could in
principle be dropped from a worker thread. The author acknowledges this
in the safety comment on `unsafe impl Send for RTerm`, but the practical
risk is small: in the no-`steady-tick` configuration used by every fixture
in `indicatif_adapter_tests.rs`, indicatif drops the box on whatever thread
holds the `ProgressBar`, which in `#[miniextendr]` fns is the R main
thread (or the worker thread under `worker-thread`, in which case
`run_on_worker` ensures everything stays consistent on that thread).

Still, given the `Send + Sync` claim, consider either:
- Gating the body of `Drop` on `is_r_main_thread()` and `eprintln!` + leak
  on the rare off-main case (matches the project's "trust no defensive
  internal invariant" principle in the *other* direction — leak over UB).
- Or document explicitly in the safety comment that callers who move the
  `ProgressBar` to a non-R thread are on their own.

Not blocking; the current comment is honest about the trade-off. Worth a
follow-up issue if not addressed here.

**S3. `term_like_stdout_with_hz` / `term_like_stderr_with_hz` are public API with no tests.**

The new `term_like_connection_with_hz` is also untested. These are
one-liners, but they're public — `clippy_all` won't catch a signature
regression. Either add a smoke test similar to `indicatif_factories_compile`
that constructs each, or drop the `_with_hz` factories if no consumer uses
them (they're not exercised internally). Nice-to-have, not blocking.

### Nice-to-have (4)

**N1. The unprotected return from `eval_base_noarg` is a latent footgun.**

Not changed by this PR, but `eval_base_noarg` (introduced in #175) returns
a freshly-allocated SEXP unprotected. Every caller has to protect
immediately. Worth opening a follow-up to either return a `Protected<SEXP>`
or to protect inside the helper. Not blocking — pre-existing.

**N2. `classify_connection` description-match fallthrough to `TerminalStdout` is subtle.**

`stdin` is documented to route through `Rprintf` "as least-surprise rather
than longjmping," but at runtime no `#[miniextendr]` fn will ever construct
an `RTerm` from `stdin()` and call `write_str` on it (indicatif draw targets
are always write-mode). A `debug_assert!(desc != "stdin")` or a stronger
`panic!` on the stdin-write-target case would catch user error earlier with
a clear message. Optional.

**N3. `write_bytes` does the `is_r_main_thread()` check, then `write_console_to` does it again.**

Harmless duplication. Could be removed in `write_bytes` since the function
short-circuits to `Ok(())` on off-thread anyway, but leaving both is
defensive. Not blocking.

**N4. Tests don't exercise the terminal-connection branch through the new dispatch.**

Existing fixtures (`indicatif_short_bar`, `indicatif_spinner_demo`, etc.)
still go through `term_like_stderr`, so they hit the `TerminalStderr` arm
of `write_bytes`. The textConnection / file / null tests cover the
`Connection` arm. Coverage is complete *across the dispatch matrix*, but no
single new test asserts both branches are reachable in one process. Optional.

### Verified-pre-existing (not regressed by this PR)

- macOS arm64 + `nonapi` cdylib link failure (`_R_nativeEncoding` /
  `_known_to_be_utf8` / `_latin1locale` undefined). The implementer's note
  is correct; this PR doesn't touch that path. Reproduces on `origin/main`
  for anyone testing locally with `nonapi` enabled.
- `miniextendr-bench/from_r` CHARSXP panic — pre-existing.
- `rpkg/tests/testthat/_snaps/` directory does not exist in tree (verified).
  The `indicatif_rterm_debug` snapshot will be a fresh file when CI
  regenerates with `indicatif` enabled. The implementer's claim that the
  test skips locally is consistent with the `skip_if_missing_feature`
  helper in `rpkg/R/feature_helpers.R:29`.
- `rpkg/R/miniextendr-wrappers.R` does not include any indicatif fixture
  wrappers (`grep -c indicatif` returns 0 on `origin/main`). The new
  `indicatif_drive_connection` / `indicatif_to_null_connection` wrappers
  will only materialise when CI rebuilds with the indicatif feature, same
  as every existing indicatif fixture. Matches established pattern.

### Verified-load-bearing claims

- `inherits("terminal")` is the correct discriminator: `newterminal` in
  `r-svn/src/main/connections.c:3178-3203` always sets
  `con->class = "terminal"`, and `do_stdin` / `do_stdout` / `do_stderr`
  attach `c(con->class, "connection")` as the SEXP class. Confirmed.
- Description strings are exactly `"stdin"` / `"stdout"` / `"stderr"`:
  `Connections[0..2] = newterminal(...)` at `connections.c:5768-5773`.
  Confirmed; the `match desc.as_str()` in `classify_connection` will
  correctly route `"stderr"` to `TerminalStderr` and `"stdout"` to the
  default `TerminalStdout` arm.
- `TermLike: Debug + Send + Sync`: confirmed in
  `rpkg/vendor/indicatif-0.18.4/src/term_like.rs:11`. The
  `unsafe impl Send for RTerm` + `unsafe impl Sync for RTerm` are required;
  the runtime `is_r_main_thread()` guard is the correct enforcement.
- `nonapi` requirement retained for the terminal-branch console hook —
  `write_console_to` uses `Rprintf` / `REprintf` which are `nonapi`.
  Confirmed; the issue's finding #4 stands.
- `RNullConnection: IntoR` correctly disarms `Drop` and transfers ownership
  (`into_r.rs:2576-2592`); `RTerm::new` then re-preserves the same SEXP.
  Net `R_PreserveObject` count is exactly one for the lifetime of
  `RTerm`. Confirmed.
- `pub(crate) write_console_to` visibility is the tightest available since
  `connection` and `progress` are siblings at the crate root; `pub(super)`
  would behave identically here. The visibility bump is correctly minimal.
  Confirmed.
- CI clippy_all feature list correctly extended (`+,indicatif,` on
  `.github/workflows/ci.yml:241`). Confirmed.
- `expect_match(rendered, "3/3", fixed = TRUE)` is load-bearing: the test
  would fail if bytes never reached the textConnection, because indicatif's
  `{pos}/{len}` template renders the literal `"3/3"` only after `pb.inc(1)`
  has been called all three times *and* the bytes have flowed through
  `RTerm::write_bytes` → `write_connection` → `R_WriteConnection`.

### Summary

Findings: 1 must-fix (M1 — PROTECT window), 3 should-fix / discuss (S1–S3),
4 nice-to-have (N1–N4).

The PR is small (396 / 147), surgical, and design-correct. M1 is the only
real correctness concern and is a one-line fix.

For the fixer: leave S1's `<TODO>` placeholder alone — the user fills that
in. Address M1 (reorder `R_PreserveObject` before `classify_connection`).
S2 is a judgment call; S3 / N1–N4 are optional.

PR #658, opened 2026-05-19
fix(build): skip link_to_r() on wasm32 targets (#482)

### LINKED ISSUE #482: miniextendr-api/build.rs: skip link_to_r() when target_arch = wasm32
## Goal

Make `cargo check --target wasm32-unknown-emscripten -p miniextendr-api` (and `-p miniextendr-engine`) work on any machine **without R installed and without any `R_HOME` env var workaround**, by gating the libR-resolution / link-emit step on `CARGO_CFG_TARGET_ARCH != "wasm32"`.

This unblocks tier-1 webR CI from including `rpkg` (once the `rpkg/configure` Rscript dependency is separately resolved, #470), and lets the `R_HOME=$RUNNER_TEMP` workaround be deleted from both `webr.yml` and `ci.yml::rust-lint`.

## Background

`miniextendr-api/build.rs:16` and `miniextendr-engine/build.rs:10` both call `link_to_r()` unconditionally. `link_to_r()` runs `R RHOME` (or reads `$R_HOME`), then panics if the resolved directory doesn't exist. On wasm32 there is no libR to link against, the toolchain doesn't even ship a host R, and `cargo check` never links — so the entire branch is pure overhead/breakage on wasm targets.

Two CI workarounds currently exist solely to dodge this:

- `.github/workflows/webr.yml:54-55` — `Set dummy R_HOME` step (`echo "R_HOME=$RUNNER_TEMP" >> $GITHUB_ENV`). Cites `build.rs:64` directly in the comment.
- `.github/workflows/ci.yml:198-204` — same `Set dummy R_HOME` step in the `rust-lint` job, for the same reason: fmt + clippy never link but `build.rs` still panics on missing R.

Both reference the same root cause. Once the guard lands, both workarounds are deletable (note: rust-lint runs on native x86_64 Linux, so the wasm32 guard alone does NOT cover it — see "Files to touch" below for the dual-axis approach).

The webR CI workflow's tier-1 scope is also blocked from extending to `rpkg` cargo check (see `webr.yml:71-75` comment). This issue is the prerequisite that comment refers to.

Related context that has already shipped (do not redo):

- PR #474 (merged 2026-05-10) — cfg-branched `[distributed_slice]` slices in `miniextendr-api/src/registry.rs` so the crate **compiles** on wasm32 (40 errors → 0). That's the in-crate code change. The build.rs guard is the missing piece outside the source tree.
- PR #480 (merged 2026-05-10) — adds `.github/workflows/webr.yml` with the dummy-R_HOME workaround.
- PR #481 (merged 2026-05-10) — `rpkg/src/rust/build.rs` (separate file, user-crate template) added wasm32 detection + `wasm_registry.rs` snapshot validation. Did **not** touch `miniextendr-api/build.rs`. That's this issue's scope.

## Approach

A targeted three-line guard at the top of `link_to_r()` (or wrapping each `link_to_r()` call site). One-liner using cargo's standard cfg env var:

```rust
if std::env::var("CARGO_CFG_TARGET_ARCH").as_deref() == Ok("wasm32") {
    return;
}
```

### Exact-arch (`target_arch == "wasm32"`) vs broader check

The proposal in the issue is exact-arch. That is correct and sufficient.

- webR is `wasm32-unknown-emscripten`. `CARGO_CFG_TARGET_ARCH` is `wasm32`.
- The hypothetical alternatives (`target_os == "unknown"`, `target_env == ""`) would either over-match (Linux statically-linked `unknown` envs) or under-match (emscripten sets `target_os = emscripten`, not `unknown`).
- All currently-relevant wasm targets — `wasm32-unknown-emscripten`, `wasm32-unknown-unknown`, `wasm32-wasi`, `wasm32-wasip1`, `wasm32-wasip2` — share `target_arch = wasm32`. Future `wasm64-*` is unlikely on the webR roadmap; if it ever lands, a one-line addition handles it.

**Decision: exact `CARGO_CFG_TARGET_ARCH != "wasm32"` only.** Don't broaden it.

### Placement

Two options, prefer the second:

1. Guard each `link_to_r()` call site individually in `main()`. Pro: explicit at call site. Con: duplicated guard in two crates, and the function still has the "panics if R isn't found" footgun in its body for any future caller.
2. **Guard at the top of `link_to_r()` itself** — early return on wasm32, so the rest of the function (with all its panics) is unreachable. Pro: function becomes self-protective. Con: function name slightly misleading on wasm (it intentionally does nothing).

Pick option 2. Add a comment explaining why.

The `set_stack_size_flags()` call in `miniextendr-api/build.rs:12` can stay unconditional — its match arms already gracefully fall through to `_ => {}` on unknown `target_os`, including emscripten. No-op on wasm32, no R dependency.

### Code (the actual fix)

- `miniextendr-api/build.rs:16` — call site of `link_to_r()`. The guard goes inside `link_to_r()` at line 57 (top of function body).
- `miniextendr-engine/build.rs:10` — same crate-level pattern. Same one-liner guard inside its own copy of `link_to_r()` at line 15.

Both `link_to_r()` implementations are near-identical copies of each other. Apply the same guard to both.

### CI cleanup (in-scope — bundle into same PR)

- `.github/workflows/webr.yml:51-55` — delete the `Set dummy R_HOME` step entirely (5 lines including comment + name + run + blank line). Also delete the workaround-citing comment at `webr.yml:71-75` once verified the tier-1 scope can be widened in a follow-up (or leave that comment for #470 to handle — keep this PR minimal and just remove the now-unused workaround).

  **Decision: only remove the `Set dummy R_HOME` step in this PR. Don't touch the tier-1 scope comment or the paths filter — that's #470 territory.**

- `.github/workflows/ci.yml:198-204` — the `rust-lint` job's `Set dummy R_HOME` step. This is a **native x86_64 Linux** runner, not wasm32, so the wasm32 guard alone does NOT fix this case.

  **Decision: this workaround stays for now.** rust-lint compiles `miniextendr-api` for native targets (clippy/fmt), where `link_to_r()` legitimately needs R. The dummy `R_HOME=$RUNNER_TEMP` trick works because `Path::new(&r_home).is_dir()` returns true for `$RUNNER_TEMP` and clippy never actually links. Removing it would require a separate refactor (e.g., gate `link_to_r()` on `CARGO_CFG_TARGET_FAMILY` + a build-time signal that this is a check-only invocation) which is out of scope for this issue.

  Cite this explicitly in the PR description so reviewers don't expect the rust-lint workaround to disappear.

### No template / scaffolder changes

`rpkg/src/rust/build.rs` is the user-crate template (separate file, updated by #481) and is **not** affected by this issue. It has its own wasm32 detection logic.

## Follow-up cleanup

- After this PR lands, #470's tier-1 scope can be widened to include `rpkg` cargo-check **only after** `rpkg/configure`'s Rscript dependency is also addressed (a parallel fix — either `setup-r@v2` in the workflow, or skip-configure on wasm32). That's `webr.yml:71-75`'s comment. Not in scope here; reference #470 in the PR body.
- The `rust-lint` dummy-R_HOME workaround in `ci.yml` is a separate concern. If someone wants to eliminate it, file a follow-up issue proposing a `MINIEXTENDR_SKIP_LINK_TO_R` env var or detecting `CARGO_CFG_*` clippy/check-only contexts. **Don't bundle into this PR.**

## Flat work items

1. Edit `miniextendr-api/build.rs::link_to_r()` — add early-return guard at top of function body. Use `CARGO_CFG_TARGET_ARCH` env var (this is build-script context; `cfg!(target_arch = "wasm32")` would refer to the **host** target of the build script binary, not the requested cross-compile target — that's why we must use the env var).
2. Edit `miniextendr-engine/build.rs::link_to_r()` — apply the same guard. Same reasoning.
3. Delete the `Set dummy R_HOME` step from `.github/workflows/webr.yml` (5 lines).
4. Verify locally: `cargo check --target wasm32-unknown-emscripten -p miniextendr-api -p miniextendr-engine` succeeds with R uninstalled and `R_HOME` unset.
5. Verify native unaffected: `cargo check --workspace --all-targets --locked` still passes on a machine **with** R installed.
6. PR body cites:
   - This issue (#482).
   - Notes that `ci.yml::rust-lint`'s dummy-R_HOME workaround intentionally stays (different root cause: native build, no link).
   - Notes that webR tier-1 widening to `rpkg` is still blocked on `rpkg/configure`'s Rscript dependency (#470 follow-up).

## Non-goals

- Do **not** wire up `rpkg/configure` for wasm32 in this PR. That's a `configure.ac` / Makevars problem, owned by #470. The `rpkg/src/rust/build.rs` wasm32 detection PR (#481) already addressed the Rust-side template; the configure-side is a separate axis.
- Do **not** remove the `rust-lint` dummy-R_HOME workaround in `ci.yml`. The wasm32 guard doesn't fix it (native target).
- Do **not** broaden the guard to `target_os == "unknown"` or `target_family == "wasm"`. Exact `target_arch == "wasm32"` is the right granularity (see "Approach > Exact-arch" above).
- Do **not** refactor `link_to_r()` into a shared utility used by both `miniextendr-api/build.rs` and `miniextendr-engine/build.rs`. Build scripts can't share code without a build-dependencies crate, and the duplication is small enough that adding one is over-engineering for ~5 LOC.
- Do **not** unify the two `link_to_r()` copies. They're identical now; if they ever diverge, that's a separate cleanup.

## Validation

- **wasm32 cross-check without R installed**: on a machine where `R` is not on PATH and `R_HOME` is unset, run:
  ```
  rustup target add wasm32-unknown-emscripten   # if not already
  cargo check --target wasm32-unknown-emscripten -p miniextendr-api
  cargo check --target wasm32-unknown-emscripten -p miniextendr-engine
  ```
  Both should succeed. Before this PR they panic at `build.rs:64` / `build.rs:22` with `Failed to run R RHOME`.
- **Native unaffected**: `cargo check --workspace --all-targets --locked` on a machine with R installed — should still resolve `R_HOME` and emit the link flags exactly as before.
- **CI**: the existing `webr.yml` job should pass **without** the `Set dummy R_HOME` step. This is the load-bearing CI signal; if it fails after deleting the step, the guard isn't catching every code path.
- **CI native lint unchanged**: `ci.yml::rust-lint` should continue passing — it still has the dummy-R_HOME workaround, which still works for the same reason as before (native build, no link, dummy dir satisfies `is_dir()` check).

## Expected diff shape

- `miniextendr-api/build.rs`: +4 / -0 (early-return guard + comment)
- `miniextendr-engine/build.rs`: +4 / -0 (same)
- `.github/workflows/webr.yml`: -5 / +0 (delete the `Set dummy R_HOME` step + its comment)

Total: roughly +8 / -5, in 3 files. Single squashed commit. No template/scaffolder churn, no NAMESPACE/wrapper regeneration, no `just vendor`.

CGMossa:
Routes around the unconditional `R RHOME` call in both `miniextendr-api/build.rs` and `miniextendr-engine/build.rs` so `cargo check --target wasm32-unknown-emscripten` works on machines without R installed.

<details>

## Summary
- `miniextendr-api/build.rs` and `miniextendr-engine/build.rs`: early `return` from `link_to_r()` when `CARGO_CFG_TARGET_ARCH == "wasm32"`. Must use the env var, not `cfg!(target_arch = "wasm32")` — the cfg resolves to the build-script binary's host arch, not the requested cross-compile target.
- `.github/workflows/webr.yml`: drop the now-redundant `Set dummy R_HOME` step (5 lines + comments) — its only purpose was dodging the panic this guard eliminates.
- `.github/workflows/ci.yml::rust-lint`'s analogous `R_HOME=$RUNNER_TEMP` workaround is intentionally left alone. That job runs on native x86_64 Linux where `link_to_r()` legitimately needs R; the wasm32 guard doesn't help it.

Closes #482.

## Test plan
- [x] `cargo check --workspace --all-targets --locked` on native target
- [ ] CI verification: webR workflow no longer needs the dummy `R_HOME` and `cargo check --target wasm32-unknown-emscripten -p miniextendr-api` succeeds

## Notes
- Both `build.rs` files carried the same broken pattern. PR #481 (wasm32 detection in `configure.ac`) had touched a different file (`rpkg/src/rust/build.rs` template), so the framework-side scripts went uncovered.
- Stack-size flag setup in `set_stack_size_flags()` already falls through to `_ => {}` on emscripten/unknown OS, so no second guard is needed.

---

</details>

### DIFF
--- a/.github/workflows/webr.yml
+++ b/.github/workflows/webr.yml
@@ -48,12 +48,6 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      # build.rs:64 unconditionally runs `R RHOME`; $RUNNER_TEMP is always a
-      # real directory so Path::new(&r_home).is_dir() passes. cargo check never
-      # links, so the dummy value is sufficient. Same pattern as rust-lint in ci.yml.
-      - name: Set dummy R_HOME
-        run: echo "R_HOME=$RUNNER_TEMP" >> "$GITHUB_ENV"
-
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
--- a/miniextendr-api/build.rs
+++ b/miniextendr-api/build.rs
@@ -54,6 +54,16 @@ fn set_stack_size_flags() {
 }
 
 fn link_to_r() {
+    // Skip libR resolution on wasm32 targets (webR / wasm32-unknown-emscripten
+    // and friends). cargo check never links and the toolchain doesn't ship a
+    // host R, so the rest of this function is pure overhead/breakage there.
+    // Must use the env var, not `cfg!(target_arch = "wasm32")`: the cfg refers
+    // to the build script binary's host arch, not the requested cross-compile
+    // target. See issue #482.
+    if env::var("CARGO_CFG_TARGET_ARCH").as_deref() == Ok("wasm32") {
+        return;
+    }
+
     // Resolve R home directory.
     let r_home = if let Ok(val) = env::var("R_HOME") {
         val
--- a/miniextendr-engine/build.rs
+++ b/miniextendr-engine/build.rs
@@ -12,6 +12,16 @@ fn main() {
 
 /// Resolves `R_HOME` and emits linker flags for libR.
 fn link_to_r() {
+    // Skip libR resolution on wasm32 targets (webR / wasm32-unknown-emscripten
+    // and friends). cargo check never links and the toolchain doesn't ship a
+    // host R, so the rest of this function is pure overhead/breakage there.
+    // Must use the env var, not `cfg!(target_arch = "wasm32")`: the cfg refers
+    // to the build script binary's host arch, not the requested cross-compile
+    // target. See issue #482.
+    if env::var("CARGO_CFG_TARGET_ARCH").as_deref() == Ok("wasm32") {
+        return;
+    }
+
     // Resolve R home directory.
     let r_home = if let Ok(val) = env::var("R_HOME") {
         val


CGMossa (commented):
Reviewed against the #482 plan. Implementation matches the prescribed approach exactly.

<details>

## Verdict

LGTM — no blockers. Ship it.

## What I checked

- **`CARGO_CFG_TARGET_ARCH == "wasm32"` env-var check (not `cfg!`)** — correct. The 6-line comment in both build.rs files explains *why* (build script's cfg reflects host arch, not the cross-compile target), which is the single most useful piece of context for the next reader to encounter this code. The issue's "Expected diff shape" budgeted a one-line comment; the PR spent five extra lines explaining the cfg-vs-env-var subtlety in-place, and it's well worth them.
- **Both build.rs files guarded** — confirmed:
  - `miniextendr-api/build.rs:57-65` — guard at top of `link_to_r()` body, before any `env::var("R_HOME")` / `Command::new("R")` call.
  - `miniextendr-engine/build.rs:15-23` — same guard, same wording. Identical copies — appropriate per the issue's non-goal to not unify them.
- **`set_stack_size_flags()` still runs on wasm32** — confirmed. The guard is inside `link_to_r()`, not before it. `main()` calls `set_stack_size_flags()` first (line 12), then `link_to_r()` (line 16) — only the latter early-returns. `set_stack_size_flags()`'s match arms fall through to `_ => {}` for emscripten (no `wasm` / `emscripten` arm), so it's a no-op on wasm32 anyway. Net behavior on wasm32: stack-flag function returns harmlessly, libR resolution skipped. Correct.
- **`webr.yml` step removal** — 6 lines removed (the named step + its comment + the trailing blank). Nothing else in `webr.yml` references `R_HOME` or depends on the deleted step. The follow-on tier-1-scope comment at the bottom of the file (lines 71-75 of the original) is correctly retained — it references #482 + #470 and stays relevant until `rpkg/configure`'s `Rscript` dependency is separately addressed.
- **`ci.yml::rust-lint` intentionally untouched** — confirmed via `git diff origin/main..pr-658-readonly -- .github/workflows/ci.yml` (no diff). The `Set dummy R_HOME` step there is still present at lines 203-204. PR description calls this out explicitly with the correct rationale (native x86_64 Linux, the wasm32 guard doesn't help it).
- **No tests** — correct. Pure build-script change; the load-bearing CI signal is the `wasm32 check` job in `webr.yml` (currently passing at 16s on this PR), which exercises the guard on a runner without R installed.

## Plan adherence

Every "Flat work item" from #482 landed:

1. `miniextendr-api/build.rs::link_to_r()` early-return guard — done.
2. `miniextendr-engine/build.rs::link_to_r()` same guard — done.
3. Delete `Set dummy R_HOME` step from `webr.yml` — done.
4-5. Validation matrix — wasm32 CI green; native CI pending but the only line changes are gated behind the wasm32 branch.
6. PR body cites #482, notes `rust-lint` retention, and notes #470 follow-up — all there.

Every non-goal respected: no `rpkg/configure` wasm32 wiring, no broader `target_family = "wasm"` match, no `link_to_r()` extraction into a shared crate, no `rust-lint` removal.

## Nits / not blockers

- The comment block exceeds the issue's "+4 / -0" budget. This is an improvement, not a defect — flagging only because future agents diffing against the plan might wonder.
- The two build.rs files now have a near-perfect duplicate of the same 6-line comment block. The issue explicitly says not to unify, so this is correct; if they ever diverge that's a separate cleanup signal.

</details>

Issue #597, opened 2026-05-17
Mirror R's LTO flag into cargo [env] RUSTFLAGS


CGMossa:
Follow-up to #601 (PR1 of the cargo `[env]` sprint). PR1 introduced the
`emit_env_block` shell function in `rpkg/configure.ac` (and both templates)
that writes a `[env]` table into `src/rust/.cargo/config.toml`. This issue
extends that block so a `--enable-lto` R build pulls the matching cargo
profile-LTO setting through to every cargo invocation.

## Goal

When R was built with `--enable-lto` (i.e. `R CMD config LTO` returns a
non-empty `-flto*` value), `cargo build --release` for the staticlib and
the wrapper-gen cdylib both compile with `lto = "thin"` (mapped from R's
`-flto` flag space). When R's LTO is off (the default), nothing changes.

## Background

- `R CMD config LTO` is the canonical probe. The semantics come from
  R's own `configure.ac` (`.r-svn/configure.ac:364-410`): when
  `--enable-lto` is passed, `LTO`/`LTOALL` substitute into
  `etc/Makeconf` as `-flto` (or whatever the user supplied via
  `LTO=...` at configure time, e.g. `-flto=thin`, `-flto=auto`, `-flto=N`).
  When `--enable-lto` is NOT passed (the default), `LTO` is empty.
- `R CMD config LTO` is reliable: even on a non-LTO R, the call succeeds
  and prints an empty line — no fallback parsing needed.
- PR #601 landed `emit_env_block()` as a shell function inside
  `AC_CONFIG_COMMANDS([cargo-config], …)`, invoked from all three
  install-mode branches (tarball / source-monorepo / source-only).
  TOML table-scoping requires `[env]` to be the first table in
  every branch — this hasn't changed.
- The `[env]` block uses cargo's default no-`force` semantics: shell
  env wins over `[env]` entries.

## Design decision (load-bearing): recommend Option C

**Option A** as sketched in the original body — emit
`CARGO_ENCODED_RUSTFLAGS = "-C lto=thin"` (or plain `RUSTFLAGS`) into
`[env]` — does **NOT work** for this codebase. Here's the gotcha:

`rpkg/src/Makevars.in` already exports `RUSTFLAGS` on the cargo
invocation line:

```make
RUSTFLAGS="$(ENV_RUSTFLAGS) $$LINK_ARGS" \
$(CARGO) … build …
```

Cargo's documented precedence for rustflags is "first non-empty source
[...]

PR #659, opened 2026-05-19
feat(configure): propagate ABI-only CFLAGS/CXXFLAGS into cargo [env]

### LINKED ISSUE #598: Propagate CFLAGS/CXXFLAGS from R config into cargo [env] (per-target)
Follow-up to PR #601 (cargo `[env]` block introduction).

PR1 emitted `MACOSX_DEPLOYMENT_TARGET`, `CARGO_TARGET_*_LINKER`, `CARGO_NET_GIT_FETCH_WITH_CLI`, and conditional `AR_<target>` / `RANLIB_<target>` into `src/rust/.cargo/config.toml`'s `[env]`. CFLAGS/CXXFLAGS propagation was deliberately deferred until we had a defensible filter design — emitting R's raw `CFLAGS` would punch R-specific `-I` paths and `-D_*` defines into every transitive C build script in the dependency graph.

## Goal

Make Rust crates' transitive C/C++ build scripts (via `cc`, `cmake`, `bindgen`, etc.) inherit only the **ABI-affecting** subset of R's compilation flags. Crates need to match R's struct layout / instruction-set / threading assumptions, but they must **not** inherit R's include search paths, R-specific preprocessor defines, warning flags, or `-O*` / `-g*` (cargo profiles own those).

## Background

- PR #601 introduced `emit_env_block()` at `rpkg/configure.ac:585-608`. It's a shell function called from inside `AC_CONFIG_COMMANDS([cargo-config], …)` and emits `[env]` as the **first table** in each install-mode branch of `src/rust/.cargo/config.toml`. TOML table-scoping requires it to precede `[source.*]` / `[patch.*]` / `[build]`.
- Derivation variables are computed at autoconf time (lines 385-499) and passed individually through the second arg of `AC_CONFIG_COMMANDS` (lines 668-679) — this avoids embedded-quote quoting hell.
- Cargo recognises per-target `CFLAGS_<target>` / `CXXFLAGS_<target>` (uppercase, hyphens-to-underscores), e.g. `CFLAGS_x86_64_unknown_linux_gnu`. They feed `cc::Build::new()` (the standard build-script C compiler crate) and most build scripts that wrap it. Source: <https://doc.rust-lang.org/cargo/reference/environment-variables.html> and <https://docs.rs/cc/latest/cc/#external-configuration-via-environment-variables>.
- The target-triple is already resolved at line 387-393 (`CARGO_ENV_RUST_TARGET` and `CARGO_ENV_RUST_TARGET_UPPER`) — reuse those.
- `[env]` is emitted **without `force = true`**, so shell env wins. Users with site-specific needs can still override at install time.

## Filter design (load-bearing)

R's `CFLAGS` on a CRAN-built Linux R is typically along the lines of:

```
-I"/usr/share/R/include" -DNDEBUG -fpic -g -O2 -fdebug-prefix-map=… -fstack-protector-strong \
-Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -Wall -mtune=generic
```

Homebrew/macOS R adds `-I/opt/homebrew/include` (the dev host's actual output). Conda-Forge R may add `-march=nocona -mtune=haswell`. Rocker images may add `-mavx2`.

Of these, only a small subset legitimately affects **ABI** between the Rust staticlib and R's `.so`. The rest is either redundant with cargo defaults, actively harmful (R-specific `-I`), or stylistic (`-W*`, `-g*`, `-O*`).

We adopt a **strict allowlist**. Anything not on the allowlist is dropped. Rationale per family:

| Flag family | Decision | Rationale |
|---|---|---|
| `-march=*` | **KEEP** | Instruction-set baseline; ABI-affecting for SIMD intrinsics and vector struct layout. Mismatch → SIGILL or silent miscompiles. |
| `-mtune=*` | **KEEP** | Not strictly ABI but commonly paired with `-march`; harmless if R picked a value. |
| `-mfpmath=*`, `-mfpu=*` | **KEEP** | Float-ABI on 32-bit ARM and some legacy x86 (FPU-vs-SSE) — load-bearing where present. |
| `-msse*`, `-mavx*`, `-mssse*`, `-msse4*`, `-mpclmul`, `-maes` | **KEEP** | Same as `-march`. Matches what conda-forge / vendor channels actually set. |
| `-mabi=*` | **KEEP** | RISC-V / ARM hard-float vs soft-float ABI. Critical where present. |
| `-mfloat-abi=*` | **KEEP** | ARM hard/soft float — explicit ABI. |
| `-pthread` | **KEEP** | Thread-local storage layout / pthread struct sizes on glibc. |
| `-D_REENTRANT` | **KEEP** | Same family as `-pthread`; some platforms use it instead. |
| `-fopenmp`, `-fopenmp=*` | **KEEP** | OpenMP runtime presence affects struct layout and linkage; if R has it, transitive C code likely needs it too. |
| `-I*` | **DROP** | R-specific include paths. Cargo deps must not see `R.h` / `Rinternals.h` by accident. |
| `-isystem*` | **DROP** | Same. |
| `-D_*` | **DROP** | R-private feature-test macros (`_FORTIFY_SOURCE`, `_FILE_OFFSET_BITS`, `_REENTRANT` already handled above as the one exception). `_FORTIFY_SOURCE` changes libc semantics; we want cargo deps' own choices. |
| `-DNDEBUG` | **DROP** | cargo's release profile handles this via `debug-assertions = false`; redundant and conflicts with cargo's debug profile. |
| `-W*` | **DROP** | Warning flags are not ABI. `R CMD check --as-cran` already objects to `-W*` in `PKG_CFLAGS` (CLAUDE.md "Common issues"); inheriting them into cargo's transitive C builds would re-introduce that hazard via the back door. |
| `-O*` | **DROP** | Optimisation level is a cargo profile concern (`debug` vs `release` vs `--profile=…`). Forcing R's `-O2` into a `cargo build --release` would be a no-op; forcing it into `cargo build` (debug) would surprise. |
| `-g*` | **DROP** | Debug info is a cargo profile concern. |
| `-fPIC`, `-fpic` | **DROP** | Cargo already passes the right PIC flag for staticlib/cdylib targets — redundant. |
| `-fstack-protector*` | **DROP** | Hardening, not ABI. Distros pick this; cargo deps' build scripts should make their own choice. |
| `-fdebug-prefix-map=*`, `-ffile-prefix-map=*` | **DROP** | Build-reproducibility flags; pointing into R's build tree is meaningless for cargo deps. |
| `-fno-plt`, `-fno-stack-clash-protection`, `-fexceptions` | **DROP** | Codegen knobs, not ABI for staticlib/cdylib boundary. |
| `-arch *` (macOS) | **DROP** | Already covered by the `*_<target>` env var name itself; emitting `-arch arm64` for an `aarch64-apple-darwin` target is redundant and breaks if cargo cross-compiles. |
| Everything else | **DROP** | Conservative default. Anything not explicitly classified ABI-relevant gets stripped; we can revisit individual flags as users surface them. |

The allowlist regex (for sed/grep): `^-march=|^-mtune=|^-mfpmath=|^-mfpu=|^-mfloat-abi=|^-mabi=|^-msse|^-mavx|^-mssse|^-mpclmul|^-maes|^-pthread$|^-D_REENTRANT$|^-fopenmp`.

`-D_REENTRANT` and `-pthread` are anchored with `$` so we don't accidentally match `-pthread-foo` or `-D_REENTRANTLY` (paranoid, but cheap).

Note: `-march`/`-mtune`/etc. on macOS clang use the same spelling as gcc, so no platform branching needed. The filter is uniform across all hosts.

## Approach

Inside `rpkg/configure.ac`, immediately after the `AR`/`RANLIB` derivation (current line 493), add:

```sh
dnl --- CFLAGS / CXXFLAGS: ABI-only allowlist from R config ----
dnl Cargo applies CFLAGS_<target> and CXXFLAGS_<target> to *every* transitive
dnl C/C++ build script (cc, cmake, etc.). R's raw CFLAGS contain R-specific
dnl -I paths, R-private defines, and warning flags that have no business in
dnl cargo deps. We allowlist a small set of ABI-affecting flags; everything
dnl else is stripped. See issue #598 for the full rationale table.
CARGO_ENV_ABI_CFLAGS=""
CARGO_ENV_ABI_CXXFLAGS=""

_r_cflags=`"${R_HOME}/bin/R" CMD config CFLAGS 2>/dev/null`
_r_cxxflags=`"${R_HOME}/bin/R" CMD config CXXFLAGS 2>/dev/null`

dnl Allowlist regex (POSIX BRE-friendly; one alternation per line via tr → grep -E).
dnl Keep ABI-affecting flags only. See issue #598 for rationale.
_abi_allow_re='^-march=|^-mtune=|^-mfpmath=|^-mfpu=|^-mfloat-abi=|^-mabi=|^-msse|^-mavx|^-mssse|^-mpclmul|^-maes|^-pthread$|^-D_REENTRANT$|^-fopenmp'

if test -n "$_r_cflags"; then
  CARGO_ENV_ABI_CFLAGS=`echo "$_r_cflags" | tr ' ' '\n' | grep -E "$_abi_allow_re" | tr '\n' ' ' | $SED 's/ *$//'`
fi
if test -n "$_r_cxxflags"; then
  CARGO_ENV_ABI_CXXFLAGS=`echo "$_r_cxxflags" | tr ' ' '\n' | grep -E "$_abi_allow_re" | tr '\n' ' ' | $SED 's/ *$//'`
fi

if test -n "$CARGO_ENV_ABI_CFLAGS"; then
  AC_MSG_NOTICE([cargo CFLAGS_<target>  = $CARGO_ENV_ABI_CFLAGS])
fi
if test -n "$CARGO_ENV_ABI_CXXFLAGS"; then
  AC_MSG_NOTICE([cargo CXXFLAGS_<target>= $CARGO_ENV_ABI_CXXFLAGS])
fi
```

m4 quoting reminders:
- Brackets in regexes inside `configure.ac` must use `@<:@` for `[` and `@:>@` for `]`. The allowlist regex above happens to use only `^`, `$`, `|`, `*`, `=`, ASCII — no brackets — so no escaping is needed. **Do not introduce character classes (`[A-Z]`)** without escaping.
- The `tr ' ' '\n' | grep -E … | tr '\n' ' '` pipe is preferred over a single `sed` because (a) sed alternation portability is patchy across BSD/GNU, (b) `grep -E` is universally available and matches one whole token at a time, (c) trailing-space cleanup via `$SED 's/ *$//'` keeps the emitted TOML value tidy.

Then inside `emit_env_block()` (current lines 585-608), after the existing `AR_<target>` / `RANLIB_<target>` block, add:

```sh
if test -n "$CARGO_ENV_ABI_CFLAGS"; then
  echo "CFLAGS_$CARGO_ENV_RUST_TARGET = \"$CARGO_ENV_ABI_CFLAGS\""
fi
if test -n "$CARGO_ENV_ABI_CXXFLAGS"; then
  echo "CXXFLAGS_$CARGO_ENV_RUST_TARGET = \"$CARGO_ENV_ABI_CXXFLAGS\""
fi
```

**Important on the key name**: Cargo's spelling is `CFLAGS_<TARGET>` where `<TARGET>` is the target triple with hyphens converted to underscores and **uppercased**. Looking at the existing code, the `AR_<target>` line at 602 uses `$CARGO_ENV_RUST_TARGET` (lowercase, hyphens) — but cargo accepts BOTH spellings for these env vars (the docs note them as case-insensitive on lookup but the canonical form is upper-with-underscores). To stay consistent with the existing `AR_<target>` emission and avoid churn, use the same `$CARGO_ENV_RUST_TARGET` form (e.g. `CFLAGS_aarch64-apple-darwin = "..."`). **If the implementer finds cargo rejects the hyphenated form in practice, switch ALL four (`AR_*`, `RANLIB_*`, `CFLAGS_*`, `CXXFLAGS_*`) to `$CARGO_ENV_RUST_TARGET_UPPER` together** — do not split styles between keys. (Reference: cc-rs and cargo both accept either; cargo's env-var lookup normalises internally.)

Update the variable-pass list at the bottom of `AC_CONFIG_COMMANDS([cargo-config], …)` (lines 668-679) to include the two new variables:

```
CARGO_ENV_ABI_CFLAGS="$CARGO_ENV_ABI_CFLAGS"
CARGO_ENV_ABI_CXXFLAGS="$CARGO_ENV_ABI_CXXFLAGS"
```

## Files to touch

- `rpkg/configure.ac`
  - Add derivation block after line 493 (after `CARGO_ENV_R_RANLIB`).
  - Extend `emit_env_block()` at lines 585-608 with two new conditional `echo` lines.
  - Extend variable-pass list at lines 668-679 with two new entries.
- `minirextendr/inst/templates/rpkg/configure.ac` — mirror (insertion points: after line 523, inside `emit_env_block()` around line 632, variable-pass list at line 708).
- `minirextendr/inst/templates/monorepo/rpkg/configure.ac` — mirror (same insertion points).
- `patches/templates.patch` — regenerate via `just templates-approve` after the two template edits.
- `rpkg/configure` — regenerate via `autoconf` in `rpkg/`. Do **not** hand-edit.

No changes needed in `Makevars.in`, `cargo-config.toml.in`, or any R/Rust source.

## Flat work items (priority order)

1. Add CFLAGS/CXXFLAGS derivation block to `rpkg/configure.ac` (line ~494, after RANLIB).
2. Extend `emit_env_block()` in `rpkg/configure.ac` with the two conditional emit lines.
3. Extend the variable-pass list in `AC_CONFIG_COMMANDS([cargo-config])` with the two new vars.
4. Mirror 1–3 into `minirextendr/inst/templates/rpkg/configure.ac`.
5. Mirror 1–3 into `minirextendr/inst/templates/monorepo/rpkg/configure.ac`.
6. Regenerate `rpkg/configure` via `autoconf` (run from `rpkg/`).
7. Run `just templates-approve` to regenerate `patches/templates.patch`.
8. Add a `build.rs` fixture in `rpkg/src/rust/` (or a tiny side-crate in `tests/`) that uses `cc::Build::new()` to compile a single C file, with a `#[test]` (or build-script `println!("cargo:warning=…")`) that verifies the resulting flags include the expected `-march` value when R's CFLAGS set one. Without a fixture, regression is invisible.
9. `bash ./configure` in `rpkg/` from clean — verify the `CFLAGS_<target>` / `CXXFLAGS_<target>` keys appear (or are absent when R's CFLAGS strip down to empty).
10. `just templates-check` — must report clean.
11. `just configure && just rcmdinstall` — must succeed end-to-end.
12. Push branch, open PR referencing this issue, watch CI matrix (Linux/macOS/Windows × R versions).

## Non-goals

- **LTO propagation** is tracked separately in #597; do not bundle.
- **`LDFLAGS_<target>` propagation** — separate consideration. R's `LDFLAGS` contains paths to R's own libs and `-L` directories that cargo deps must not link against. Defer to a follow-up issue if needed.
- **Forcing values over shell env** (`force = true`) — out of scope; mirrors PR #601's no-force-floor stance.
- **Per-platform allowlist tuning** (e.g. Windows-specific `-mwindows`) — none of the current allowlist entries are platform-conditional; revisit if Windows R surfaces a real ABI flag not in the list.
- **`R_CFLAGS_PROPAGATE_EXTRA` user-facing escape hatch** — premature. If users need a flag we filtered, they can set `CFLAGS_<target>=…` in their shell env (which wins over `[env]`).

## Validation

- **Unit-level**: from clean `rpkg/` tree, `bash ./configure` and inspect `src/rust/.cargo/config.toml`:
  - On macOS (homebrew R) where `R CMD config CFLAGS` returns only `-I/opt/homebrew/include`: the filter drops everything, no `CFLAGS_*` key should appear (the `test -n` guard kicks in).
  - On CRAN-built Linux R: `-march=`/`-mtune=` should make it through if R was built with them; `-DNDEBUG`, `-D_FORTIFY_SOURCE=2`, `-Wall`, `-O2`, `-g`, `-I/usr/share/R/include` all stripped.
  - Synthetic test: temporarily wrap `R CMD config CFLAGS` to return `"-I/r -DNDEBUG -march=haswell -mtune=generic -Wall -O2 -g -fopenmp -pthread"` (e.g. via `R_HOME` shim) and verify the emitted line is exactly `CFLAGS_<target> = "-march=haswell -mtune=generic -fopenmp -pthread"`.
- **Integration (the fixture from work-item 8)**: a `cc::Build`-driven C compile under a transitive build script picks up the `-march` value. Compare `cargo build -vv` output for the C compile invocation with and without `CFLAGS_<target>` set in `[env]`.
- **Template sync**: `just templates-check` must report clean after step 7.
- **CI matrix**: must remain green on Linux/macOS/Windows × R-release / R-devel / R-oldrel.
- **No regression in `R CMD check --as-cran`**: the `-W*` filter is specifically motivated by avoiding the "non-portable compilation flags" WARNING (CLAUDE.md "Common issues" — same hazard, different code path). Confirm `just devtools-check` stays clean.

## Expected diff shape

- `rpkg/configure.ac`: +~25 lines (derivation block) +~6 lines (emit block) +~2 lines (variable pass).
- Each of the two template `configure.ac` files: same +~33 lines.
- `patches/templates.patch`: regenerated (delta between rpkg and templates should be unchanged from before — the new block lives in both ends).
- `rpkg/configure`: regenerated (substantial line-count delta but mechanical from `autoconf`).
- Optional new fixture under `rpkg/src/rust/` or `tests/`: small C file (≤10 lines) + a `build.rs` snippet that uses `cc::Build::new()`.

## Acceptance

- [ ] Configure derives a curated, ABI-only subset of CFLAGS/CXXFLAGS from R config.
- [ ] Emitted as `CFLAGS_<target>` / `CXXFLAGS_<target>` in `[env]`, only when non-empty after filtering.
- [ ] Documented which flags are kept vs stripped, with rationale (this issue body — the allowlist table above is the source of truth).
- [ ] Regression test: a `build.rs` invoking `cc::Build::new()` picks up the expected `-march` value end-to-end.
- [ ] `just templates-check` passes; `rpkg/configure` regenerated.

## Open questions

- **Key casing**: stick with hyphenated-lowercase (`CFLAGS_aarch64-apple-darwin`) to match existing `AR_<target>` style, or migrate all four to canonical upper-with-underscores (`CFLAGS_AARCH64_APPLE_DARWIN`)? Cargo accepts both; consistency within the file matters more than which one we pick. Recommend: stay hyphenated for now; file a follow-up cleanup issue if anyone hits a tooling rough edge.
- **`-fopenmp` keep**: arguably this should be a Linux-only keep (macOS clang ignores `-fopenmp` without libomp, Windows uses different flag spelling). For now keep it uniformly — if it's in R's CFLAGS, R successfully uses it, so the host toolchain supports it. Revisit if a user reports macOS clang breakage.
- **Conda-Forge `-march=nocona`**: conda's pinned baselines may be older than the local CPU. Keep — that's exactly the ABI-matching point. Users on conda already accept this baseline.

Refs: PR #601 (PR1 — `[env]` block introduction), #597 (LTO mirror — sibling follow-up).

CGMossa:
Closes #598. Follow-up to PR #601 (the initial `[env]` block introduction).

Implements the strict ABI-only allowlist filter on R's `CFLAGS`/`CXXFLAGS` and emits the survivors as `CFLAGS_<target>` / `CXXFLAGS_<target>` in `src/rust/.cargo/config.toml`'s `[env]` table. Every transitive C/C++ build script in the cargo dep graph (`cc`, `cmake`, `bindgen`, ...) now sees R's ABI baseline — and nothing else.

<details>
<summary>AI-written rationale, validation evidence, and the full allowlist table</summary>

## Allowlist (the source of truth)

| Flag family | Decision | Rationale |
|---|---|---|
| `-march=*` | **KEEP** | Instruction-set baseline; ABI-affecting for SIMD intrinsics and vector struct layout. Mismatch → SIGILL or silent miscompiles. |
| `-mtune=*` | **KEEP** | Not strictly ABI but commonly paired with `-march`; harmless if R picked a value. |
| `-mfpmath=*`, `-mfpu=*` | **KEEP** | Float-ABI on 32-bit ARM and some legacy x86 (FPU-vs-SSE) — load-bearing where present. |
| `-msse*`, `-mavx*`, `-mssse*`, `-msse4*`, `-mpclmul`, `-maes` | **KEEP** | Same as `-march`. Matches what conda-forge / vendor channels actually set. |
| `-mabi=*` | **KEEP** | RISC-V / ARM hard-float vs soft-float ABI. Critical where present. |
| `-mfloat-abi=*` | **KEEP** | ARM hard/soft float — explicit ABI. |
| `-pthread` | **KEEP** | Thread-local storage layout / pthread struct sizes on glibc. |
| `-D_REENTRANT` | **KEEP** | Same family as `-pthread`; some platforms use it instead. |
| `-fopenmp`, `-fopenmp=*` | **KEEP** | OpenMP runtime presence affects struct layout and linkage. |
| `-I*` | **DROP** | R-specific include paths. Cargo deps must not see `R.h`/`Rinternals.h` by accident. |
| `-isystem*` | **DROP** | Same. |
| `-D_*` (other than `-D_REENTRANT`) | **DROP** | R-private feature-test macros (`_FORTIFY_SOURCE`, `_FILE_OFFSET_BITS`, ...). |
| `-DNDEBUG` | **DROP** | cargo's release profile handles this via `debug-assertions`. |
| `-W*` | **DROP** | Warning flags are not ABI. `R CMD check --as-cran` already objects to `-W*` in `PKG_CFLAGS`. |
| `-O*` | **DROP** | Optimisation level is a cargo profile concern. |
| `-g*` | **DROP** | Debug info is a cargo profile concern. |
| `-fPIC`, `-fpic` | **DROP** | Cargo already passes the right PIC flag for staticlib/cdylib. |
| `-fstack-protector*` | **DROP** | Hardening, not ABI. |
| `-fdebug-prefix-map=*`, `-ffile-prefix-map=*` | **DROP** | Build-reproducibility flags; meaningless for cargo deps. |
| `-fno-plt`, `-fno-stack-clash-protection`, `-fexceptions` | **DROP** | Codegen knobs, not ABI for staticlib/cdylib boundary. |
| `-arch *` (macOS) | **DROP** | Covered by the `*_<target>` env var name itself. |
| Everything else | **DROP** | Conservative default. |

Allowlist regex (POSIX BRE-friendly, no character classes so no m4 bracket escaping needed):
```
^-march=|^-mtune=|^-mfpmath=|^-mfpu=|^-mfloat-abi=|^-mabi=|^-msse|^-mavx|^-mssse|^-mpclmul|^-maes|^-pthread$|^-D_REENTRANT$|^-fopenmp
```

`-pthread$` and `-D_REENTRANT$` are anchored so they don't match longer prefixes like `-pthread-foo` or `-D_REENTRANTLY` (paranoid, but cheap).

## Filter pipeline

```sh
echo "$_r_cflags" | tr ' ' '\n' | grep -E "$_abi_allow_re" | tr '\n' ' ' | $SED 's/ *$//'
```

Chosen over a single `sed` because (a) sed alternation portability is patchy across BSD/GNU; (b) `grep -E` is universally available and matches one whole token at a time; (c) trailing-space cleanup keeps the emitted TOML tidy.

## Key casing decision

Cargo accepts both `CFLAGS_aarch64-apple-darwin` and `CFLAGS_AARCH64_APPLE_DARWIN` (cc-rs reads both). To stay consistent with the existing `AR_<target>` / `RANLIB_<target>` emission style in `emit_env_block`, we use the hyphenated-lowercase form (`$CARGO_ENV_RUST_TARGET`). Verified empirically — cc-rs's build-script output enumerates **both** forms when probing env:
```
cargo:rerun-if-env-changed=CFLAGS_aarch64_apple_darwin
cargo:rerun-if-env-changed=CFLAGS_aarch64-apple-darwin
```

## Filter unit tests (in-tree, no fixture needed)

Seven synthetic inputs reproduced via the same `tr | grep -E | tr | sed` pipe:

| Input shape | Output |
|---|---|
| Debian/Ubuntu CFLAGS (only `-mtune=generic`) | `-mtune=generic` |
| Conda-forge `-march=nocona -mtune=haswell` | `-march=nocona -mtune=haswell` |
| Rocker `-mavx2 -fopenmp -pthread -D_REENTRANT` | `-mavx2 -fopenmp -pthread -D_REENTRANT` |
| macOS homebrew (only `-I/opt/homebrew/include`) | `` (empty — no key emitted) |
| ARM `-mfloat-abi=hard -mfpu=neon -march=armv7-a` | `-mfloat-abi=hard -mfpu=neon -march=armv7-a` |
| `-D_REENTRANT -D_REENTRANTLY -D_REENTRANT_FOO` | `-D_REENTRANT` (anchored) |
| `-pthread -pthread-foo` | `-pthread` (anchored) |

## End-to-end `cc::Build` propagation evidence

A throwaway `cc-probe` crate (`cargo` + `build.rs` + `cc::Build::new().file("src/foo.c").compile("foo")`) was built three ways with `CC_ENABLE_DEBUG_OUTPUT=1` to capture the `cc` invocation cc-rs actually emits:

**Baseline (no `CFLAGS_*` set)**:
```
[cc-probe 0.0.0] CFLAGS_aarch64-apple-darwin = None
[cc-probe 0.0.0] running: env -u IPHONEOS_DEPLOYMENT_TARGET LC_ALL="C" "cc" "-O0" "-ffunction-sections" ... "-mmacosx-version-min=26.5" "-Wall" "-Wextra" "-o" ... "-c" "src/foo.c"
```
No `-march=`. cc-rs defaults to `-Wall -Wextra`.

**Override via env var `CFLAGS_aarch64-apple-darwin=-march=armv8.5-a`**:
```
[cc-probe 0.0.0] CFLAGS_aarch64-apple-darwin = Some(-march=armv8.5-a)
[cc-probe 0.0.0] running: env -u IPHONEOS_DEPLOYMENT_TARGET LC_ALL="C" "cc" "-O0" ... "-mmacosx-version-min=26.5" "-march=armv8.5-a" "-o" ... "-c" "src/foo.c"
```
`-march=armv8.5-a` reaches `cc`. Note `-Wall -Wextra` is replaced by user flags (validates dropping `-W*` from the allowlist — we don't want to re-inject distro warning flags).

**Via the actual code path — `.cargo/config.toml [env]`**:
```toml
[env]
"CFLAGS_aarch64-apple-darwin" = "-march=armv8.5-a -mtune=generic -fopenmp"
```
```
[cc-probe 0.0.0] CFLAGS_aarch64-apple-darwin = Some(-march=armv8.5-a -mtune=generic -fopenmp)
[cc-probe 0.0.0] running: env ... "cc" "-O0" ... "-march=armv8.5-a" "-mtune=generic" "-fopenmp" "-o" ... "-c" "src/foo.c"
```
All three tokens flow through cargo `[env]` → process env → cc-rs shlex → `cc` argv. End-to-end confirmed.

(Cleanup: the `cc-probe` crate lives only at `/tmp/598-fixture/` and is **not** added to the tree. Adding a permanent build-script test in `rpkg/` would entangle the rpkg build with a custom Cargo workspace member that doesn't earn its keep — the filter logic itself is shell, and the propagation is owned by cargo+cc-rs.)

## Configure-time integration test

Synthetic R shim returning CRAN-Debian-style CFLAGS:
```
R CMD config CFLAGS  → -I"/usr/share/R/include" -DNDEBUG -fpic -g -O2 \
  -fdebug-prefix-map=… -fstack-protector-strong -Wformat -Werror=format-security \
  -Wdate-time -D_FORTIFY_SOURCE=2 -Wall -mtune=generic -march=haswell -mavx2 \
  -fopenmp -pthread
```

`R_HOME=/tmp/598-fake-r bash ./configure` →
```
configure: cargo CFLAGS_<target>  = -mtune=generic -march=haswell -mavx2 -fopenmp -pthread
configure: cargo CXXFLAGS_<target>= -mtune=generic -march=haswell -fopenmp
```

Resulting `src/rust/.cargo/config.toml [env]`:
```toml
CFLAGS_aarch64-apple-darwin = "-mtune=generic -march=haswell -mavx2 -fopenmp -pthread"
CXXFLAGS_aarch64-apple-darwin = "-mtune=generic -march=haswell -fopenmp"
```

Real `R CMD config CFLAGS` on this host (homebrew R 4.6 / macOS arm64):
```
$ R CMD config CFLAGS
-I/opt/homebrew/include
```
Filter strips to empty → no `CFLAGS_*` key emitted in the generated config.

## Validation summary

- [x] `just templates-check` — clean (exit 0)
- [x] `bash ./configure` — exit 0 on this host (homebrew R, no CFLAGS_* emitted as expected)
- [x] `bash ./configure` with synthetic CRAN-style CFLAGS shim — emits expected filtered values
- [x] `just rcmdinstall` — `* DONE (miniextendr)` end-to-end
- [x] cc::Build fixture proves `-march=` flows from `[env]` → process env → cc-rs → `cc` argv

## Out of scope (intentionally)

- `LDFLAGS_<target>` (R's `LDFLAGS` would need a similar but different allowlist; defer if/when users surface a real need).
- LTO propagation (#597).
- `force = true` (mirrors PR #601's no-force-floor stance — shell env always wins).
- A permanent in-tree `build.rs` cc-rs regression test (not earning its keep; the throwaway `cc-probe` evidence is in this PR body, and the filter is a tightly-bounded shell pipeline).

## Files touched

- `rpkg/configure.ac` — derivation block (+25 lines), `emit_env_block` extension (+6), variable-pass list (+2).
- `rpkg/configure` — regenerated via `autoconf`.
- `minirextendr/inst/templates/rpkg/configure.ac` — mirror.
- `minirextendr/inst/templates/monorepo/rpkg/configure.ac` — mirror.
- `patches/templates.patch` — regenerated via `just templates-approve` (line-number drift only).

</details>

### DIFF
--- a/minirextendr/inst/templates/monorepo/rpkg/configure.ac
+++ b/minirextendr/inst/templates/monorepo/rpkg/configure.ac
@@ -470,6 +470,38 @@ dnl when R's values differ — otherwise we'd just add noise to the config.
 CARGO_ENV_R_AR=`"${R_HOME}/bin/R" CMD config AR 2>/dev/null`
 CARGO_ENV_R_RANLIB=`"${R_HOME}/bin/R" CMD config RANLIB 2>/dev/null`
 
+dnl --- CFLAGS / CXXFLAGS: ABI-only allowlist from R config ----
+dnl Cargo applies CFLAGS_<target> and CXXFLAGS_<target> to *every* transitive
+dnl C/C++ build script (cc, cmake, etc.). R's raw CFLAGS contain R-specific
+dnl -I paths, R-private defines, and warning flags that have no business in
+dnl cargo deps. We allowlist a small set of ABI-affecting flags; everything
+dnl else is stripped. See issue #598 for the full rationale table.
+CARGO_ENV_ABI_CFLAGS=""
+CARGO_ENV_ABI_CXXFLAGS=""
+
+_r_cflags=`"${R_HOME}/bin/R" CMD config CFLAGS 2>/dev/null`
+_r_cxxflags=`"${R_HOME}/bin/R" CMD config CXXFLAGS 2>/dev/null`
+
+dnl Allowlist regex. Keep ABI-affecting flags only:
+dnl   -march=, -mtune=, -mfpmath=, -mfpu=, -mfloat-abi=, -mabi=, -m{sse,avx,ssse,pclmul,aes},
+dnl   -pthread, -D_REENTRANT, -fopenmp[=…]
+dnl No character classes are used (no `[...]`), so no m4 bracket escaping needed.
+_abi_allow_re='^-march=|^-mtune=|^-mfpmath=|^-mfpu=|^-mfloat-abi=|^-mabi=|^-msse|^-mavx|^-mssse|^-mpclmul|^-maes|^-pthread$|^-D_REENTRANT$|^-fopenmp'
+
+if test -n "$_r_cflags"; then
+  CARGO_ENV_ABI_CFLAGS=`echo "$_r_cflags" | tr ' ' '\n' | grep -E "$_abi_allow_re" | tr '\n' ' ' | $SED 's/ *$//'`
+fi
+if test -n "$_r_cxxflags"; then
+  CARGO_ENV_ABI_CXXFLAGS=`echo "$_r_cxxflags" | tr ' ' '\n' | grep -E "$_abi_allow_re" | tr '\n' ' ' | $SED 's/ *$//'`
+fi
+
+if test -n "$CARGO_ENV_ABI_CFLAGS"; then
+  AC_MSG_NOTICE([cargo CFLAGS_<target>  = $CARGO_ENV_ABI_CFLAGS])
+fi
+if test -n "$CARGO_ENV_ABI_CXXFLAGS"; then
+  AC_MSG_NOTICE([cargo CXXFLAGS_<target>= $CARGO_ENV_ABI_CXXFLAGS])
+fi
+
 dnl Individual env-derivation variables are propagated to AC_CONFIG_COMMANDS
 dnl via the trailing variable-pass list, and the [env] block body is composed
 dnl inline inside that shell context. Pre-composing here and passing a single
@@ -583,6 +615,12 @@ AC_CONFIG_COMMANDS([cargo-config],
     if test -n "$CARGO_ENV_R_RANLIB" && test "$CARGO_ENV_R_RANLIB" != "ranlib"; then
       echo "RANLIB_$CARGO_ENV_RUST_TARGET = \"$CARGO_ENV_R_RANLIB\""
     fi
+    if test -n "$CARGO_ENV_ABI_CFLAGS"; then
+      echo "CFLAGS_$CARGO_ENV_RUST_TARGET = \"$CARGO_ENV_ABI_CFLAGS\""
+    fi
+    if test -n "$CARGO_ENV_ABI_CXXFLAGS"; then
+      echo "CXXFLAGS_$CARGO_ENV_RUST_TARGET = \"$CARGO_ENV_ABI_CXXFLAGS\""
+    fi
     echo ""
   }
 
@@ -655,6 +693,8 @@ AC_CONFIG_COMMANDS([cargo-config],
  CARGO_ENV_RUST_TARGET_UPPER="$CARGO_ENV_RUST_TARGET_UPPER"
  CARGO_ENV_R_AR="$CARGO_ENV_R_AR"
  CARGO_ENV_R_RANLIB="$CARGO_ENV_R_RANLIB"
+ CARGO_ENV_ABI_CFLAGS="$CARGO_ENV_ABI_CFLAGS"
+ CARGO_ENV_ABI_CXXFLAGS="$CARGO_ENV_ABI_CXXFLAGS"
  SED="$SED"])
 
 dnl ---- 3) Cargo.lock shape check (tarball mode only) ----
--- a/minirextendr/inst/templates/rpkg/configure.ac
+++ b/minirextendr/inst/templates/rpkg/configure.ac
@@ -522,6 +522,38 @@ dnl when R's values differ — otherwise we'd just add noise to the config.
 CARGO_ENV_R_AR=`"${R_HOME}/bin/R" CMD config AR 2>/dev/null`
 CARGO_ENV_R_RANLIB=`"${R_HOME}/bin/R" CMD config RANLIB 2>/dev/null`
 
+dnl --- CFLAGS / CXXFLAGS: ABI-only allowlist from R config ----
+dnl Cargo applies CFLAGS_<target> and CXXFLAGS_<target> to *every* transitive
+dnl C/C++ build script (cc, cmake, etc.). R's raw CFLAGS contain R-specific
+dnl -I paths, R-private defines, and warning flags that have no business in
+dnl cargo deps. We allowlist a small set of ABI-affecting flags; everything
+dnl else is stripped. See issue #598 for the full rationale table.
+CARGO_ENV_ABI_CFLAGS=""
+CARGO_ENV_ABI_CXXFLAGS=""
+
+_r_cflags=`"${R_HOME}/bin/R" CMD config CFLAGS 2>/dev/null`
+_r_cxxflags=`"${R_HOME}/bin/R" CMD config CXXFLAGS 2>/dev/null`
+
+dnl Allowlist regex. Keep ABI-affecting flags only:
+dnl   -march=, -mtune=, -mfpmath=, -mfpu=, -mfloat-abi=, -mabi=, -m{sse,avx,ssse,pclmul,aes},
+dnl   -pthread, -D_REENTRANT, -fopenmp[=…]
+dnl No character classes are used (no `[...]`), so no m4 bracket escaping needed.
+_abi_allow_re='^-march=|^-mtune=|^-mfpmath=|^-mfpu=|^-mfloat-abi=|^-mabi=|^-msse|^-mavx|^-mssse|^-mpclmul|^-maes|^-pthread$|^-D_REENTRANT$|^-fopenmp'
+
+if test -n "$_r_cflags"; then
+  CARGO_ENV_ABI_CFLAGS=`echo "$_r_cflags" | tr ' ' '\n' | grep -E "$_abi_allow_re" | tr '\n' ' ' | $SED 's/ *$//'`
+fi
+if test -n "$_r_cxxflags"; then
+  CARGO_ENV_ABI_CXXFLAGS=`echo "$_r_cxxflags" | tr ' ' '\n' | grep -E "$_abi_allow_re" | tr '\n' ' ' | $SED 's/ *$//'`
+fi
+
+if test -n "$CARGO_ENV_ABI_CFLAGS"; then
+  AC_MSG_NOTICE([cargo CFLAGS_<target>  = $CARGO_ENV_ABI_CFLAGS])
+fi
+if test -n "$CARGO_ENV_ABI_CXXFLAGS"; then
+  AC_MSG_NOTICE([cargo CXXFLAGS_<target>= $CARGO_ENV_ABI_CXXFLAGS])
+fi
+
 dnl Individual env-derivation variables are propagated to AC_CONFIG_COMMANDS
 dnl via the trailing variable-pass list, and the [env] block body is composed
 dnl inline inside that shell context. Pre-composing here and passing a single
@@ -635,6 +667,12 @@ AC_CONFIG_COMMANDS([cargo-config],
     if test -n "$CARGO_ENV_R_RANLIB" && test "$CARGO_ENV_R_RANLIB" != "ranlib"; then
       echo "RANLIB_$CARGO_ENV_RUST_TARGET = \"$CARGO_ENV_R_RANLIB\""
     fi
+    if test -n "$CARGO_ENV_ABI_CFLAGS"; then
+      echo "CFLAGS_$CARGO_ENV_RUST_TARGET = \"$CARGO_ENV_ABI_CFLAGS\""
+    fi
+    if test -n "$CARGO_ENV_ABI_CXXFLAGS"; then
+      echo "CXXFLAGS_$CARGO_ENV_RUST_TARGET = \"$CARGO_ENV_ABI_CXXFLAGS\""
+    fi
     echo ""
   }
 
@@ -707,6 +745,8 @@ AC_CONFIG_COMMANDS([cargo-config],
  CARGO_ENV_RUST_TARGET_UPPER="$CARGO_ENV_RUST_TARGET_UPPER"
  CARGO_ENV_R_AR="$CARGO_ENV_R_AR"
  CARGO_ENV_R_RANLIB="$CARGO_ENV_R_RANLIB"
+ CARGO_ENV_ABI_CFLAGS="$CARGO_ENV_ABI_CFLAGS"
+ CARGO_ENV_ABI_CXXFLAGS="$CARGO_ENV_ABI_CXXFLAGS"
  SED="$SED"])
 
 dnl ---- 3) Cargo.lock shape check (tarball mode only) ----
--- a/patches/templates.patch
+++ b/patches/templates.patch
@@ -1,6 +1,6 @@
 diff -ruN -U2 a/monorepo/rpkg/bootstrap.R b/monorepo/rpkg/bootstrap.R
---- a/monorepo/rpkg/bootstrap.R	2026-05-17 13:14:25
-+++ b/monorepo/rpkg/bootstrap.R	2026-05-17 13:14:25
+--- a/monorepo/rpkg/bootstrap.R	2026-05-19 21:13:10
++++ b/monorepo/rpkg/bootstrap.R	2026-05-19 21:13:10
 @@ -1,119 +1,11 @@
 -# bootstrap.R - Run before package build (Config/build/bootstrap: TRUE).
 -# Invoked by pkgbuild (devtools::build, r-lib/actions/check-r-package) in
@@ -130,8 +130,8 @@ diff -ruN -U2 a/monorepo/rpkg/bootstrap.R b/monorepo/rpkg/bootstrap.R
  
  if (.Platform$OS.type == "windows") {
 diff -ruN -U2 a/monorepo/rpkg/configure.ac b/monorepo/rpkg/configure.ac
---- a/monorepo/rpkg/configure.ac	2026-05-17 13:41:31
-+++ b/monorepo/rpkg/configure.ac	2026-05-17 13:23:10
+--- a/monorepo/rpkg/configure.ac	2026-05-20 00:23:16
++++ b/monorepo/rpkg/configure.ac	2026-05-20 00:24:33
 @@ -1,3 +1,3 @@
 -AC_INIT([miniextendr], [0.1.0])
 +AC_INIT([{{package}}], [0.1.0])
@@ -367,20 +367,20 @@ diff -ruN -U2 a/monorepo/rpkg/configure.ac b/monorepo/rpkg/configure.ac
 +dnl as CARGO_TARGET_DIR above).
  VENDOR_OUT="$abs_top_srcdir/vendor"
  VENDOR_OUT_CARGO="$abs_top_srcdir_cargo/vendor"
-@@ -500,4 +478,5 @@
+@@ -532,4 +510,5 @@
  
  dnl ---- output files ----
 +dnl Create .cargo directory (it's a hidden dir not included in tarball)
  dnl AC_CONFIG_FILES handles Makevars and the win.def. The cargo config is
  dnl written by AC_CONFIG_COMMANDS below because its contents differ across
-@@ -716,5 +695,5 @@
+@@ -756,5 +735,5 @@
      if test "$CARGO_MAJOR" -lt 1 || \
         (test "$CARGO_MAJOR" -eq 1 && test "$CARGO_MINOR" -lt 78); then
 -      echo "configure: WARNING: cargo $CARGO_VERSION cannot read Cargo.lock v$LOCKFILE_VERSION; regenerating" >&2
 +      echo "configure: WARNING: cargo $CARGO_VERSION cannot read Cargo.lock v$LOCKFILE_VERSION; regenerating lockfile" >&2
        rm -f "$LOCKFILE_PATH"
        $CARGO_CMD generate-lockfile --manifest-path src/rust/Cargo.toml $CARGO_OFFLINE_FLAG
-@@ -732,13 +711,22 @@
+@@ -772,13 +751,22 @@
  AC_CONFIG_COMMANDS([post-config],
  [
 +  dnl Ensure Cargo.lock exists (needed by cargo for reproducible builds)
@@ -406,8 +406,8 @@ diff -ruN -U2 a/monorepo/rpkg/configure.ac b/monorepo/rpkg/configure.ac
  
  AC_OUTPUT
 diff -ruN -U2 a/rpkg/configure.ac b/rpkg/configure.ac
---- a/rpkg/configure.ac	2026-05-17 13:41:31
-+++ b/rpkg/configure.ac	2026-05-17 13:21:51
+--- a/rpkg/configure.ac	2026-05-20 00:23:16
++++ b/rpkg/configure.ac	2026-05-20 00:23:57
 @@ -1,3 +1,3 @@
 -AC_INIT([miniextendr], [0.1.0])
 +AC_INIT([{{package}}], [0.1.0])
@@ -589,20 +589,20 @@ diff -ruN -U2 a/rpkg/configure.ac b/rpkg/configure.ac
 +dnl as CARGO_TARGET_DIR above).
  VENDOR_OUT="$abs_top_srcdir/vendor"
  VENDOR_OUT_CARGO="$abs_top_srcdir_cargo/vendor"
-@@ -500,4 +530,5 @@
+@@ -532,4 +562,5 @@
  
  dnl ---- output files ----
 +dnl Create .cargo directory (it's a hidden dir not included in tarball)
  dnl AC_CONFIG_FILES handles Makevars and the win.def. The cargo config is
  dnl written by AC_CONFIG_COMMANDS below because its contents differ across
-@@ -716,5 +747,5 @@
+@@ -756,5 +787,5 @@
      if test "$CARGO_MAJOR" -lt 1 || \
         (test "$CARGO_MAJOR" -eq 1 && test "$CARGO_MINOR" -lt 78); then
 -      echo "configure: WARNING: cargo $CARGO_VERSION cannot read Cargo.lock v$LOCKFILE_VERSION; regenerating" >&2
 +      echo "configure: WARNING: cargo $CARGO_VERSION cannot read Cargo.lock v$LOCKFILE_VERSION; regenerating lockfile" >&2
        rm -f "$LOCKFILE_PATH"
        $CARGO_CMD generate-lockfile --manifest-path src/rust/Cargo.toml $CARGO_OFFLINE_FLAG
-@@ -732,13 +763,22 @@
+@@ -772,13 +803,22 @@
  AC_CONFIG_COMMANDS([post-config],
  [
 +  dnl Ensure Cargo.lock exists (needed by cargo for reproducible builds)
--- a/rpkg/configure
+++ b/rpkg/configure
@@ -2720,6 +2720,30 @@ esac
 CARGO_ENV_R_AR=`"${R_HOME}/bin/R" CMD config AR 2>/dev/null`
 CARGO_ENV_R_RANLIB=`"${R_HOME}/bin/R" CMD config RANLIB 2>/dev/null`
 
+CARGO_ENV_ABI_CFLAGS=""
+CARGO_ENV_ABI_CXXFLAGS=""
+
+_r_cflags=`"${R_HOME}/bin/R" CMD config CFLAGS 2>/dev/null`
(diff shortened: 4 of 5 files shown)


CGMossa (commented):
Reviewed against issue #598 (planner spec) and reproduced the filter + configure run locally. Implementation follows the planner's design point-for-point; templates are in sync; `just templates-check` passes; the cc-rs propagation evidence in the PR body is plausible and the configure-time emit reproduces here with a synthetic R shim.

## Verification

### Allowlist regex matches the planner's spec verbatim

```
^-march=|^-mtune=|^-mfpmath=|^-mfpu=|^-mfloat-abi=|^-mabi=|^-msse|^-mavx|^-mssse|^-mpclmul|^-maes|^-pthread$|^-D_REENTRANT$|^-fopenmp
```

All seven documented filter cases reproduce on this host:

| Input | Filtered output |
|---|---|
| Debian CRAN-style (`-DNDEBUG -fpic -g -O2 -Wall -mtune=generic ...`) | `-mtune=generic` |
| Conda-forge (`-march=nocona -mtune=haswell ...`) | `-march=nocona -mtune=haswell` |
| Rocker (`-mavx2 -fopenmp -pthread -D_REENTRANT`) | (pass-through) |
| Homebrew (`-I/opt/homebrew/include`) | (empty) |
| ARM (`-mfloat-abi=hard -mfpu=neon -march=armv7-a`) | (pass-through) |
| `-D_REENTRANT -D_REENTRANTLY -D_REENTRANT_FOO` | `-D_REENTRANT` (anchored) |
| `-pthread -pthread-foo` | `-pthread` (anchored) |

### Filter pipeline edge cases

- Empty input → guarded by `test -n` → no key emitted. Confirmed.
- Single token, no match → backtick captures empty string, downstream guard works. `grep`'s nonzero exit doesn't escape because there's no `set -e` (verified `configure.ac` doesn't enable it).
- Trailing whitespace in input → `tr | grep | tr | sed 's/ *$//'` strips it.
- Multi-space tokens → `tr ' ' '\n'` collapses empty splits via subsequent `grep` (empty lines never match the anchored regex).

### m4 escaping

The regex contains no `[...]` character classes — confirmed by inspection. The `AC_MSG_NOTICE([cargo CFLAGS_<target>...])` body uses `<target>` not `[target]`, so no bracket-quoting needed. `autoconf`'s expansion (`rpkg/configure:2739-2744`) preserves the regex single-quoted exactly as written and converts `AC_MSG_NOTICE` to the standard 5/6-fd `printf` form. No m4 hazards observed.

### Insertion site

`emit_env_block()` body in `rpkg/configure.ac:617-646`:

```
[env]
... existing MACOS / WIN_LINKER / FETCH_WITH_CLI / AR / RANLIB lines ...
+ CFLAGS_$CARGO_ENV_RUST_TARGET = "..."
+ CXXFLAGS_$CARGO_ENV_RUST_TARGET = "..."
echo ""    # terminating blank
```

The new conditional `echo` lines sit before the function's final `echo ""` separator and before any later `[…]` heading would close `[env]`. TOML table scoping is preserved.

### Variable propagation

`AC_CONFIG_COMMANDS([cargo-config], ..., [..., CARGO_ENV_ABI_CFLAGS="$CARGO_ENV_ABI_CFLAGS", CARGO_ENV_ABI_CXXFLAGS="$CARGO_ENV_ABI_CXXFLAGS", ...])` at `rpkg/configure.ac:717-718`. Variables reach the function scope correctly — verified the regenerated `rpkg/configure:3437-3438` carries them through.

### Templates and patch sync

- `rpkg/configure.ac`, `minirextendr/inst/templates/rpkg/configure.ac`, and `minirextendr/inst/templates/monorepo/rpkg/configure.ac` all contain the same 32 added lines (block + emit + propagation).
- Verified content identity of the new block across all three files.
- `just templates-check` → exit 0.

### Configure regeneration

`rpkg/configure` regenerated cleanly via autoconf. Diff is mechanical (regex passed through verbatim, `AC_MSG_NOTICE` expanded to the standard dual-fd form). No `_LT_FOREACH`/m4sh churn beyond the directly-added lines + downstream `as_lineno` line-number drift.

### Key casing

Hyphenated-lowercase `CFLAGS_aarch64-apple-darwin` matches the existing `AR_<target>` / `RANLIB_<target>` emission style. Consistent within the file. Cargo accepts both forms (cc-rs `rerun-if-env-changed` enumerates both). The planner's recommendation to "stay hyphenated for now" is honored.

### End-to-end via shim

With a synthetic `R` shim returning CRAN-Debian-style CFLAGS:

```
configure: cargo CFLAGS_<target>  = -mtune=generic -march=haswell -mavx2 -fopenmp -pthread
configure: cargo CXXFLAGS_<target>= -mtune=generic -march=haswell -fopenmp
```

Matches the PR body's documented output exactly. The generated `src/rust/.cargo/config.toml [env]` section emits:

```
CFLAGS_aarch64-apple-darwin = "-mtune=generic -march=haswell -mavx2 -fopenmp -pthread"
CXXFLAGS_aarch64-apple-darwin = "-mtune=generic -march=haswell -fopenmp"
```

With the host's real R (returns only `-I/opt/homebrew/include`), the filter strips to empty and no `CFLAGS_*` key is emitted — `test -n` guard works.

### cc-rs propagation chain

The PR body documents a `cc-probe` throwaway crate showing `[env]` → cargo → process env → cc-rs → `cc` argv with `-march=armv8.5-a`. The chain itself is cargo+cc-rs's documented behavior (per-target `CFLAGS_<target>` is the canonical hook). The chain is not re-tested here, but the configure-time emit is what this PR is on the hook for, and that path is now fully verified.

## Notes / observations (non-blocking)

- `-fopenmp` is unanchored, so it also matches `-fopenmp-simd` and `-fopenmp=libomp`. Both are OpenMP-family flags, so this is consistent with the planner's "if R has it, R successfully uses it" rationale. Worth flagging for future readers but not a defect.
- The PR body's claim that no permanent in-tree `cc::Build` fixture is added is borne out — diff is exactly the four files listed in the issue plan plus the regenerated `configure`. The trade-off (filter is shell, propagation is cargo+cc-rs, both well-tested elsewhere) is reasonable for a tightly-bounded change.
- Out-of-scope items (LDFLAGS, LTO #597, `force = true`, per-platform allowlist tuning) are explicitly preserved as deferred — no scope creep.

## Verdict

LGTM. No blockers. Implementation matches the planner's design, filter behavior is correct across documented and edge cases, templates are in sync, regenerated `configure` is clean, and the configure-time emit reproduces end-to-end. Suggest the implementer also verify CI green on at least one Linux runner (where R's CFLAGS will actually carry non-trivial content like `-mtune=generic`) before merge, since the local host's homebrew R filters to empty and exercises only the guard path.

PR #661, opened 2026-05-19
fix(connections): make custom-connection tests pass

### LINKED ISSUE #568: fix(connections): custom connection tests fail when connections feature is enabled

CGMossa:
Pre-existing `RCustomConnection::build` bugs caused all 12 custom-connection tests in `rpkg/tests/testthat/test-connections.R` to skip via `skip_custom_conn_bug()`. This PR fixes the underlying issues so the tests run end-to-end with `FAIL 0`.

## Root causes fixed

`R_new_custom_connection` (R's stable C entry point) returns a connection with a defaults block that is wrong for nearly every custom impl:

```
isopen   = FALSE
text     = TRUE     (regardless of mode)
blocking = FALSE
canread  = canwrite = TRUE  /* "in principle" */
vfprintf = null_vfprintf   (raises "Error writing to connection")
```

`RCustomConnection::build` now papers over each of these:

1. **Pre-open.** R auto-opens on some paths (`readLines`) but not on `writeBin` / `writeLines` / direct `seek`. The builder invokes the registered `open` trampoline before returning, with destroy-on-failure cleanup. R's later auto-open path sees `isopen == TRUE` and short-circuits — no double-open.
2. **Text inference.** `text` is inferred from the mode string when the caller didn't set `.text(...)`. `'b'` in mode → binary (`text = FALSE`); else text. Matches R's stock `file()` behavior.
3. **`vfprintf` fallback.** Our trampoline now mirrors R's own `dummy_vfprintf`: format the args via the host `vsnprintf` into a 10 KiB stack buffer, then forward the bytes to the user's `write` callback. The trait method still exists for custom impls that want to override.
4. **Blocking default.** Default `blocking = true`. `do_readLines` treats `text && !blocking` as a non-blocking text stream and pushes back the final incomplete line on EOF, silently dropping a trailing line without `\n`. The builder now matches R's `file()` / `url()`.

## Diff shape

- `miniextendr-api/src/connection.rs`: ~95 lines of changes inside `RCustomConnection::build`, the `vfprintf_trampoline`, and the `vfprintf` trait docstring.
- `rpkg/tests/testthat/test-connections.R`: drop `skip_custom_conn_bug` shim + 12 call sites (-22 LoC).
- Doc updates: `.claude/skills/miniextendr-connections/SKILL.md` and `miniextendr-api/CLAUDE.md` reflect that `build()` returns an open, mode-aware connection.

## Test plan

- [x] `cargo check -p miniextendr-api --features connections` — passes.
- [x] `just rcmdinstall` — clean install.
- [x] `just devtools-test` — `FAIL 0 | WARN 7 | SKIP 16 | PASS 5912`. The 7 warnings are R's expected `incomplete final line found on '<conn>'` for tests whose inputs intentionally lack trailing `\n`.
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean.
- [ ] CI to run `clippy_all` with the curated feature set.

Closes #568

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.claude/skills/miniextendr-connections/SKILL.md
+++ b/.claude/skills/miniextendr-connections/SKILL.md
@@ -11,7 +11,7 @@ R's connection system provides an I/O abstraction used by `file()`, `url()`, `gz
 
 - "How do I create a custom R connection in Rust?"
 - "What is `RConnectionImpl`?"
-- "Why does my connection start closed and have to be explicitly opened?"
+- "Why is my connection returned already-open, and where does the open callback get invoked?"
 - "How do panics inside connection callbacks get handled?"
 - "What is `catch_connection_panic`?"
 - "What is the connections ABI version check?"
@@ -26,22 +26,24 @@ The `connections` feature in miniextendr is gated for this reason. Before using 
 
 The API requires R >= 4.3.0. Use `check_runtime_connections_support()` for a runtime probe.
 
-### Connections start in CLOSED state
+### Connections from `RCustomConnection::build` are returned open
 
-`R_new_custom_connection` creates the connection in a closed state (`isopen = FALSE`). After building the connection SEXP, you must explicitly invoke the `open` callback to put it into an open state before returning it to R callers who expect an open connection.
+`R_new_custom_connection` (R's underlying C entry point) returns a connection in the CLOSED state (`isopen = FALSE`) with `text = TRUE` regardless of mode. `RCustomConnection::build` papers over both:
 
-The builder's `.build(impl)` method does not automatically call `open`. If you want an auto-opened connection:
+- The mode string is parsed for `'b'` to infer `text`; explicit `.text(...)` still wins.
+- The `open` callback is invoked before the SEXP is returned. R's own auto-open path (used by `readLines` / `scan` / etc.) sees `isopen == TRUE` and short-circuits, so there is no double-open.
 
 ```rust
 let conn_sexp = RCustomConnection::new()
     .description("my conn")
+    .mode("r+b")
     .can_read(true)
+    .can_write(true)
     .build(MyConnectionImpl { /* ... */ });
-// conn_sexp is CLOSED here. Open it:
-unsafe { open_connection(conn_sexp)? };
+// conn_sexp is already OPEN. readLines / writeBin / seek all work.
 ```
 
-`open_connection` is a helper that calls the connection's registered `open` trampoline through R's internal dispatch. See `miniextendr-api/src/connection.rs` for the implementation.
+If your `RConnectionImpl::open` returns `false`, the boxed state is dropped via the destroy trampoline and `build` returns `SEXP::nil()`. Make sure `open` returns `true` on success.
 
 ### `RConnectionImpl` trait
 
@@ -116,7 +118,7 @@ Panics inside `RConnectionImpl` methods are caught by `catch_connection_panic`. 
 
 ## Common pitfalls
 
-- **Connection starts CLOSED**: `R_new_custom_connection` creates in closed state. Always call the open callback explicitly after `build()` if you want an already-open connection. Returning a closed connection to R code that calls `readLines()` on it immediately will produce an "connection is not open" error.
+- **Underlying API starts CLOSED with `text = TRUE`**: `R_new_custom_connection` returns a closed connection with `text = TRUE` regardless of mode. `RCustomConnection::build` opens it for you and infers `text` from the mode string before returning, so this is invisible to callers — but if you ever bypass the builder and call `R_new_custom_connection` directly, you need to handle both yourself.
 
 - **Panics are silently absorbed**: unlike `#[miniextendr]` functions (which use the tagged-SEXP error transport to surface Rust panics as R errors), connection trampolines use the fallback-value pattern. A panic in `read()` returns `0` bytes — the R caller sees a short read, not an error. Design your `RConnectionImpl` to return error signals through return values, not panics.
 
--- a/miniextendr-api/CLAUDE.md
+++ b/miniextendr-api/CLAUDE.md
@@ -27,7 +27,7 @@ Runtime crate — FFI, ExternalPtr, ALTREP, worker thread, error/condition trans
 - **PROTECT discipline against R-devel GC** — `SEXP::scalar_string` and `scalar_logical(true)` allocate fresh STRSXP/LGLSXP; protect across `SET_VECTOR_ELT`/`SETATTRIB`. R-devel's GC is more aggressive — `make_rust_condition_value` crossed the threshold in PR #344 commit `af6b4875` (recursive gc + segfault at small offsets). R-release passing ≠ safe.
 - **`#[macro_export]` collides with same-named modules at crate root** — `pub mod error`/`pub mod condition` shadow `error!`/`condition!` macros under `use miniextendr_api::{error, condition}`. Invoke via fully-qualified path; `message!` / `warning!` have no module conflict.
 - **`R_GetCCallable` throws on miss** — does NOT return NULL. Force DLL load via `NAMESPACE importFrom(...)`.
-- **`R_new_custom_connection` creates connections CLOSED** (`isopen = FALSE`); explicitly call the `open` callback after building.
+- **`R_new_custom_connection` creates connections CLOSED with `text = TRUE`** — `RCustomConnection::build` papers over both (infers `text` from mode, invokes the `open` trampoline before returning). If you bypass the builder, you must handle both yourself or `writeBin`/`writeLines`/`seek` will fail and binary modes will reject `readBin`/`writeBin`.
 - **Pointer provenance for ExternalPtr** — cache `*mut T` via `&mut T` / `downcast_mut` / `Box::into_raw`. Never write through a `*mut` derived from `&T` / `downcast_ref` — UB under Stacked Borrows.
 
 ## Features
--- a/miniextendr-api/src/connection.rs
+++ b/miniextendr-api/src/connection.rs
@@ -611,12 +611,22 @@ pub trait RConnectionImpl: Sized + 'static {
         0
     }
 
-    /// Formatted print (vfprintf-style).
+    /// Formatted print (`vfprintf`-style).
     ///
-    /// This is rarely needed - R typically uses `write` for output.
-    /// Return the number of characters written, or -1 on error.
+    /// R calls into this through `Rconn_printf` from `writeLines`, `cat`,
+    /// sink dispatch, and similar. The default implementation returns `-1`
+    /// to signal "delegate to the standard format-and-write path," in
+    /// which case the framework formats the args into a fixed buffer via
+    /// the host `vsnprintf` and forwards the bytes to [`write`](Self::write).
     ///
-    /// The default returns -1 (not implemented).
+    /// Override this only if you need a custom output path that bypasses
+    /// [`write`](Self::write) (e.g., to stream into a pre-encoded sink).
+    /// Return the number of bytes written on success, or `-1` on error.
+    ///
+    /// # Safety
+    ///
+    /// `fmt` is a null-terminated C string from R. `ap` is the `va_list`
+    /// of R's printf-style call; treat it as opaque and pass to libc.
     fn vfprintf(&mut self, _fmt: *const c_char, _ap: *mut c_void) -> i32 {
         -1
     }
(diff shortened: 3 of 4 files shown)


CGMossa (commented):
**Verdict: LGTM**

The fix matches the planner's diagnosis and the two beyond-plan corrections are both grounded in R's source. Verified against `wch/r-source/trunk/src/main/connections.c`:

- `init_con` defaults: `blocking = FALSE`, `text = TRUE`, `isopen = FALSE` (line 713). PR's `blocking = true` default + text inference from mode + post-open call are all correct papering-over.
- `do_readLines` at lines 4511-4523 is exactly the `text && !blocking` silent-pushback path the PR comment describes; flipping `blocking = true` triggers the `incomplete final line found` warning branch instead, which matches the 7 expected warnings.
- `dummy_vfprintf` at lines 471-513 uses the same `BUFSIZE = 10000` stack buffer the PR mirrors. R's version additionally falls back to `vasprintf`/`R_alloc(10*BUFSIZE)` on truncation; the PR doesn't, but documents the omission and `Rconn_printf` on a custom connection is rare. Acceptable.

CLOSED-state fix: `open_trampoline::<T>(conn)` is sequenced after all callback wiring + `(*conn).private` assignment and before the unprotected `sexp` return. On failure, `destroy_trampoline` runs, which drops the boxed state and nulls `private`. The dispatch through `R_RegisterCFinalizerEx`'s `ex_ptr` keeps the connection table slot alive, so returning `SEXP::nil()` on failure is safe.

`vfprintf_trampoline`: gives user override first crack (returns >= 0 wins), else `vsnprintf` into `[u8; 10_000]` + forward to `state.write`. `catch_connection_panic` wraps the whole thing. `vsnprintf` is libc — can't panic. Stack budget at 10 KiB is fine on R's 16 MB worker stack.

`Cargo.lock` / `rv.lock` / vendor tarball not in the PR — 4 files only, all expected.

Test integrity: all 13 custom-conn assertions use `expect_equal` / `expect_length` on actual data. Warnings don't mask anything.

---

**Nits (informational, none blocking):**

1. `va_list` aliasing footgun in `vfprintf_trampoline` (`connection.rs:1075-1090`). The trait's default `vfprintf` returns `-1` without touching `ap`, so the framework's subsequent `vsnprintf(buf, BUFSIZE, fmt, ap)` is safe. But a user override that *does* consume `ap` via `vsnprintf`/`vfprintf` and then returns `-1` will leave `ap` in an indeterminate state, and the framework will read it again — UB per C standard. R's own `dummy_vfprintf` uses `va_copy(aq, ap)` for exactly this reason. Worth a sentence in the `RConnectionImpl::vfprintf` doc: "if your override returns -1 the framework will re-read `ap`; do not consume `ap` unless you return >= 0."

2. The trampoline doc-comment says "we still write the truncated portion and return the truncated length" but the code returns `res` (the full *desired* length from `vsnprintf`, which can exceed `BUFSIZE`). Trivial doc/code mismatch — `res` is the right thing to return because `Rconn_printf` callers treat it as "would have written this many bytes if buffer were larger." Minor wording fix.

3. `R_new_custom_connection` (line 7415-7462) does NOT call `R_PreserveObject(ans)`; the returned SEXP is unprotected after the internal `UNPROTECT(3)`. The new `open_trampoline(conn)` call runs *before* returning `sexp` — if a user `open()` impl ever allocates via the R API and triggers GC, `sexp` could be reclaimed. The `Connections[ncon]` table holds the underlying `Rconn*`, and the `ex_ptr` finalizer is registered, so the integer-scalar `ans` is recoverable via the table, but it's a subtle pre-existing pattern not introduced by this PR. Not a blocker; flagging only because the planner asked.

4. Branch name `fix/custom-connection-tests-v2` — noted, fine, just unusual.

5. `MEMORY.md` already documents the `R_new_custom_connection` CLOSED-state gotcha; consider porting the `blocking = FALSE` and `vfprintf = null_vfprintf` defaults into the same bullet so future reviewers don't have to rediscover from R source.

PR #684, opened 2026-05-21
feat(gc_protect): generic Protected<'a, T> primitive (#681)

### LINKED ISSUE #681: Generic Protected<'a, T> primitive — extract the OwnedProtect + Rust-view pattern that ProtectedStrVec already uses
## Motivation

Several places in `miniextendr-api` bundle a Rust-side view of an
SEXP with an `OwnedProtect` guard so the caller can hold the view
without worrying about R's GC reclaiming the underlying SEXP. The
established example is `ProtectedStrVec`:

```rust
// strvec.rs
pub struct ProtectedStrVec {
    inner: StrVec,
    len: isize,
    _protect: Option<OwnedProtect>,
}
```

`Root<'a>` does roughly the same thing for the `ProtectScope`-based
batched flavour, but it's specifically a `SEXP` wrapper — it can't
hold a Rust view (a `Vec<T>`, a `HashMap`, etc.) borrowing into the
protected SEXP.

That leaves an awkward gap: every time we want a new "safe-to-hold
borrowed view of SEXP-backed data" type, we re-roll
`{ view, _protect: OwnedProtect, _marker: PhantomData<&'a ()> }` by
hand. #671's proposed `BorrowedRows<'a, T>` (RAII variant of the
`dataframe_to_vec` follow-up) is the next case in line and would
re-roll it again.

## Proposed primitive

Extract the shared shape:

```rust
/// A Rust value (`T`) bundled with an `OwnedProtect` guard on an SEXP
/// it borrows from. The protect releases on drop; the lifetime ties
/// any borrows inside `T` to `&self`, so `T`'s SEXP-internal
/// references can't outlive the protection.
pub struct Protected<'a, T> {
    inner: T,
    _protect: OwnedProtect,
    _marker: core::marker::PhantomData<&'a ()>,
}

impl<'a, T> Protected<'a, T> {
    /// # Safety
    ///
    /// `sexp_to_protect` must be a valid SEXP. `inner` may borrow
    /// from it; the lifetime `'a` is tied to `&self` thereafter.
    /// Caller is on the R main thread (see `OwnedProtect::new`).
    #[inline]
    pub unsafe fn new(sexp_to_protect: SEXP, inner: T) -> Self {
        Self {
            inner,
            _protect: unsafe { OwnedProtect::new(sexp_to_protect) },
            _marker: core::marker::PhantomData,
        }
    }

    /// Borrow the inner view.
    #[inline]
    pub fn get(&self) -> &T { &self.inner }

    /// Consume the protect *and* the inner view together. The SEXP
    /// becomes unprotected as soon as this returns, so any owned
    /// data extracted from `T` must not retain SEXP references.
    #[inline]
    pub fn into_inner(self) -> T { self.inner }
}

impl<'a, T> core::ops::Deref for Protected<'a, T> {
    type Target = T;
    fn deref(&self) -> &T { &self.inner }
}
```

(Variants with `ProtectScope` and `Root` are possible — a
`ProtectedScoped<'scope, T>` flavour for the batched case — but
likely not needed initially.)

## Sites that collapse onto this

- `ProtectedStrVec` (today): `Protected<'static, StrVec>`-ish.
  (The current type holds the protect via `Option<OwnedProtect>` to
  handle the `from_sexp_trusted` case where the caller already owns
  the protection. The generic primitive would need an
  `unprotected` constructor variant, or the strvec keeps a thin
  wrapper for the trusted path.)
- `BorrowedRows<'a, T>` from #671 (proposed):
  `Protected<'a, Vec<T>>`.
- Future `ProtectedNumVec`, `ProtectedListView`, etc. — same shape.

## What this *doesn't* replace

- `OwnedProtect` itself — it's the underlying primitive
  `Protected<'a, T>` is built on, still useful for the
  "raw SEXP, no Rust view" case.
- `ProtectScope` / `Root<'a>` — different lifecycle (batched
  multi-value protection within a function body). Lives alongside.

## Naming bikeshed

`Protected<'a, T>`, `Guarded<'a, T>`, `Rooted<'a, T>`, `Held<'a, T>`.
`Protected` matches the existing `ProtectedStrVec` naming.

## Related

- #671 — the borrowed-row RAII follow-up that motivated this
  extraction. Once `Protected<'a, T>` exists, the `BorrowedRows`
  sketch in that issue collapses to a type alias.

CGMossa:
Lands the generic `Protected<'a, T>` primitive from #681. Future borrowed-view types (#671b `BorrowedRows`, etc.) collapse onto this shape instead of re-rolling the `OwnedProtect` + view bundle by hand. `ProtectedStrVec` migrates to wrap it as the API-sufficiency check.

<details>
<summary><h3>🤖 Implementation notes & test plan (AI-written)</h3></summary>

## Summary

- **New type `Protected<'a, T>`** added to `miniextendr-api/src/gc_protect.rs` in a new `// region: Protected` block between `OwnedProtect` (line ~872) and `ReprotectSlot`. Holds `inner: T`, `_protect: Option<OwnedProtect>`, and `_marker: PhantomData<&'a ()>`.
- **Two constructors**: `Protected::new(sexp, inner)` calls `Rf_protect` and stores a `Some(OwnedProtect)`; `Protected::from_trusted(_sexp, inner)` stores `None` (no protect push) for SEXPs already protected by R.
- **Three inherent methods**: `get() -> &T`, `into_inner() -> T`, plus `Deref<Target = T>` and `Debug` (when `T: Debug`).
- **`ProtectedStrVec` migrated** in `miniextendr-api/src/strvec.rs`: old `{ inner: StrVec, len: isize, _protect: Option<OwnedProtect> }` becomes `{ protected: Protected<'static, StrVec>, len: isize }`. All public methods now delegate via `self.protected.get().<method>()`. No behavioural change visible to callers.
- **Exports**: `Protected` added to `lib.rs` `pub use gc_protect::{ ... }` block and to `prelude.rs` Types region alongside `OwnedProtect` and `ProtectedStrVec`.
- **5 unit tests** added to the `#[cfg(test)]` block in `gc_protect.rs`: `protected_get_returns_inner`, `protected_deref_reaches_inner`, `protected_into_inner_moves_value`, `protected_from_trusted_does_not_hold_protect` (verifies `Debug` shows `protected: false`), and `protected_is_copy_of_struct_fields_compile_check`.

## Test plan

- [x] `just fmt` — clean
- [x] `just check` — clean across all workspace members (root, rpkg, cross-package)
- [x] `just clippy` — clean
- [x] `cargo test -p miniextendr-api strvec` — 1 passed, 0 failures
- [x] `cargo test -p miniextendr-api gc_protect` — 18 passed (14 pre-existing + 5 new tests), 0 failures
- [x] CI `clippy_all` feature sweep (full feature list from `.github/workflows/ci.yml`) — clean, `-D warnings`
- [x] `R CMD INSTALL rpkg` (via `R_LIBS_USER` pointing to main repo rv library) — DONE, installed successfully
- [x] `devtools::test(pkg = "rpkg")` — `[ FAIL 0 | WARN 7 | SKIP 19 | PASS 5967 ]` (warnings and skips are pre-existing)

## Notes

- The `'static` lifetime on `Protected<'static, StrVec>` in `ProtectedStrVec` is correct: `StrVec` itself carries no lifetime parameter, so the `Protected` lifetime says "no external borrows". The `StrVec` methods handle SEXP-internal borrows via lifetime elision on `&self`.
- The `from_trusted` constructor keeps `_sexp` in the signature (unused in the body) as a documentation pattern: callers can see which SEXP is being trusted.
- The 3 collateral changes (`rpkg/R/miniextendr-wrappers.R` line-number drift, `rpkg/src/rust/Cargo.lock` git-source removal after `--config patch`, `rpkg/src/rust/condition_sidecar_tests.rs` formatting) were left unstaged: they are pre-existing drift unrelated to this PR and require a NAMESPACE re-document pass to satisfy the pre-commit hook cleanly. They can be squashed in at review time if desired.
- `OwnedProtect` import was retained in `strvec.rs` because `StrVec::set_charsxp` and `StrVec::set_str` still use it directly.

---
_Drafted by Claude (claude-sonnet-4-6). Reviewed by the author._

Closes #681.

</details>

### DIFF
--- a/miniextendr-api/src/gc_protect.rs
+++ b/miniextendr-api/src/gc_protect.rs
@@ -1595,20 +1595,16 @@ mod tests {
         let p = unsafe {
             Protected::<'static, Vec<u8>>::from_trusted(crate::ffi::SEXP(core::ptr::null_mut()), v)
         };
-        // _protect is private — verify indirectly: if it were Some(OwnedProtect),
-        // the OwnedProtect drop would call Rf_unprotect(1) without a matching
-        // protect (crashing R). Since we're not running R here, just confirm the
-        // Debug output shows `protected: false`.
+        // This test uses `from_trusted` to avoid touching the protect stack;
+        // `Protected::new` would require an initialized R runtime (see
+        // `tests/gc_protect.rs` for that path). Verify the `_protect` field is
+        // `None` indirectly via the `Debug` output.
         assert!(format!("{p:?}").contains("protected: false"));
     }
 
-    /// Verify `Protected<'a, T>` with `Some(OwnedProtect)` is `!Send`
-    /// (compile-time — the type-system test is implicit: if this test builds
-    /// we confirm the field layout compiles cleanly).
+    /// Smoke-test: `Protected::from_trusted` + `Deref` compiles for a scalar `T`.
     #[test]
-    fn protected_is_copy_of_struct_fields_compile_check() {
-        // Smoke-test: constructing Protected<'static, i32> via from_trusted
-        // (the None path, which is auto-Send) must compile.
+    fn protected_from_trusted_compile_check() {
         let p = unsafe {
             Protected::<'static, i32>::from_trusted(crate::ffi::SEXP(core::ptr::null_mut()), 99)
         };
--- a/miniextendr-api/src/gc_protect.rs
+++ b/miniextendr-api/src/gc_protect.rs
@@ -931,6 +931,128 @@ impl std::ops::Deref for OwnedProtect {
 }
 // endregion
 
+// region: Protected
+
+/// A Rust value (`T`) bundled with an [`OwnedProtect`] guard on an SEXP
+/// the value borrows from. The protect releases on drop; the lifetime
+/// ties any borrows inside `T` to `&self`, so `T`'s SEXP-internal
+/// references can't outlive the protection.
+///
+/// # When to use `Protected<'a, T>` vs the alternatives
+///
+/// | Pattern | Use | Why |
+/// |---------|-----|-----|
+/// | [`OwnedProtect`] | raw SEXP, no Rust view | one-shot protect/unprotect on a single SEXP |
+/// | [`ProtectScope`] + [`Root`] | several SEXPs in one function body | batched UNPROTECT, no Rust view |
+/// | `Protected<'a, T>` | SEXP + Rust view of its data | hand the bundle to callers; borrows in `T` tied to `&self` |
+///
+/// # Notes on Send/Sync
+///
+/// When constructed via [`Protected::new`], the inner [`OwnedProtect`] carries
+/// `!Send + !Sync` (via `NoSendSync`). When constructed via
+/// [`Protected::from_trusted`], the `_protect` field is `None` and the type
+/// becomes auto-`Send`/`Sync` — the same behaviour as [`ProtectedStrVec`] today.
+pub struct Protected<'a, T> {
+    inner: T,
+    _protect: Option<OwnedProtect>,
+    _marker: core::marker::PhantomData<&'a ()>,
+}
+
+impl<'a, T> Protected<'a, T> {
+    /// Create a protected bundle. Calls `Rf_protect` on `sexp`.
+    ///
+    /// `inner` may borrow from `sexp`; the lifetime `'a` is tied to
+    /// `&self` thereafter, so any borrow inside `inner` cannot outlive
+    /// this `Protected`.
+    ///
+    /// # Safety
+    ///
+    /// - Must be called from the R main thread.
+    /// - `sexp` must be a valid SEXP.
+    /// - If `inner` borrows from `sexp`, its lifetime parameter must
+    ///   match `'a`.
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// use miniextendr_api::{Protected, OwnedProtect};
+    /// use miniextendr_api::ffi::SEXP;
+    ///
+    /// unsafe fn wrap_view(sexp: SEXP, view: MyView<'_>) -> Protected<'_, MyView<'_>> {
+    ///     // Protect the SEXP and bundle it with the view.
+    ///     // The view's borrow is tied to the returned Protected.
+    ///     Protected::new(sexp, view)
+    /// }
+    /// ```
+    #[inline]
+    pub unsafe fn new(sexp: SEXP, inner: T) -> Self {
+        let guard = unsafe { OwnedProtect::new(sexp) };
+        Self {
+            inner,
+            _protect: Some(guard),
+            _marker: core::marker::PhantomData,
+        }
+    }
+
+    /// Create a protected bundle without adding to the protect stack.
+    ///
+    /// Use when `sexp` is already protected by R (a `.Call` argument,
+    /// a [`ProtectScope`] slot, an enclosing [`OwnedProtect`]) to avoid
+    /// double-protecting. The lifetime contract is unchanged — `'a`
+    /// still ties any borrows inside `inner` to `&self`.
+    ///
+    /// # Safety
+    ///
+    /// - Must be called from the R main thread.
+    /// - `sexp` must be a valid SEXP.
+    /// - `sexp` must remain GC-protected for the lifetime of the
+    ///   returned `Protected`.
+    /// - Lifetime contract same as [`Protected::new`].
+    #[inline]
+    pub unsafe fn from_trusted(_sexp: SEXP, inner: T) -> Self {
+        Self {
+            inner,
+            _protect: None,
+            _marker: core::marker::PhantomData,
+        }
+    }
+
+    /// Borrow the inner view.
+    #[inline]
+    pub fn get(&self) -> &T {
+        &self.inner
+    }
+
+    /// Consume the bundle and return the inner view.
+    ///
+    /// The [`OwnedProtect`] guard drops here, releasing the SEXP from the protect
+    /// stack. Any owned data extracted from `T` must not retain SEXP references
+    /// after this point.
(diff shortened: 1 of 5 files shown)


CGMossa (commented):
Minor issues found. The core implementation matches the plan; flagging one gap and a couple of nits before undraft.

## Missing: R-runtime stack-balance integration tests

The plan called for two runtime tests:

1. **Drop unprotects** (`new` path) — `ProtectScope::count()` nets zero before/after.
2. **`from_trusted` does not protect** — same check, nets zero before/after.

Neither can be expressed as a unit test (no embedded R), and the repo already documents this: `miniextendr-api/src/gc_protect.rs:1544-1552` explicitly defers balance tests to integration tests in `miniextendr-api/tests/gc_protect.rs`, and that file already has analogous coverage for `OwnedProtect` (`owned_protect_basic`, `owned_protect_deref`), `ProtectScope` (`protect_scope_balance_basic/empty/large`, `nested_scopes_independent`, `deeply_nested_scopes`), and `ReprotectSlot` (`reprotect_slot_count_stays_one`, `reprotect_slot_value_updates`).

The added unit test `protected_from_trusted_does_not_hold_protect` only asserts on the `Debug` output string (`"protected: false"`) — that's a proxy for the field state, not a check that nothing was pushed onto R's PROTECT stack. The plan explicitly flagged this risk:

> The Debug-string assertion ("`protected: false`") is a proxy, not a runtime check — flag if it's the only verification.

**Action:** add a `// region: Protected tests` block to `miniextendr-api/tests/gc_protect.rs` mirroring the `owned_protect_basic` pattern:

```rust
#[test]
fn protected_new_drops_unprotect() {
    r_test_utils::with_r_thread(|| unsafe {
        let scope = ProtectScope::new();
        let outer = scope.protect(SEXP::scalar_integer(1));
        let depth_before = scope.count();
        {
            let p = Protected::<'static, miniextendr_api::ffi::SEXP>::new(
                outer.get(),
                outer.get(),
            );
            // While `p` is alive, an additional protect entry exists.
            // The drop at the end of this block UNPROTECTs(1).
            let _ = p.get();
        }
        // Net change after p drops is zero.
        assert_eq!(scope.count(), depth_before);
    });
}

#[test]
fn protected_from_trusted_no_stack_change() {
    r_test_utils::with_r_thread(|| unsafe {
        let scope = ProtectScope::new();
        let outer = scope.protect(SEXP::scalar_integer(7));
        let depth_before = scope.count();
        {
            let p = Protected::<'static, miniextendr_api::ffi::SEXP>::from_trusted(
                outer.get(),
                outer.get(),
            );
            let _ = p.get();
        }
        assert_eq!(scope.count(), depth_before);
    });
}
```

This is the only blocking gap from the plan.

## Misleading test name

`protected_is_copy_of_struct_fields_compile_check` (`gc_protect.rs:1609`) doesn't check anything about `Copy` — `Protected` cannot be `Copy` (it can hold `OwnedProtect`, which isn't `Copy`). The body is a smoke test that `from_trusted` + `Deref` compiles for `Protected<'static, i32>`. Rename to `protected_from_trusted_compile_check` (or fold into one of the other unit tests — there's already substantial overlap with `protected_get_returns_inner` and `protected_deref_reaches_inner`).

## Cosmetic

- `gc_protect.rs:1600` — the comment "if it were `Some(OwnedProtect)`, the `OwnedProtect` drop would call `Rf_unprotect(1)` without a matching protect (crashing R)" is true but misleading in unit-test context (R isn't initialized; the call would crash on `Rf_unprotect` itself, not on the protect imbalance). The Debug-string assertion is what's actually verified; the rationale block can be tightened or pointed at the new integration test once it exists.
- PR body's AI-attribution footer says "Reviewed by the author." inside the `<details>` block — fine, matches the skill convention.

## What's right

- `// region: Protected` block correctly placed between `OwnedProtect` and `ReprotectSlot` (`gc_protect.rs:934`).
- Field shape exactly matches plan: `inner: T`, `_protect: Option<OwnedProtect>`, `_marker: PhantomData<&'a ()>`.
- `from_trusted(_sexp, inner)` keeps `_sexp` in the signature with leading underscore as documentation, per plan.
- `Protected::new` rustdoc carries `# Safety` (R main thread, valid SEXP, lifetime contract) and `# Example` (≤12 lines, `ignore`d, matches house-style placeholder pattern from `ProtectScope::protect_with_index` / `alloc_vector`).
- `Protected::from_trusted` rustdoc adds the additional "`sexp` must remain GC-protected for the lifetime of the returned `Protected`" clause, per plan.
- `Deref<Target = T>` and `Debug` (conditional on `T: Debug`) impls match plan exactly.
- `ProtectedStrVec` migration is clean: `inner: StrVec` and `_protect: Option<OwnedProtect>` both removed; only `protected: Protected<'static, StrVec>` + cached `len: isize` remain. All public methods delegate via `self.protected.get()`. No layering, no shim.
- Exports: `lib.rs:608` adds `Protected` to the `pub use gc_protect::{…}` block; `prelude.rs:64` adds it to the Types region. Both alphabetically slotted next to `OwnedProtect` / `ProtectedStrVec`.
- No `mod.rs`, no `// =====` section markers, no defensive panics, no `#[allow(...)]` added.
- No out-of-scope creep — no `BorrowedRows`, no `ProtectedScoped`, no other types migrated.
- No unrelated collateral in the PR diff (the 3 locally-dirty files `rpkg/R/miniextendr-wrappers.R`, `rpkg/src/rust/Cargo.lock`, `rpkg/src/rust/condition_sidecar_tests.rs` are not in the diff, per `gh pr view --json files`).
- `Rust lint` (clippy_default + clippy_all), `wasm32 check`, `Generated Files Check`, `Bootstrap Vendor Test` already green on CI.

Address the missing integration tests and the test rename, then this is good to undraft.

PR #710, opened 2026-05-25
refactor: cargo feature hygiene (rename default-* + mutex compile_error)

CGMossa:
Renames the five `default-*` cargo features to `*-default` for clarity (the
`default-*` prefix wrongly suggests the feature is in Cargo's `default` set)
and updates the existing r6/s7 mutex `compile_error!` with the more helpful
wording proposed in #676. No backwards-compat shims — project is unreleased.

<details>
<summary>AI-written details</summary>

## What changed

**Five renames** (codegen-default selector pattern):

| Before | After |
|---|---|
| `default-worker` | `worker-default` |
| `default-r6` | `r6-default` |
| `default-s7` | `s7-default` |
| `default-strict` | `strict-default` |
| `default-coerce` | `coerce-default` |

**Mutex enforcement**: the existing `compile_error!` in
`miniextendr-macros/src/lib.rs` was already covering `r6-default` ⊕ `s7-default`
(formerly `default-r6` ⊕ `default-s7`). Its message has been updated to match
the wording proposed in #676, explaining the fallback rather than just stating
incompatibility.

**Note on `strict` ⊕ `coerce`**: the issue body of #676 speculated they may be
mutually exclusive. They are not — `docs/FEATURE_DEFAULTS.md` explicitly
demonstrates `#[miniextendr(strict, coerce)]` is the valid baseline behavior
when both features are enabled. They're orthogonal concepts (strict = error
on lossy conversions; coerce = autocast where compatible), so no second
`compile_error!` is added.

## Sites touched

- `miniextendr-api/Cargo.toml`, `miniextendr-macros/Cargo.toml`, `rpkg/src/rust/Cargo.toml`
- `miniextendr-macros/src/{lib.rs, miniextendr_fn.rs, miniextendr_impl.rs, miniextendr_impl_trait.rs, miniextendr_impl_trait/vtable.rs, c_wrapper_builder.rs}` — all `cfg!(feature = "default-*")` and `#[cfg(feature = "default-*")]` sites
- `.github/workflows/ci.yml` — `clippy_all` feature list + note
- `README.md`, `AGENTS.md`, `CLAUDE.md`, `miniextendr-api/{README.md,CLAUDE.md}`, `miniextendr-api/src/lib.rs` (rustdoc)
- `docs/{FEATURE_DEFAULTS,FEATURES,MINIEXTENDR_ATTRIBUTE,STRICT_MODE,BENCHMARKS}.md`
[...]

PR #735, opened 2026-05-26
ci(vendor): pin self-references to HEAD in just vendor

CGMossa:
## Summary

Unblocks cross-crate cargo-surface edits (the case PR #710 hit). `just vendor` previously stripped the path-based `[patch]` override and fell back to resolving `miniextendr-{api,macros,lint}` from origin's default branch (main). Any PR that edits a feature on `miniextendr-api` alongside the rpkg consumer that requests it can't pass CI's vendor step until main has the change — chicken-and-egg.

This PR injects `rev = "<HEAD>"` into the three self-referencing dep lines in `rpkg/src/rust/Cargo.toml` for the duration of `cargo generate-lockfile`, then restores via `trap`. Cargo's `[patch]` rejects same-URL overrides, so direct dep modification is the simplest lever. Source replacement at install time matches by bare URL — verified `[source."git+url"]` correctly resolves `git+url?rev=<sha>` lookups — so `configure.ac` needs no changes.

## Test plan

- [ ] CI green on this PR (no behavior change on main — pin produces same lockfile content)
- [ ] Rebase #710 on top after merge, confirm Bootstrap Vendor Test + CRAN-like check + R CMD check pass

### DIFF
--- a/justfile
+++ b/justfile
@@ -446,11 +446,44 @@ vendor:
     set -euo pipefail
     rust_dir="{{justfile_directory()}}/rpkg/src/rust"
     cargo_cfg="$rust_dir/.cargo/config.toml"
+    rpkg_toml="$rust_dir/Cargo.toml"
+    # Self-reference pin: cross-crate edits that touch the cargo surface of
+    # miniextendr-{api,macros,lint} (e.g. renaming a feature on the api crate
+    # alongside the rpkg consumer that requests it) only resolve consistently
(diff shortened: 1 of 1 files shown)


CGMossa:
Withdrawing per discussion — sed-on-tracked-Cargo.toml is the wrong layer. Trying a cleaner approach (path-based self-refs with tarball-mode fallback) in a worktree.

Issue #491, opened 2026-05-10
webR CI tier 2: Docker `R CMD INSTALL` workflow


CGMossa:
Tier 2 of the webR CI plan (formerly `plans/webr-ci.md`). Tier 1 (`cargo check --target wasm32-unknown-emscripten`) landed in #480; tier 2 extends it to a real install pass that catches link errors and Makevars regressions.

## What

Add a CI job that runs inside `ghcr.io/r-wasm/webr@<digest>` (matching `Dockerfile.webr`) and does:

```bash
export R_MAKEVARS_USER=/opt/webr/packages/webr-vars.mk
bash ./rpkg/configure
R_LIBS=/tmp/wasm-lib /opt/webr/host/R-4.5.1/bin/R CMD INSTALL \
  --no-docs --no-test-load --no-staged-install \
  --library=/tmp/wasm-lib \
  rpkg
```

Catches: link errors, missing `wasm_registry.rs`, Makevars wasm-mode bugs, the actual rwasm/emcc dance.

Cost: ~10–15 min/PR.

## Dependencies

- #482 (gate `link_to_r()` on `target_arch != "wasm32"`) — without it, the rpkg pass needs a dummy R_HOME shim or a real R install in the runner.

## Open questions to resolve during implementation

- Does the base webR image's host R actually support `R CMD INSTALL` against the wasm lib tree without `pak`/`rwasm` wrappers? Verify locally via `just docker-webr-shell` before wiring.
- `R_MAKEVARS_USER=webr-vars.mk` is required — without it the install builds with the host gcc and produces a native shared object instead of a wasm side-module.

## Refs

- Umbrella: #470
- Local validation: `tests/webr-smoke.sh` (the local smoke script driven by `just docker-webr-smoke`) is the prior-art template for the install commands.

Issue #494, opened 2026-05-10
webR: verify wasm side-module RUSTFLAGS against rwasm


CGMossa:
`rpkg/src/Makevars.in` does not currently set RUSTFLAGS for the wasm32 install path. The proposed set (per `plans/webr-support.md`) is:

```
-C relocation-model=pic -C link-args=-s SIDE_MODULE=1
```

These have NOT been empirically validated against webR's pinned Emscripten or `rwasm`'s actual flag set. The smoke script (`tests/webr-smoke.sh` / `just docker-webr-smoke`) is the validator: if the wasm side-module fails to link cleanly or to load via `library(miniextendr)` in webR, this is the first thing to verify.

## What

1. Run `just docker-webr-smoke` end-to-end. Observe whether Phase 2 (`R CMD INSTALL` with `webr-vars.mk`) produces a loadable side-module.
2. If it doesn't link, inspect `rwasm`'s `cargo build` flags (the canonical reference) and add the required `RUSTFLAGS` to `rpkg/src/Makevars.in`'s wasm path.
3. Update `docs/WEBR.md` once the empirically-correct set is locked in.

## Refs

- Umbrella: #470
- Empirical validator: `tests/webr-smoke.sh` (#488)

CGMossa:
Empirically validated the wasm side-module flags via the astra downstream smoke — and the proposed `-s SIDE_MODULE=1` doesn't work for our Rust-staticlib packages. Documenting the finding and two fixes here since #494 called this out as the thing to verify.

<details>
<summary><h4>AI-written details</h4></summary>

## Summary
- The proposed wasm link flag `-s SIDE_MODULE=1` fails for any miniextendr-style package, because the Rust staticlib drags thousands of mangled stdlib/dep symbols into the side-module. With `SIDE_MODULE=1` (export every default-visibility symbol) the EXPORT section balloons to ~3000+ entries / ~330 KB.
- webR's JS-side `dyn.load` then fails. On the Emscripten pinned in the current `miniextendr-webr-dev` image it surfaces as `TypeError: Cannot read properties of undefined (reading 'value')` during `library(<pkg>)`. On newer Emscripten (4.0.8+) the same root cause is a hard `emcc: error: invalid export name: _ZN40_$LT$str$...` (emcc now rejects `$` in export names).
- Two fixes, both confirmed to shrink the EXPORT table to what webR can load:
[...]

PR #750, opened 2026-05-27
ci(webr): tier-3 — drive library(miniextendr) in webR Node session (#492)

### LINKED ISSUE #492: webR CI tier 3: Node smoke test in CI
Tier 3 of the webR CI plan. Builds on tier 2 (#491) — actually loads the wasm side-module in a webR Node session and runs a minimal smoke test (or the testthat suite, with tolerance for known-failures).

## What

After tier 2 produces a wasm-installed package at `/tmp/wasm-lib/miniextendr/`, drive a Node script that:

1. Imports `@r-wasm/webr` from `/opt/webr/src` (linked locally, no network)
2. NODEFS-mounts the wasm R lib tree
3. `library(miniextendr)`
4. Runs a minimal smoke (e.g. one exported `#[miniextendr]` function returns expected output) — OR `testthat::test_local()` with non-zero exit only on infrastructure failures

Cost: tier 2 + ~5 min for Node + webR startup + test execution.

## Dependencies

- Tier 2 (#491) must land first.

## Prior art

`tests/webr-smoke.sh` already does this end-to-end locally. The CI job is essentially a stripped-down version that runs against a runner without the docker indirection.

## Refs

- Umbrella: #470

CGMossa:
Closes #492. Builds on tier 2 (#491, landed via #744) — adds a Phase 3 step set to the existing `webr-install` job that boots a webR Node session against the wasm install Phase 2 just produced and drives `library(miniextendr)`. The emcc link succeeding on its own isn't proof the side-module is actually loadable in webR, so this is the first real end-to-end runtime gate.

Runner code lives in `tests/webr-node-smoke/` (package.json + smoke.mjs) so the YAML stays tight and the runner is editable as a normal file. Mirrors the Phase 3 structure of `tests/webr-smoke.sh` without the docker indirection.

<details>

## Summary
- `tests/webr-node-smoke/package.json` — pins `webr` via `file:///opt/webr/src` (local link on the webR base image, no registry traffic).
- `tests/webr-node-smoke/smoke.mjs` — boots WebR, NODEFS-mounts `/tmp/wasm-lib`, pre-installs miniextendr's hard Imports (`cli`/`lifecycle`/`R6`/`S7`/`vctrs`) from `repo.r-wasm.org`, then runs `library(miniextendr)` and a minimal `packageVersion()` post-load sanity call.
- `.github/workflows/webr.yml`:
  - two new steps after "Verify wasm side-module landed": `npm install --prefer-offline` then `timeout 900 node smoke.mjs`.
  - header comment block updated to reflect tier-2 + tier-3 in one job.
  - `paths:` filter gains `tests/webr-node-smoke/**` so PRs that only touch the runner still trigger CI.

## Test plan
- [ ] tier-2 phases (Phase 1 + Phase 2 + verifier) still pass — no behaviour change in those steps.
- [ ] Tier 3 — `npm install` step links `/opt/webr/src` successfully.
- [ ] Tier 3 — `node smoke.mjs` reports `[tier3] OK: library(miniextendr) loaded.` and `[tier3] Tier-3 PASSED.`.

## Notes
- Single-job design: all three phases run in `webr-install` because they share the container, the filesystem (`/tmp/wasm-lib/miniextendr`), and the webR base image's `/opt/webr/*` paths. Splitting would force an artifact handoff with no improvement to the green/red signal that named steps already give us.
- 15-minute `timeout` wrap on the smoke runner so a wedged Node session fails visibly instead of consuming the job's 90-minute cap.
- Imports are fetched from `repo.r-wasm.org` per-run rather than baked into the image — the webR mirror has stable URLs and this keeps the dependency surface explicit in the runner.
- Local-runtime parity: `tests/webr-smoke.sh`'s Phase 3 currently inlines the same logic via docker heredocs. A future cleanup could have the local smoke source `tests/webr-node-smoke/smoke.mjs` directly — out of scope for this PR.

Refs #470 (umbrella), follows #491 / #744 (tier 2).

---

</details>


CGMossa:
Heads up on this one before anyone tries to merge it: the first CI run surfaced two blockers, and it's now gated on #751.

<details>

## Summary
- The tier-3 step failed on a Node module-resolution error, not on anything R/wasm: `Cannot find module '.../node_modules/webr/dist/webr.mjs'`. The `file:///opt/webr/src` link resolves to webR *source* whose `dist/` isn't built in the CI base image. The local `tests/webr-smoke.sh` Phase 3 has the same latent bug (it was never validated end-to-end). Fix is still being worked out (build webR's npm package, or install the published `webr` package and point `baseUrl` at the image's R wasm).
- Even once the Node resolution is fixed, this PR cannot go fully green until #751 (the #494 `-Zdefault-visibility=hidden` export-bloat fix) lands: without it, `library(miniextendr)` would fail at `dyn.load` with the ~3000-symbol EXPORT-table bloat (`TypeError: Cannot read properties of undefined`). Tier-3 is precisely the runtime validator for that fix.

## Notes
- Sequencing: merge #751 first → rebase this PR → fix the Node resolution → tier-3 goes green and becomes the standing runtime proof that the side-module loads.
- Holding this PR until #751 is in.

---

</details>

PR #757, opened 2026-05-29
fix(minirextendr): correct @inheritParams + restore FORCE_WRAPPER_GEN env

CGMossa:
Two small minirextendr scaffolding fixes: a malformed `@inheritParams` tag in `create_miniextendr_package()`, and a `Sys.setenv()` in `miniextendr_build()` that leaked into the R session.

<details>

## Summary
- `create.R`: `create_miniextendr_package()` documented its params with `@inheritParams usethis::create_package open rstudio check_name`. roxygen2's `@inheritParams` takes a single function and auto-fills any matching params, so the trailing arg list is invalid. Reduced to `@inheritParams usethis::create_package` (`open` / `rstudio` / `check_name` are all inherited).
- `workflow.R`: `miniextendr_build()` ran `Sys.setenv(MINIEXTENDR_FORCE_WRAPPER_GEN = TRUE)` and never unset it, leaking the override into the rest of the R session (forcing the cdylib wrapper-gen pass on every later build, including tarball-mode installs). Now sets it to `"1"` and restores the prior value on `on.exit`, matching the `COPYFILE_DISABLE` save/restore idiom already used in `miniextendr_vendor()`.

## Test plan
- [ ] `just minirextendr-document` (roxygen2) runs clean, with `open` / `rstudio` / `check_name` documented on `create_miniextendr_package()`.
- [ ] `just minirextendr-test` passes.
- [ ] After a `miniextendr_build()` call, `Sys.getenv("MINIEXTENDR_FORCE_WRAPPER_GEN")` is unset again in the session.

---

</details>

### DIFF
--- a/minirextendr/R/create.R
+++ b/minirextendr/R/create.R
@@ -7,12 +7,11 @@
 #' `use_miniextendr()`.
 #'
 #' @param path Path where to create the package
-#' @param open Whether to open the new project in RStudio
-#' @param rstudio Whether to create an RStudio project file
+#' @inheritParams usethis::create_package
 #' @return Path to the created package (invisibly)
 #' @export
 create_miniextendr_package <- function(path, open = interactive(),
-                                        rstudio = TRUE) {
+                                        rstudio = TRUE, check_name = FALSE) {
   # Validate package name (derived from directory basename)
   # R package names: ASCII letters, digits, and dots only.
   # Must start with a letter and not end with a dot. Minimum 2 characters.
@@ -30,7 +29,7 @@ create_miniextendr_package <- function(path, open = interactive(),
     path,
     open = FALSE,
     rstudio = rstudio,
-    check_name = FALSE
+    check_name = check_name
   )
 
   # Set project to the new package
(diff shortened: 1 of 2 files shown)


CGMossa:
Empirically reproduced and confirmed this fix on a fresh standalone scaffold (`/tmp/wraptest/wrapdemo`, no `.git` ancestor).

**Pre-#757 (bug):** a clean standalone build without the FORCE env hard-fails in tarball mode:
\`\`\`
configure: install mode = tarball install (offline, vendored)
tarball install: using pre-shipped R/-wrappers.R
ERROR: tarball is missing pre-generated .../R/wrapdemo-wrappers.R
ERROR: compilation failed for package 'wrapdemo'
\`\`\`

**With #757:** `miniextendr_build()` logs `MINIEXTENDR_FORCE_WRAPPER_GEN set: regenerating wrappers from cdylib (tarball mode override)`, generates `R/wrapdemo-wrappers.R`, NAMESPACE exports `add`/`hello`, and after install `library(wrapdemo); hello("world")` → `"Hello, world!"`. Env is `<unset>` after the build — the on.exit restore works too. This entry point is fully closed.

One residual gap I found and fixed in #774: the scaffolded `lib.rs` still told users to build with `devtools::document()` / `devtools::install()`, which bypass the FORCE env and reproduce the original empty-namespace bug. Follow-up regression-coverage issue filed as #775.

PR #865, opened 2026-06-05
fix(convert): relax TryFromList derive Error bound (#861)

### LINKED ISSUE #861: TryFromList derive hardcodes `Error = SexpError` → #[miniextendr(list)] fails E0271 on Vec<f64> fields
## Summary

`#[derive(TryFromList)]` (and therefore `#[miniextendr(list)]`, which bundles
it) fails to compile with **E0271** on any struct that has a `Vec<f64>` field
— or, more generally, any field whose `TryFromSexp::Error` associated type is
not exactly `SexpError`.

## Cause

`miniextendr-macros/src/list_derive.rs` emits, for every non-ignored field, a
where-clause bound that **pins the associated `Error` type** to `SexpError`:

- named structs — line ~222:
  ```rust
  bounds.push(parse_quote!(#ty: ::miniextendr_api::from_r::TryFromSexp<Error = ::miniextendr_api::from_r::SexpError>));
  ```
- tuple structs — line ~259: same bound.

`Vec<f64>` implements `TryFromSexp` with `type Error = SexpTypeError`, not
`SexpError`, so the equality bound `Error = SexpError` is unsatisfiable:

```
error[E0271]: type mismatch resolving `<Vec<f64> as TryFromSexp>::Error == SexpError`
```

## Reproduction

```rust
#[miniextendr(list)]
pub struct FitResult {
    pub estimate: Vec<f64>,   // <- E0271
    pub sigma: f64,
}
```

Workaround I used: drop `#[miniextendr(list)]` and use
`#[derive(IntoList, PreferList)]` for a return-only struct (skips the
`TryFromList` derive entirely). That's fine when you never parse the struct
*from* R, but it means `#[miniextendr(list)]` is unusable for any struct that
holds a numeric vector, even purely for returning.

## Suggested fix

Don't constrain the associated `Error` type. Relax the generated bound to
`#ty: TryFromSexp` plus a convertibility bound, and convert each field's error
on extraction, e.g.

```rust
#ty: ::miniextendr_api::from_r::TryFromSexp,
::miniextendr_api::from_r::SexpError: ::core::convert::From<<#ty as ::miniextendr_api::from_r::TryFromSexp>::Error>,
```

then let `?` (or an explicit `.map_err(SexpError::from)`) do the conversion in
the extraction code, instead of the current hard `Error = SexpError` equality.
(Assumes the relevant `From`/`Into` impls into `SexpError` exist — worth
confirming for `SexpTypeError` and the other field-error types.)

Found while wrapping a `diffsol` engine; `minirextendr` @ current `main`.

CGMossa:
Fixes #861.

<details>

## Summary
- `#[derive(TryFromList)]` (and `#[miniextendr(list)]`, which bundles it) pinned every field's `TryFromSexp::Error` to `SexpError` via the where-clause `TryFromSexp<Error = SexpError>`. Any field whose error is not exactly `SexpError` (e.g. `Vec<f64>`, whose error is `SexpTypeError`) failed to compile with E0271, making list mode unusable for structs holding numeric vectors even purely for returning.
- Relaxed the per-field bound to `TryFromSexp` plus `SexpError: From<<field as TryFromSexp>::Error>`, and convert at extraction time via `map_err(SexpError::from)`. All the `From<...>` impls into `SexpError` already exist (`SexpTypeError`, `SexpLengthError`, `SexpNaError`, and reflexive `SexpError`).
- The same over-constraint lived in `List::get_index` / `List::get_named` / `NamedList::{get, get_index}`, which discard the conversion error via `.ok()` so the bound was pure over-constraint. Relaxed those to `T: TryFromSexp`, and added `List::get_named_sexp` for raw by-name element access.
- Bonus correctness: the derive now fetches the raw element SEXP and converts directly, so a present-but-wrong-type field reports its real conversion error instead of being misreported as a missing field (named) or length error (tuple).
- Added a `MxFit { estimate: Vec<f64>, sigma: f64 }` fixture in `rpkg` with `mx_fit_create` (exercises `IntoList` with a numeric-vector field) and `mx_fit_roundtrip` (GC-safe `IntoList` -> `TryFromList` round-trip, the only path that drives the new `Vec<f64>` extraction at runtime since list-mode structs are not generated as argument types).
- Expanded the List Mode section in `docs/MINIEXTENDR_ATTRIBUTE.md` (what the three derives do, named/tuple/unit list shape, field-type requirements, `#[into_list(ignore)]`, and the return-only-by-default caveat).

## Test plan
- [x] `cargo test -p miniextendr-macros` — new `derive_try_from_list_does_not_pin_error_type` guard + updated `get_named_sexp` / `get` assertions pass.
- [x] `cargo test -p miniextendr-api -p miniextendr-macros` — green.
- [x] `just rcmdinstall` — the `Vec<f64>` list-mode struct now compiles end-to-end.
- [x] `just devtools-test` — FAIL 0 | PASS 6335 (includes the new #861 round-trip tests).
- [x] `cargo clippy` (default) + `clippy_all` curated feature list + `just clippy` (all manifests) — clean under `-D warnings`.
- [x] `cargo fmt --all` — no changes.
- [x] Regenerated `R/miniextendr-wrappers.R` + `NAMESPACE` + `man/*.Rd` + `wasm_registry.rs` committed in lockstep.

## Notes
- 7 pre-existing `readLines` "incomplete final line found" WARN in `test-connections.R` are unrelated to this change (different subsystem, untouched) and left out of scope.

---

</details>


CGMossa:
## CI root cause + fix (oldrel-1 compile failure)

The `R CMD check / Linux oldrel-1` failure was **not** a code bug in the relaxation — it was a **vendor latch leak via the CI cache** (the #441 class).

**Chain:** the `r-check-linux` and `cran-check` jobs restored a cached `rpkg/inst/vendor.tar.xz` *before* `setup-vendor` ran `just configure`. A present tarball flips `configure` into tarball mode → it writes a `[source]` replacement and **no** `[patch."git+url"]` → siblings. `just vendor`'s `cargo-revendor` needs that `[patch]` to copy `miniextendr-{api,lint,macros}` from the workspace; without it it reported **`Local packages to vendor: 0`** and vendored the framework from **git@main** (pre-relaxation). The new `MxFit { estimate: Vec<f64> }` fixture then failed against the old strict `TryFromSexp<Error = SexpError>` bound (E0271/E0599).

release/devel/cran-check masked it by reusing a cached `libminiextendr.a`; oldrel-1's evicted rust-cache forced a cold compile that surfaced it.

**Reproduced locally:** tarball-present → `configure` tarball-mode → `just vendor` → *"0 local packages"* → STRICT vendored macro; remove tarball first → DEV mode → *"3 local packages"* → RELAXED macro.

**Fix (this branch):**
- `setup-vendor`: `rm -f rpkg/inst/vendor.tar.xz` before `just configure`, guaranteeing dev-monorepo mode so the `[patch]` is written and local edits are vendored.
- `ci.yml`: removed both now-dead "Cache vendor tarball" steps (`just vendor --force` always regenerates, so caching it only armed the leak).

Also merged `origin/main` (RCoerce rename #872 + #875) and regenerated wrappers/NAMESPACE/man/wasm_registry in lockstep. Follow-up for a loud-fail guard: #876.

CGMossa:
Update: the CI vendor latch-leak fix has been split out into its own PR — #877. This branch now contains **only** the #861 relaxation (+ a merge of current `main`).

Because the relaxation adds an rpkg fixture (`MxFit`) that needs the framework vendored from local workspace source, this PR's `R CMD check` will stay red until #877 lands on `main`. **Merge #877 first**, then I'll sync `main` into this branch and CI will go green. Diagnosis details in the comment above and in #877.

Issue #876, opened 2026-06-08
vendor: fail loudly when a monorepo framework crate vendors from git instead of local workspace


CGMossa:
## Problem

`just vendor` (via `cargo-revendor`) **silently** vendored `miniextendr-{api,lint,macros}` from **git@main** instead of the local workspace checkout when `.cargo/config.toml` was missing the `[patch."git+url"]` → sibling-path overrides. This silently drops any in-PR edit to a framework crate, shipping a vendor tarball that doesn't match the PR's source.

This was the root cause of #865's CI failure (see PR #865). The chain:

1. CI `actions/cache` restored a `rpkg/inst/vendor.tar.xz` *before* `setup-vendor` ran `just configure`.
2. A present tarball flips `configure` into **tarball mode** → writes a `[source]` replacement, **no `[patch]`**.
3. `cargo-revendor` reads `[patch]` to copy local crates; with it absent it reported **`Local packages to vendor: 0`**, fell back to `--versioned-dirs`, and vendored the framework from git@main.
4. The pre-relaxation (git@main) macro failed to compile the PR's new `MxFit` fixture (E0271/E0599).

The whole thing was masked for ~20 minutes per job by stale `rust-cache` `libminiextendr.a` artifacts — only the one job whose cache was evicted surfaced it. **Silent vendoring of the wrong source is the dangerous part**; a loud failure would have caught it immediately on the first run.

## Proposal

`just vendor` knows this is the miniextendr monorepo, so it knows the three framework crates **must** vendor from local workspace source. After `cargo revendor`, assert that and fail otherwise:

- Parse cargo-revendor's `Local packages to vendor: N` (or inspect `rpkg/vendor/` layout: local crates land as `miniextendr-macros/`, git-fetched ones as `miniextendr-macros-0.1.0/` under `--versioned-dirs`).
- If any of `miniextendr-{api,lint,macros}` vendored from git (N == 0, or versioned-dir layout present), **error out** with a pointer to the latch-leak (configure ran in tarball mode; remove `inst/vendor.tar.xz` and re-`just configure`).

Optionally, a generic guard in `cargo-revendor` itself: warn when a `[patch."git+url"]` entry points at a path that does not exist (the patch silently doesn't apply).

## Acceptance

- `just vendor` fails (non-zero) with a clear message if a framework crate would be vendored from git rather than the local workspace.
- Regression coverage in the bootstrap-vendor test path.

Follow-up to PR #865.

PR #878, opened 2026-06-08
fix(vendor): fail loudly when a framework crate vendors from git instead of local (#876)

CGMossa:
## Problem

`just vendor` (via `cargo-revendor`) **silently** vendored `miniextendr-{api,lint,macros}` from `git@main` instead of the local workspace checkout whenever `.cargo/config.toml` was missing the `[patch."git+url"]` -> sibling-path overrides. That drops any in-PR edits to a framework crate and ships a vendor tarball that doesn't match the PR's source. It was the root cause of a multi-hour CI debugging session (#865 / #877).

The immediate trigger (a CI cache flipping `configure` into tarball mode -> `[source]` replacement, no `[patch]`) was fixed in #877 (merged, `87117330`). This is the **defense-in-depth follow-up (#876)**: make `just vendor` itself fail loudly so the next leak — from any source — is caught at the point of damage.

## Fix

- **`justfile` `vendor:` recipe** — after `cargo revendor`, capture its `--json` stdout (the `-v` progress still streams to stderr; the two coexist) and assert every framework crate appears in `local_crates`. If any was vendored from git, `exit 1` with the latch-leak remedy:
  `just clean-vendor-leak && just configure && just vendor`.
  Greps the JSON array rather than shelling out to `jq` — `jq` isn't guaranteed on every CI runner that runs `just vendor` (the `setup-vendor` action runs on ubuntu/almalinux/macos and installs no `jq`), and grep adds zero dependency.
- **`tests/vendor-loud-fail.sh`** — hermetic regression test wired into `just test-bootstrap-vendor` (the #876 acceptance test path). Builds a tiny package whose only dep is a crate served from a **local bare git repo** (`file://`, no crates.io deps, no network), with **no `[patch]` override** — the exact latch-leak shape — and runs the recipe's assertion to prove it flags the git-vendored framework crate. Skips (exit 0) if `cargo-revendor` isn't on PATH, matching `bootstrap-produces-vendor.sh`.
- **`cargo-revendor`** (optional guard from the issue, implemented):
  - `discover_from_patch_config` now emits a `warning:` when a `[patch."git+url"]` entry's resolved path has **no `Cargo.toml`** (the override silently doesn't apply otherwise).
  - 3 new always-run unit tests (`patch_config_resolves_existing_path`, `patch_config_skips_missing_path`, `patch_config_no_patch_table_is_empty`).
  - 2 new `#[ignore]` (network) integration tests in `patch_from_config.rs` asserting the `--json local_crates` signal includes a patched local crate and excludes a git-fetched one.

### Why `local_crates` (not dir layout)

The issue body suggested discriminating local-vs-git by directory shape (`foo/` vs `foo-0.1.0/`). That's stale: `versioned_dirs` is now the default, so **both** land as versioned dirs. The structured `--json` `local_crates` array is the authoritative signal — it lists exactly the crates `cargo-revendor` vendored from local source.

## Verification

- `cargo fmt --all --manifest-path cargo-revendor/Cargo.toml --check` clean.
- `just revendor-clippy` (`-D warnings`) clean; `just revendor-test` 105 pass / 62 ignored (incl. the 3 new unit tests + 2 new integration tests).
- New integration tests run green with `--ignored` (local `file://` git, no network).
- `tests/vendor-loud-fail.sh` -> `OK: loud-fail assertion correctly flagged git-vendored framework crate(s): miniextendr-api miniextendr-lint miniextendr-macros`.
- **Happy path** end-to-end: `just configure` (dev mode) + `just vendor` -> exit 0, "Local packages to vendor: 3", new success line `Vendored framework crates from local workspace: miniextendr-{api,lint,macros}`.
- **Simulated leak** end-to-end: stripped the `[patch]` block from `.cargo/config.toml`, re-ran `just vendor` -> **exit 1** with:
  > ERROR: framework crate(s) miniextendr-api miniextendr-lint miniextendr-macros were vendored from git, not the local workspace. ... Fix:  just clean-vendor-leak && just configure && just vendor
- Cleanup: restored `.cargo/config.toml`, `just clean-vendor-leak`, reverted transient `Cargo.lock` drift from the vendor run, removed `rpkg/vendor/`. No vendor tarball or stray tracked files left behind (`git status` shows only the 4 intended changes).

Related: #877 (CI trigger fix, merged), #865 (the original failure).

### DIFF
--- a/cargo-revendor/src/metadata.rs
+++ b/cargo-revendor/src/metadata.rs
@@ -343,6 +343,20 @@ pub fn discover_from_patch_config(manifest_path: &Path) -> Result<Vec<LocalPacka
 
                 let crate_manifest = crate_path.join("Cargo.toml");
                 if !crate_manifest.exists() {
+                    // The `[patch]` entry points at a path with no Cargo.toml,
+                    // so cargo silently doesn't apply the override and the crate
+                    // is vendored from its git/registry source instead. For a
+                    // monorepo framework crate this silently drops in-PR edits
+                    // (the #865/#876 latch-leak failure mode). Warn loudly so the
+                    // misconfiguration is visible; the caller's local-crate
+                    // assertion is what ultimately fails the build.
+                    eprintln!(
+                        "  warning: [patch] entry for `{crate_name}` in {} points at `{}`, \
+                         which has no Cargo.toml — the override will NOT apply and `{crate_name}` \
+                         will be vendored from its git/registry source instead of this path.",
+                        config_path.display(),
+                        crate_path.display(),
+                    );
                     continue; // path doesn't resolve on this machine — skip
                 }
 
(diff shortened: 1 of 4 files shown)


CGMossa:
This is an issue with making AI recognize that the way to build miniextendr R packages is through first generating a build-tarball via `R CMD build .` followed by then an install `R CMD INSTALL <build-tarball>`.

Issue #883, opened 2026-06-08
vendor: lock pre-resolution targets git@main, not PR HEAD — cross-crate cargo-surface PRs can't get green CI


CGMossa:
> _Placeholder for framing — filed from the session that merged #710; this is the underlying infra bug that forced #710 to be admin-merged red._

<details>
<summary><h4>AI-written details</h4></summary>

## Problem
A PR that changes the **cargo feature/dependency surface** of `miniextendr-{api,macros,lint}` *and* the `rpkg` consumer in the same commit (e.g. the #710 `default-* → *-default` rename) cannot get green CI before merge. Six jobs fail (`Bootstrap Vendor Test`, `Sync Checks`, all `R CMD check`, `CRAN-like check`), all with:

```
error: failed to select a version for `miniextendr-api`.
package `miniextendr` depends on `miniextendr-api` with feature `coerce-default`
but `miniextendr-api` does not have that feature.
  available features: ... default-coerce, default-r6, ...   ← main's OLD names
```

Today this is "resolved" by admin-merge (see `docs/CARGO_LOCK_SHAPE.md`), which works (the next PR's bootstrap is green once main has the rename) but is a workaround, not a fix.

## Root cause (one step, not a cargo law)
The vendored tarball already ships the PR's code: `cargo-revendor` copies the framework crates from the **local workspace checkout**, with a `--json`/`local_crates` assertion (#876) that fails the build if it ever ships `git@main`. The breakage is isolated to the **lock pre-generation** in `rpkg/bootstrap.R` (and the mirror in `just vendor`, `justfile:451`):

1. move `.cargo/config.toml` aside (kills the `[patch."git+url"]` → local redirect),
2. `rm Cargo.lock`,
3. `cargo generate-lockfile` against the **bare git URL with no `rev`** → cargo fetches `main`'s HEAD.

That step exists only to stamp `source = "git+url#<sha>"` into the lock so the offline tarball's `[source]` replacement matches. But `cargo generate-lockfile` does a *full feature resolution*, and resolving `rpkg`'s new-name request against `main`'s api surface fails. The content is correct; the pre-resolution targets the wrong commit (`main`, not the PR HEAD whose sources will actually be vendored).

## Why CARGO_LOCK_SHAPE.md's "can't fix cleanly" is too pessimistic
[...]

PR #888, opened 2026-06-08
fix(minirextendr): repair_lock() retains checksum lines (post-#408 policy)

### LINKED ISSUE #886: minirextendr: miniextendr_repair_lock() still strips checksum lines (pre-#408 behavior)
## What

`minirextendr::miniextendr_repair_lock()` (`minirextendr/R/repair-lock.R`) still
strips `checksum = "..."` lines from `src/rust/Cargo.lock` (see lines 17, 54,
65, 107-109). The rationale comment (107-109) reads:

> Strip checksum lines — vendored crates ship with empty `.cargo-checksum.json`;
> registry checksums in the lock cause offline install to abort with "the listed
> checksum of '<crate>' has changed".

That stopped being true after **#408**: `cargo-revendor` now writes *valid*
`.cargo-checksum.json` files (real SHA-256s, `package` field preserved, `files`
map updated post-CRAN-trim), so cargo's offline source-replacement verifies the
retained `checksum =` lines successfully.

This is the same realization that drove dropping checksum-stripping from
`just vendor`, `miniextendr_vendor()` (workflow.R), and the lock-shape docs in
the #883 PR. `repair-lock.R` was missed because it's an end-user repair tool,
not part of the maintainer vendor flow, and end users don't hit the #883
cross-crate-rename path. But the stale strip means `repair_lock()` now diverges
from the canonical `miniextendr_vendor()` output (which retains checksums) — a
user who runs the repair tool ends up with a lock that differs from a freshly
vendored one.

## Fix sketch

- Drop the checksum-stripping branch from `repair_lock()`; keep only the
  `path+` → `git+url#<sha>` source repair (the part that's still needed).
- Update the function's `@description`/`@details` roxygen (lines 6-17) and the
  inline comment (107-109) to match: checksums are retained; only `path+`
  sources are rewritten.
- Reconcile with `docs/CARGO_LOCK_SHAPE.md` ("checksum lines are now allowed").
- Add/adjust a `tests/testthat` case asserting checksums survive a repair.

## Context

Spun out of the #883 vendor-local-lock-resolution PR (which fixed the
maintainer-side flow). Not a #883 blocker — filed so the end-user repair tool
catches up to the post-#408 checksum policy.

CGMossa:
Spun out of #883. `miniextendr_repair_lock()` was the last place still stripping `checksum =` lines from `src/rust/Cargo.lock` — stale pre-#408 behavior that made the end-user repair tool produce a lock diverging from a freshly vendored one. This drops the strip and reconciles `doctor.R`, which was flagging the (now-canonical) checksum lines as drift.

<details>

## Summary
- `minirextendr/R/repair-lock.R`: dropped the checksum-stripping branch. Tarball-shape is now defined solely by the absence of `path+` source entries. The `path+` → `git+url#<sha>` source repair (move `.cargo/config.toml` aside, `cargo update`, verify no `path+` remain) is kept unchanged. `cargo update` already writes the lockfile, so the function no longer rewrites it. Roxygen `@description`/`@details` and the inline comment updated to state checksums are retained (since #408 `cargo-revendor` writes valid `.cargo-checksum.json` files).
- `minirextendr/R/doctor.R`: the Cargo.lock-shape check counted `checksum =` lines as "drifted into source-shape" and told users to run `miniextendr_repair_lock()`. Post-fix that was a contradiction (doctor said "drifted", repair said "already tarball-shape"). Now only `path+` source entries count as drift.
- `minirextendr/tests/testthat/test-repair-lock.R`: rewrote the old "strips checksum lines" case to assert a checksum-only lock is treated as already-tarball-shape and left byte-for-byte untouched; added a case asserting checksums survive a real `path+` repair; switched the config-move test to a `path+` fixture (the old checksum-only fixture now short-circuits before touching `.cargo/config.toml`).
- `minirextendr/man/miniextendr_repair_lock.Rd`: regenerated from the roxygen changes.
- `docs/CARGO_LOCK_SHAPE.md` already states checksum lines are allowed (landed with #883), so no doc change was needed there.

Closes #886.

## Test plan
- [x] `just minirextendr-test repair-lock` → `FAIL 0 | PASS 14`
- [x] `just minirextendr-test git-hooks-lock-shape` → `FAIL 0 | PASS 8`
- [x] `just minirextendr-test regressions` (exercises `miniextendr_doctor()`) → `FAIL 0 | PASS 33`
- [x] `just minirextendr-document` regenerated the `.Rd` cleanly
- [ ] CI green on the full `minirextendr-job`

## Notes
- The full local `just minirextendr-test` run hung on an unrelated scaffolding/`R CMD check` test (frozen for ~25 min with no progress, no relation to `repair-lock.R` / `doctor.R`); verification was done via the three relevant test files above, which cover both touched files.
- The `miniextendr_doctor()` smoke test in `test-regressions.R` does not write a `Cargo.lock`, so it never reaches the changed lock-shape branch — there is no existing test asserting doctor's checksum-vs-drift behavior, and adding one was out of the minimal scope here.

---

</details>


CGMossa:
I still have not figured out if we ought to strip checksums or not. But this is more a feature geared towards CRAN acceptance, rather than enhancing (mini)extendr.

Issue #894, opened 2026-06-08
Optional: warn on relative local path-dep in src/rust/Cargo.toml (follow-up to #864)


CGMossa:
Follow-up to #864 (docs PR documents the absolute-path requirement).

## Idea

`miniextendr_doctor()` / `configure` could detect a relative `path = ...`
dependency in `src/rust/Cargo.toml` and warn that it will break under
`R CMD INSTALL`'s temp-copy (a relative path resolves against the temporary
build dir, not the real crate location), suggesting an absolute path.

`miniextendr_doctor()` already reads `src/rust/Cargo.toml` (the
`miniextendr-api` declaration check), so it's the natural home.

## Why it wasn't implemented in the #864 docs PR

The naive check (flag any relative `path = ...` in `[dependencies]` /
`[patch.crates-io]`) **false-positives on the supported `use_vendor_lib()`
flow**, which *intentionally* writes a relative `dev_path` into
`[patch.crates-io]`:

```toml
[patch.crates-io]
my_engine = { path = "../../../my_engine" }
```

That relative path is correct there — the generated `configure.ac`
`vendor-lib-<crate>` block rewrites it to the extracted vendor copy in
tarball/CRAN mode, so it never reaches the offline build. A useful warning
therefore has to distinguish:

- a **hand-rolled** relative `path = ...` with no rewriting machinery (bad), from
- a **`use_vendor_lib()`-managed** relative `dev_path` whose `configure.ac`
  block rewrites it (fine).

Distinguishing them means cross-checking `configure.ac` for a matching
`vendor-lib-<crate>` block per relative-path crate — doable but fragile, and
out of scope for a docs PR.

## Possible designs

- Warn only on relative `path = ...` in `[dependencies]` (not
  `[patch.crates-io]`), since `use_vendor_lib()` only writes to the patch
  table — narrower, fewer false positives.
- Or: for each relative-path crate, suppress the warning when a
  `vendor-lib-<crate>` `AC_CONFIG_COMMANDS` block exists in `configure.ac`.
- Or: `configure` rewrites the relative path to absolute for the temp build
  (the issue's "or rewrite it to absolute" suggestion) — heavier, mutates
  config during build (violates the "configure.ac never mutates sources" rule
  unless scoped to the temp copy only).

Docs now cover the manual case (#864), so this is genuinely optional.

PR #895, opened 2026-06-08
docs: local path-dep in src/rust/Cargo.toml must be absolute (#864)

### LINKED ISSUE #864: Docs: local path-dependencies in src/rust/Cargo.toml must be absolute (R CMD INSTALL temp-copy)
## Summary (docs request)

Document (and ideally warn at scaffold/build time) that a **local path
dependency** in `src/rust/Cargo.toml` must use an **absolute** path, not a
relative one.

## Why

`R CMD INSTALL` / `devtools::install()` copy the package to a temporary build
directory (e.g. `/private/tmp/.../R.INSTALL.../<pkg>/`) before compiling the
Rust staticlib. A relative path dep like

```toml
my_engine = { path = "../../../engine" }
```

resolves against that temp location and fails:

```
error: failed to load manifest for dependency `my_engine`
  ... No such file or directory
```

An absolute path works, because the engine source is read live from its real
location at install time (cargo-vendor captures the engine's *registry* deps;
the engine's own source is read from the path):

```toml
my_engine = { path = "/abs/path/to/engine" }
```

## Suggested action

- A line in the docs / scaffolder README on wrapping a local engine crate.
- Optionally, `miniextendr_build()` / `configure` could detect a relative
  `path = ...` dep and warn (or rewrite it to absolute for the temp build).

Found while wrapping a local engine crate; `minirextendr` @ current `main`.

CGMossa:
Closes #864.

Documents that a **local `path = ...` dependency** in `src/rust/Cargo.toml`
(wrapping an "engine" crate or sibling library) must use an **absolute** path,
not relative. `R CMD INSTALL` / `devtools::install()` copy the package into a
temporary build dir before compiling the Rust staticlib, so a relative path
resolves against the temp location and fails:

```toml
my_engine = { path = "../../../engine" }   # ❌ failed to load manifest for dependency
my_engine = { path = "/abs/path/to/engine" }  # ✅ read live from its real location
```

`cargo vendor` captures only the engine's *registry* deps; the engine's own
source is read from the `path` at install time — which is why the path must
keep resolving after the temp-copy.

## What changed

- **`docs/MINIREXTENDR.md`** — new "Wrapping a local Rust crate: use an
  absolute path" subsection under Vendoring, with the failing-vs-working
  `Cargo.toml` snippets from the issue and the cargo-vendor interaction. Notes
  the contrast with `use_vendor_lib()` (its relative `dev_path` is intentional
  — rewritten to the vendor copy in tarball mode — so the absolute-path rule
  applies to *hand-added* deps only).
- **`docs/TROUBLESHOOTING.md`** — "failed to load manifest for dependency"
  entry under Vendor / Configure Issues, cross-linked to the above.
- **`minirextendr/R/vendor-lib.R`** — `use_vendor_lib()` gains an `@details`
  block + clarified `@param dev_path` explaining the relative-vs-absolute
  distinction; regenerated `man/use_vendor_lib.Rd`.

## Warning at scaffold/build time

Not implemented here — a naive relative-`path` check in `miniextendr_doctor()`
false-positives on the supported `use_vendor_lib()` flow (which intentionally
writes a relative `dev_path` into `[patch.crates-io]`, rewritten by its
`configure.ac` block in tarball mode). Filed as follow-up **#894** with the
design options. The manual case the issue describes is now fully documented.

### DIFF
--- a/docs/MINIREXTENDR.md
+++ b/docs/MINIREXTENDR.md
@@ -251,6 +251,50 @@ vendor_sync(path = "mypackage")
 miniextendr_available_versions()
 ```
 
+### Wrapping a local Rust crate: use an absolute path
+
+When your R package depends on a Rust crate that lives **outside the package**
+(e.g. an "engine" crate in the same monorepo, or a sibling library on disk),
+the `path = ...` dependency you add to `src/rust/Cargo.toml` **must be
+absolute**, not relative.
+
+`R CMD INSTALL` and `devtools::install()` copy the package into a temporary
+build directory (e.g. `/private/tmp/.../R.INSTALL.../<pkg>/`) *before*
+compiling the Rust staticlib. A relative path resolves against that temporary
+location — which doesn't contain your engine crate — and the build fails:
+
+```toml
+# ❌ breaks: relative path resolves against the temp build dir
+my_engine = { path = "../../../engine" }
+```
+
+```
+error: failed to load manifest for dependency `my_engine`
+  ... No such file or directory
+```
+
+An **absolute** path is read live from its real location, so the temp-copy
+doesn't matter:
+
+```toml
+# ✅ works: absolute path is read from its real location at install time
+my_engine = { path = "/abs/path/to/engine" }
+```
+
+How this interacts with vendoring: `cargo vendor` / the vendor tarball captures
+only the engine crate's **registry** dependencies (its transitive crates.io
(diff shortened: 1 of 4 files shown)


CGMossa:
They don't have to be absolute, only due to staged installs, and this is a whole headache we have to address, somehow.

PR #896, opened 2026-06-08
docs(troubleshooting): dlopen flat-namespace stub.c FAQ entry

### LINKED ISSUE #863: FAQ: dlopen 'symbol not found in flat namespace' when staticlib→.so doesn't dead-strip like rustc — stub in src/stub.c
## Summary (docs request)

Add an FAQ entry covering `dyn.load` failures of the form:

```
unable to load shared object '.../<pkg>.so':
  dlopen(...): symbol not found in flat namespace '_<symbol>'
```

that occur when a miniextendr package links a Rust staticlib bundling a heavy
native dependency (LLVM, etc.).

## Why it happens

miniextendr links the Rust crate as a `staticlib` (`.a`) and R then links that
into `<pkg>.so` with `clang ... -undefined dynamic_lookup`. That flag defers
undefined symbols to load time, and `dyn.load` uses `RTLD_NOW`, so any
unresolved symbol fails at load.

A `cargo build`/`cargo test` **binary** of the same crate links and runs fine,
because `rustc`'s binary link **dead-strips** unreachable code — including
references to symbols that are never actually called. R's `.so` link does not
dead-strip those references (and `-Wl,-dead_strip` in `PKG_LIBS` doesn't help
when the staticlib is a single `codegen-units = 1` object that keeps the
reference statically reachable). The result: a symbol that is *referenced but
never called* — e.g. a deprecated API a dependency dropped, or one removed in a
newer toolchain version — links in a binary but breaks `dlopen`.

## Suggested guidance for the FAQ

1. Diagnose: build the staticlib (`cargo build --release --lib`) and compare
   undefined vs. defined symbols, e.g.
   ```sh
   comm -23 <(nm -u LIB | awk '{print $NF}' | sort -u) \
            <(nm -g --defined-only LIB | awk '{print $NF}' | sort -u)
   ```
   then cross-check the residual against the native libs you link.
2. If the residual symbols are genuinely unreachable (the cargo binary runs
   without them), satisfy them with **abort-on-call stubs in `src/stub.c`** —
   R compiles `src/*.c` and links the `.o` into the `.so` via `$(OBJECTS)`, so
   the references resolve at link time and abort loudly if ever invoked.

Concrete instance I hit: `diffsol → diffsl → inkwell` references
`LLVMConstMul` / `LLVMConstNSWMul` / `LLVMConstNUWMul`, which **LLVM 21
removed**; stubbing them in `src/stub.c` made the package load. This is a
dependency-vs-toolchain issue, not a miniextendr bug — but the staticlib→`.so`
linking model is miniextendr-specific, so an FAQ pointer would save the next
person the multi-hour dig.

CGMossa:
## Summary

Adds a new **"Load Issues"** section to `docs/TROUBLESHOOTING.md` covering the `dlopen` / "symbol not found in flat namespace" failure mode requested in #863.

- **File changed**: `docs/TROUBLESHOOTING.md` — new `## Load Issues` section with one subsection
- **Content**: symptom (macOS flat namespace / Linux undefined symbol), root cause (staticlib + `-undefined dynamic_lookup` + `RTLD_NOW` + no dead-strip), `nm`-based diagnosis command, abort-on-call `src/stub.c` stub fix, and the concrete `diffsol → diffsl → inkwell` / LLVM 21 example from the issue
- **Cross-reference**: notes that `stub.c` already exists for the force-link anchor and these abort stubs are an *additional, separate* use of the same file

Closes #863

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -163,6 +163,118 @@ Ensure the function is `pub` and has `#[miniextendr]`, then run `just rcmdinstal
 
 ---
 
+## Load Issues
+
+### `dlopen` fails: "symbol not found in flat namespace '_<symbol>'"
+
+**Symptom** (macOS):
+
+```
+unable to load shared object '.../mypkg.so':
+  dlopen(.../mypkg.so, 0x000A): symbol not found in flat namespace '_LLVMConstMul'
+```
+
+(Linux shows `undefined symbol` instead of "flat namespace".)
+
+**Why it happens**
+
+miniextendr links the Rust crate as a `staticlib` (`.a`) and R then links that
+archive into `<pkg>.so` with:
+
+```
+clang ... -undefined dynamic_lookup ... libmypkg.a
+```
+
+`-undefined dynamic_lookup` defers every unresolved symbol to load time.
+`dyn.load()` uses `RTLD_NOW`, so *every* unresolved symbol must be present in
+an already-loaded library when the `.so` is first opened -- or the load fails.
+
+A `cargo build`/`cargo test` binary of the same crate may link and run fine.
+That's because `rustc`'s **binary** linker dead-strips unreachable code,
+including dead references to symbols that are never actually called. R's `.so`
+link does not dead-strip those references (and `-Wl,-dead_strip` in `PKG_LIBS`
+doesn't help when `codegen-units = 1` produces a single object that keeps the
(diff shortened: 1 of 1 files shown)


CGMossa:
I should probably make these annoying pieces add-on to R package. Certain makes things complicated, like ALTREP classes, and also this export-trait-r-pkg feature, etc.

PR #900, opened 2026-06-08
refactor(serde): deduplicate SEXPTYPE name table — delegate to sexptype_name

### LINKED ISSUE #833: refactor: deduplicate the SEXPTYPE -> R-type-name table
## Problem

The `SEXPTYPE` -> human-readable R-type-name mapping exists in (at least) two places that drift independently:

- `typed_list::sexptype_name` (`miniextendr-api/src/typed_list.rs:652-683`) — the full table (`INTSXP => "integer"`, `REALSXP => "numeric"`, `STRSXP => "character"`, …).
- `serde/de.rs:75-78` — a partial inline copy of the same arms.

A quick `grep -rn 'REALSXP => "' miniextendr-api/src` finds both; check for any further partial copies while you're in there.

## Proposed

Consolidate on one canonical helper. Two options:

1. Have `serde/de.rs` call `crate::typed_list::sexptype_name` (cheapest).
2. Have `sexptype_name` delegate to R's own `Rf_type2char(SEXPTYPE)` (returns the canonical R type name as a C string) so there is a single source of truth that matches R exactly.

If you take (2), note R's canonical names differ from the current table in spots (e.g. R calls `REALSXP` `"double"`, not `"numeric"`). Reconcile any test/error-message expectations that depend on the current spelling before switching.

**Priority: low.**

CGMossa:
## Summary

Took **Option 1** from #833: have `serde/de.rs` call `crate::typed_list::sexptype_name` instead of maintaining its own inline match arms.

- **Tables matched exactly** — every arm in `serde/de.rs` (`NILSXP`, `LGLSXP`, `INTSXP`, `REALSXP`, `STRSXP`, `RAWSXP`) had identical spellings to the canonical `typed_list::sexptype_name`. This change is **behaviour-preserving**.
- `VECSXP` retains its intentional extension inline: serde distinguishes `"named list"` vs `"list"` for clearer error messages; `sexptype_name` just returns `"list"` for both. The arm is preserved as a special-case `if` before delegating everything else.
- No visibility change needed: `sexptype_name` is already `pub` and re-exported from `miniextendr-api/src/lib.rs`.

**Files changed:**
- `miniextendr-api/src/serde/de.rs` — removed 5 duplicated match arms, replaced with `crate::typed_list::sexptype_name(t)` call.

**Option 2** (`Rf_type2char`) is noted for future consideration — it would provide a single C-level source of truth but changes canonical spellings (e.g. `REALSXP` becomes `"double"` not `"numeric"`) and would require reconciling many error-message/test expectations across the codebase. Out of scope for a low-priority dedup.

Closes #833

### DIFF
--- a/miniextendr-api/src/serde/de.rs
+++ b/miniextendr-api/src/serde/de.rs
@@ -71,21 +71,17 @@ impl RDeserializer {
 
     fn type_name(&self) -> String {
         let t = self.sexp_type();
-        match t {
-            SEXPTYPE::NILSXP => "NULL".to_string(),
-            SEXPTYPE::LGLSXP => "logical".to_string(),
-            SEXPTYPE::INTSXP => "integer".to_string(),
-            SEXPTYPE::REALSXP => "numeric".to_string(),
-            SEXPTYPE::STRSXP => "character".to_string(),
-            SEXPTYPE::VECSXP => {
-                if self.has_names() {
-                    "named list".to_string()
-                } else {
-                    "list".to_string()
-                }
+        // VECSXP gets special treatment: distinguish named vs unnamed list
+        // for clearer serde error messages. Everything else delegates to the
+        // canonical table in typed_list.
+        if t == SEXPTYPE::VECSXP {
+            if self.has_names() {
+                "named list".to_string()
+            } else {
+                "list".to_string()
             }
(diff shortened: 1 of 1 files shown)


CGMossa:
Option 2 doesn't seem like a good idea, but it might resolve things faster, as Rust may not be using the address as the identifier, as C would.

PR #901, opened 2026-06-08
perf(optionals): read logical R columns via as_slice instead of per-element logical_elt

### LINKED ISSUE #832: perf(optionals): read primitive R columns via as_slice instead of per-element *_elt loops
## Problem

Several optional-feature conversions read a whole primitive column one element at a time via `*_elt` getters, where `SexpExt::as_slice::<T>()` gives a flat slice and avoids the per-element FFI call:

- `optionals/bitvec_impl.rs:75-87` (and the sibling at `:157`) — a `logical_elt(i)` loop building `RBitVec`; could iterate `sexp.as_slice::<RLogical>()`.
- `optionals/arrow_impl.rs:585` — `logical_elt(i)` loop.

These are correctness-fine, just slower and stylistically inconsistent with `convert::gather_native` and `list::from_scalars_or_list`, which already read via `as_slice`.

## Proposed

Convert the bulk reads to `as_slice::<T>()` and iterate the slice. Keep the existing NA checks (`RLogical::NA` / `NA_LOGICAL`). String/list columns stay element-wise (write-barriered, no flat slice).

## Out of scope

The serde Deserializer paths (`serde/de.rs`, `serde/dataframe_de.rs`) read one value at a time *by design* — a streaming visitor model, not a bulk copy. Leave them alone.

**Priority: low.**

CGMossa:
## Summary

- **`optionals/bitvec_impl.rs` (both sites):** `RBitVec` (`Lsb0`) and `BitVec<u8, Msb0>` `TryFromSexp` impls now call `sexp.as_slice::<RLogical>()` and iterate the slice, replacing the `for i in 0..len { sexp.logical_elt(i) }` index loop. Dropped now-unused `NA_LOGICAL` import.
- **`optionals/arrow_impl.rs` (~line 623):** `BooleanArray` `TryFromSexp` impl converted the same way.

NA semantics are identical: `val.is_na()` (`RLogical::is_na()` checks `self.0 == i32::MIN`) replaces `== NA_LOGICAL`/`crate::altrep_traits::NA_LOGICAL`. The NA branch behaviour (bitvec: early-return `InvalidValue`; arrow: `append_null`) is unchanged.

**Out of scope (untouched):**
- `serde/de.rs` and `serde/dataframe_de.rs` — streaming-visitor model, element-wise by design.
- `StringArray` and all string/list columns — write-barriered, no flat slice available.

Closes #832

### DIFF
--- a/miniextendr-api/src/optionals/arrow_impl.rs
+++ b/miniextendr-api/src/optionals/arrow_impl.rs
@@ -620,15 +620,14 @@ impl TryFromSexp for BooleanArray {
             }
             .into());
         }
-        let len = sexp.len();
-        let mut builder = arrow_array::builder::BooleanBuilder::with_capacity(len);
+        let slice: &[crate::RLogical] = unsafe { sexp.as_slice() };
+        let mut builder = arrow_array::builder::BooleanBuilder::with_capacity(slice.len());
 
-        for i in 0..len {
-            let val = sexp.logical_elt(i as R_xlen_t);
-            if val == crate::altrep_traits::NA_LOGICAL {
+        for &val in slice {
+            if val.is_na() {
                 builder.append_null();
             } else {
-                builder.append_value(val != 0);
+                builder.append_value(val.to_i32() != 0);
             }
         }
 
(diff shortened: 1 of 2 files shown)


CGMossa:
This is good enough for now, but there might be some asm inspection necessary to figure out if autovectorisation is happening (unfortunately, it doesn't happen at all atm with the lack of --target-cpu=native)

Issue #510, opened 2026-05-11
feat: unified GC protection API (Protector trait + pool)


CGMossa:
# Plan: Unified GC Protection API

Background: `analysis/gc-protection-strategies.md`

## Goal

Replace the fragmented protection landscape (manual `Rf_protect`/`Rf_unprotect` calls,
`preserve.rs` DLL, `gc_protect.rs` scope) with a unified `Protector` trait. Every FFI
function that returns an unprotected SEXP gets a safe wrapper that returns a protected
handle. Users never call `Rf_protect` directly.

## Which R primitives are used, and when

Two R mechanisms, plus one Rust-side mechanism:

1. **Protect stack** (`Rf_protect` / `Rf_unprotect`) — for temporaries within a `.Call`.
   7.4 ns/op, LIFO, 50k limit. Exposed via `ProtectScope`. Also provides
   `R_ProtectWithIndex`/`R_Reprotect` for replace-in-loop (3.8 ns/op, no stack growth).

2. **R's precious list** (`R_PreserveObject` / `R_ReleaseObject`) — for a small number
   of long-lived cross-call objects that are NEVER released in a loop. 13 ns/op for
   single protect+release. SEXP is its own handle — zero Rust-side bookkeeping.
   O(n) release but LIFO-fast (recently preserved objects found first). Ideal for
   ExternalPtr (you already hold the SEXP, there are few of them, they live a long time).
   **NEVER use in a loop** — 15 seconds at 10k iterations due to O(n²) list growth.

3. **VECSXP pool** (new, replaces DLL preserve list) — for all other cross-call usage.
   9.6 ns/op single, 2.5x faster than DLL at batch, 7x faster at replace-in-loop.
   One GC-traced VECSXP with slot management via `slotmap` (11.4 ns/op with safety)
   or `Vec` (9.6 ns/op raw). Zero R allocation per insert/release. Generational keys
   prevent stale-handle bugs.

The **DLL preserve list** (`preserve.rs`) is retired. Benchmarks show it's 3-7x slower
than the VECSXP pool across all workloads (28.9 ns vs 9.6 ns single-op, 271 µs vs 45 µs
replace-in-loop at 10k). Its only theoretical advantage (R-object handle, memory shrinkage)
didn't produce measurable benefits.

### Decision tree

```
Is the SEXP returned directly to R (single allocation, no intermediate)?
  → No protection needed. R protects on receipt.

Is the SEXP temporary within a .Call (used, then discarded before return)?
[...]

PR #941, opened 2026-06-09
build(miniextendr-api): demote refcount-fast-hash from default features

### LINKED ISSUE #679: build: re-evaluate whether `refcount-fast-hash` belongs in `default` features
\`miniextendr-api\` ships with \`default = [\"doc-lint\", \"refcount-fast-hash\"]\`. The \`refcount-fast-hash\` feature switches \`refcount_protect::Arena\`'s map backing from \`std::collections::HashMap\` to \`ahash\`, pulling \`ahash\` into the dependency graph for every default user.

## The question

The \`refcount_protect\` module is one of three GC-protection mechanisms exposed by the api crate. Grepping for callers across the whole workspace turns up:

- Tests in \`rpkg/src/rust/refcount_protect_tests.rs\`
- The module's own implementation files

No production callers in the rpkg exemplar or anywhere else. Most users go through \`gc_protect::OwnedProtect\` / \`ProtectScope\` / \`protect_pool::ProtectPool\` — \`refcount_protect\` is the third pillar of an over-large set of options (see also #510).

So: every default user pays for \`ahash\` (transitive: zerocopy, version, etc.) without using \`refcount_protect\`. That looks like default-feature debt.

## Proposal

Demote \`refcount-fast-hash\` from \`default\`:

\`\`\`diff
- default = [\"doc-lint\", \"refcount-fast-hash\"]
+ default = [\"doc-lint\"]
\`\`\`

Users who want the fast-hash variant of \`Arena\` opt in explicitly.

Sanity-check before merging:

1. Confirm no crate or test relies on the default flag being on.
2. Measure binary size delta with/without (likely small, but worth a number for the PR body).
3. Re-test CI's curated clippy_all without the flag.

## Related

- #509 — scope-as-allocator
- #510 — unified GC protection API (proposes consolidating ProtectPool + Arena anyway)
- audit recommendation alongside PR #675: also consider whether \`refcount_protect.rs\`'s full surface (8 reexports — \`Arena\`, \`RefCountedArena\`, \`HashMapArena\`, \`ThreadLocalArena\`, \`MapStorage\` trait, …) is justified.

CGMossa:
## Default-feature debt

`miniextendr-api` shipped `default = ["doc-lint", "refcount-fast-hash"]`. The `refcount-fast-hash` feature adds an ahash-backed `FastHashMap` / `FastHashMapArena` variant of `refcount_protect::Arena`, pulling `ahash` (and its transitives) into the dependency graph for **every default user**.

But `refcount_protect` has **no production callers**. Grepping the workspace, the only references are:

- The module's own implementation (`miniextendr-api/src/refcount_protect.rs`).
- `rpkg/src/rust/refcount_protect_tests.rs` — which uses the BTreeMap-based `RefCountedArena`, *not* any fast-hash type.
- Opt-in-only consumers: `miniextendr-bench`'s own `refcount-fast-hash` forwarding feature + the feature-gated `fast_hash_benches` module, and `miniextendr-cli`'s `bench` command (passes the feature explicitly to `cargo bench`).

Production code goes through `gc_protect::OwnedProtect` / `ProtectScope` / `protect_pool::ProtectPool` — `refcount_protect` is the third (unused) pillar (see #510, which proposes consolidating `ProtectPool` + `Arena`). So every default user paid for `ahash` they never exercised.

## The change

```diff
- default = ["doc-lint", "refcount-fast-hash"]
+ default = ["doc-lint"]
```

The `std::collections::HashMap` backing already exists unconditionally (`HashMapArena = Arena<HashMap<usize, Entry>>`); the feature only *adds* `FastHashMap` / `FastHashMapArena`. The module compiles identically with and without the flag — this is purely a default-graph change, no fallback to write. Users who want the ahash variant opt in with `features = ["refcount-fast-hash"]`.

Also updated the lib.rs feature-table doc line (`enabled by default` → `opt-in`).

## Dependency-graph delta

`cargo tree -p miniextendr-api -e no-dev` (production graph, dev-deps excluded):

| | crates |
|---|---|
| with `refcount-fast-hash` (old default) | 18 |
| default after demotion | 11 |

7 crates leave the default production graph: **`ahash`, `zerocopy`, `version_check` (build-dep), `getrandom`, `once_cell`, `cfg-if`, `libc`**.

`cargo tree -p miniextendr-api -i ahash`:
- default after demotion → `error: package ID specification 'ahash' did not match any packages`
- `--features refcount-fast-hash` → `ahash v0.8.12 └── miniextendr-api`

(`ahash` remains in the root `Cargo.lock` because `miniextendr-bench` still declares the optional forwarding feature — the workspace lockfile records all reachable deps. No lockfile churn: root `Cargo.lock` and `rpkg/src/rust/Cargo.lock` are both unchanged.)

## CI / clippy coverage

Before this PR, the ahash path was linted via `clippy_default` (which runs default features). After demotion it would no longer be. The curated `clippy_all` list did **not** include `refcount-fast-hash`, so post-demotion **neither** clippy job would exercise the opt-in path — a silent coverage drop.

Fix: appended `refcount-fast-hash` to the explicit `clippy_all` feature list in `.github/workflows/ci.yml`, so the ahash `FastHashMap*` path stays linted as an opt-in.

## Sanity checks (from the issue)

1. **No by-default reliance** — confirmed. The flag is still in `default` on `main`; the only references are opt-in (bench feature, CLI bench command, feature-gated bench module) plus doc lines. The one production-side test uses `RefCountedArena` (BTreeMap), not a fast-hash type.
2. **Graph delta** — 18 → 11 production crates (above). A size build was not run (heavy release link); the dependency-graph delta stands in for it.
3. **clippy_all** — did not pass the flag; now does (above).

## Verification

- `cargo check -p miniextendr-api` (default, no ahash) → pass (exit 0).
- `cargo check -p miniextendr-api --features refcount-fast-hash` (ahash compiled in) → pass (exit 0).
- `cargo tree` before/after for the dependency delta (above).

Refs #510.

Closes #679

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -306,7 +306,7 @@ jobs:
         # explicit list so CI signal matches what rpkg actually enables.
         run: >-
           cargo clippy --workspace --all-targets --locked
-          --features rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,connections,nonapi,indicatif,strict-default,coerce-default,r6-default,worker-default,jiff
+          --features rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,connections,nonapi,indicatif,strict-default,coerce-default,r6-default,worker-default,jiff,refcount-fast-hash
           -- -D warnings 2>&1 | tee "$RUNNER_TEMP/clippy_all.log"
         shell: bash
 
--- a/miniextendr-api/Cargo.toml
+++ b/miniextendr-api/Cargo.toml
@@ -12,7 +12,7 @@ documentation = "https://docs.rs/miniextendr-api"
 rust-version = "1.85"
 
 [features]
-default = ["doc-lint", "refcount-fast-hash"]
+default = ["doc-lint"]
 ## Enable non-API R functions. These may break with R updates and cause
 ## `R CMD check` warnings. See NONAPI.md for details.
 nonapi = []
--- a/miniextendr-api/src/lib.rs
+++ b/miniextendr-api/src/lib.rs
@@ -207,7 +207,7 @@
 //! | `doc-lint` | Warn on roxygen doc comment mismatches (enabled by default) |
 //! | `macro-coverage` | Expose macro coverage test module for `cargo expand` auditing |
 //! | `growth-debug` | Track and report collection growth events (zero-cost when off) |
-//! | `refcount-fast-hash` | Use ahash for refcount arenas (enabled by default, not DOS-resistant) |
+//! | `refcount-fast-hash` | Use ahash for refcount arenas (opt-in, not DOS-resistant) |
 // Re-export linkme for use by generated code (distributed_slice entries)
 #[doc(hidden)]
 pub use linkme;


CGMossa:
Sure, I think this was done, because I thought, that I would make a Hash-backed mechanism for gc a default part of miniextendr. But I didn't,

PR #950, opened 2026-06-09
feat(macros): `#[miniextendr(error_direct)]` — C-side direct error raise (#665)

### LINKED ISSUE #665: perf: `#[miniextendr(error_direct)]` knob (C-side direct Rf_error, skip R re-raise)
## Context

Follow-up from commit e3279011. See \`analysis/scaffolding-perf-roadmap.md\` (option F) for the full evidence chain.

## What

A \`#[miniextendr(error_direct)]\` knob that, when a Rust panic / RCondition::Error is caught, calls \`Rf_error()\` directly from C with a pre-built class vector — skipping the R-side \`.miniextendr_raise_condition\` → \`stop(structure(...))\` step entirely.

## Why deprioritised (but worth tracking)

M4 (error-path attribution, \`miniextendr-bench/benches/error_path_attribution.rs\`) measured:

| Layer | min (ns) | What |
|---|---:|---|
| L1: \`make_rust_condition_value\` (build the tagged SEXP) | 125 | C-side build |
| L2: \`with_r_unwind_protect\` catching \`panic!()\` | **5833** | Rust panic + catch + build |
| L3: catching \`RCondition::Error\` | **6999** | the \`error!()\` macro path |

Cross-referenced with the R-side total (\`demo_error\` ≈ 21 μs vs \`stop()\` ≈ 7 μs, 14 μs delta):

- **~7 μs** = C-side panic catch + tagged SEXP build (this is what L2/L3 measure)
- **~7 μs** = R-side \`stop(structure(...))\` re-raise

\`error_direct\` would save the **R-side ~7 μs half** by directly calling \`Rf_error\` from C. But:

1. Custom class layering (\`rust_error\`, \`rust_warning\`, \`rust_message\`, \`rust_condition\`) would need to be built in C — currently it's an R-side \`switch (.val$kind)\` in \`.miniextendr_raise_condition\` (see \`miniextendr-api/src/registry.rs:738\`).
2. \`Rf_error\` only does errors. Warnings/messages would still need the existing path.
3. The bigger half of the cost (the Rust panic infrastructure itself — \`Box<dyn Any+Send>\` alloc, stack unwinding, \`panic_payload_to_string\`, panic-hook firing) is untouched. See sibling issue on typed-error transport.

## Why deprioritised

The error path isn't usually hot. M13 (Rprof over testthat) showed the wrapper layer at 1.1% of total test time — and the error path is a small fraction of that. For correct production code, error_direct shaves ~7 μs off the unhappy path — useful but not urgent.

Build cost is moderate (~4-8 hrs) because of the class-vector construction in C.

## When to revisit

If a user reports a workload that intentionally triggers Rust errors (e.g. tryCatch sweeps over expected-failure inputs, validation loops). At ~21 μs per error today, 100K such calls = 2.1 seconds. After \`error_direct\` that would drop to ~1 second.

## Related

- Roadmap: \`analysis/scaffolding-perf-roadmap.md\` (option F)
- Evidence: \`analysis/scaffolding-deep-findings-2026-05-20.md\`
- M4 bench: \`miniextendr-bench/benches/error_path_attribution.rs\`
- See also: typed-error transport (bigger redesign that would address the other ~7 μs half)

CGMossa:
Implements `#[miniextendr(error_direct)]` from #665: a fn-level knob that raises error-shaped failures directly from C via `Rf_eval(stop(structure(...)))`, skipping the R-side `.miniextendr_raise_condition` -> `stop(structure(...))` re-raise.

<details>
<summary>AI-written summary (Claude Opus 4.8)</summary>

## What shipped

The knob (not the fallback). When `error_direct` is set, the main-thread C wrapper routes through a new `with_r_unwind_protect_error_direct` (in `miniextendr-api/src/unwind_protect.rs`) instead of `with_r_unwind_protect`. On the `Err(payload)` arm it copies out the owned message/class, drops the panic payload, then calls the existing (proven, ALTREP-path) `raise_rust_condition_via_stop` for the error/panic kinds. Non-error signals (`warning!`/`message!`/`condition!`) fall back to the tagged-SEXP path, since `stop()` only does errors (point 2 of the issue).

Plus: fixture file `rpkg/src/rust/error_direct_fixtures.rs`, R test `rpkg/tests/testthat/test-error-direct.R`, macro docs, UI snapshot, regenerated wrappers/NAMESPACE/man/wasm_registry.

## Soundness — is the longjmp point Drop-free? Yes.

`stop()` longjmps without running Rust destructors, so the call must be reached with no live Rust `Drop` value. It is: by the time control reaches the `Err` arm, `run_r_unwind_protect`'s inner `catch_unwind` has **already unwound every Rust frame of the user closure** (files, `OwnedProtect` guards, `CString`s, `Box`es). The only live Rust value is the panic payload `Box<dyn Any + Send>`, which is `drop()`ed before the longjmp. `raise_rust_condition_via_stop` builds the condition object in R-protected SEXPs and holds no Rust `Drop` value across the `Rf_eval`; the protect stack is reclaimed by R's longjmp unwind. This is the **same pattern already shipping** on the ALTREP `RUnwind` guard path (`with_r_unwind_protect_or_raise`). It uses `Rf_eval(stop(...))`, not direct `Rf_error`, so MXL300 is satisfied.

## Class layering — preserved, and verified.

A `tryCatch` over an `error_direct` fixture sees exactly `c([custom,] "rust_error", "simpleError", "error", "condition")` — identical to the R-side `.miniextendr_raise_condition` `error`/`panic` arms. `test-error-direct.R` (21 assertions, **FAIL 0 / WARN 0**) covers:
- exact class vector + a direct-vs-indirect class-equality test (same fixture with and without the knob)
- `conditionMessage` round-trip (incl. empty + unicode + embedded newline)
- custom-class layering + `tryCatch` dispatch by both custom and `rust_error` classes
- `conditionCall` carries the wrapper call
- `warning!`/`message!` still signal correctly via the fallback
- a no-arg fixture (gctorture sweep target)

A `gctorture(TRUE)` pass over the error path (130 raises across all kinds) is clean — no GC corruption across the C-side condition build + `stop()` longjmp.

## Benefit — made measurable.

The issue cited `analysis/scaffolding-perf-roadmap.md`, `analysis/scaffolding-deep-findings-2026-05-20.md`, and M4 bench `miniextendr-bench/benches/error_path_attribution.rs` — **none of these exist on `main` or any git ref** (verified via `git log --all`; they're from an unmerged perf sprint), so the original 7/21/7 us figures were unverified. Instead of trusting them, this PR measures the real delta via an R microbenchmark comparing the same error through the direct vs. indirect path (3 runs, low variance):

| path | us/call |
|---|---:|
| `error_direct` (C-side `stop`) | ~17.6 |
| default (R-side re-raise) | ~22.4 |
| **delta** | **~4.8 us/call (~22%)** |

That validates the issue's "~7 us R-side half" claim (measured ~4.8 us on R 4.6 / arm64; the same order of magnitude).

## Scope cuts

- **fn-level only.** Impl-method paths and the `worker` / `rng` variants keep the tagged-SEXP transport. `rng + error_direct` must not skip `PutRNGstate()` via a longjmp through the RNG outer `catch_unwind`, so it deliberately stays on the tagged path; worker error handling is a separate mechanism. (Tracked as a follow-up.)
- **`Result::Err` / `Option::None` returns are unaffected** — those build tagged SEXPs that return *normally* from the closure (not via panic), so they reach the `Ok` arm. Only `panic!()`/`error!()` (panic-payload travel) are accelerated, which is exactly the issue's stated target.
- The C-built condition omits the `$kind` slot the R-side path attaches (`kind = "error"`). Documented in the test; `tryCatch` dispatch (by class), `conditionMessage`, and `conditionCall` are all unaffected.

## Verified

`cargo check`/`clippy -D warnings`/`fmt` on api+macros; `cargo test -p miniextendr-macros` (incl. regenerated UI snapshot); `just configure && just rcmdinstall && just force-document` clean (a roxygen `@title` link warning was fixed by rewording a doc comment); `just devtools-test error-direct` -> FAIL 0 PASS 21; gctorture clean; R microbench.

</details>

Closes #665


CGMossa:
I don't think this is a good idea actually. 
It would need a bunch of analysis and audits to ensure that this strategy is as good as the current condition-object signals.

PR #957, opened 2026-06-09
chore: remove .rust-llm-docs directory

CGMossa:
Accidentally pushed; not intended for the repository. Removes all 23 files under `.rust-llm-docs/`.

History rewrite can follow separately if desired (requires a force-push to `main`).


CGMossa:
Superseded — keeping the directory, just renaming and documenting it.

PR #988, opened 2026-06-11
test(worker): assert re-usability + bound longjmp-path leak (#931)

### LINKED ISSUE #931: Worker R-longjmp: assert worker re-usability + bound the ~8-byte leak (follow-up to #733)
Follow-up to #733 (closed by PR #930). PR #930 added a testthat fixture (`gc_stress_with_r_thread_stop()`) that characterises the **condition class** of an R error raised inside `with_r_thread` on a `run_on_worker` job: it surfaces as a bare `simpleError`, not the framework's tagged `rust_error` — because `R_ContinueUnwind` resumes the original `Rf_error` longjmp directly to R's top level, so `run_on_worker` never returns and the tagged-condition transport is bypassed.

Two sub-questions raised in #733 remain **unasserted** (the suite passing afterward is suggestive but not proof):

1. **Worker re-usability after an unwind.** Does the worker thread remain usable for subsequent `run_on_worker` jobs after an R-longjmp tears through a job mid-`with_r_thread`? Add a fixture that triggers the unwind and then dispatches a *second* normal worker job, asserting it succeeds (proves the worker re-armed its `recv()` loop and wasn't left in a poisoned state).

2. **The ~8-byte longjmp leak.** `with_r_unwind_protect_error_in_r` is documented to leak ~8 bytes (RErrorMarker + Box header) on the `R_ContinueUnwind` path (see `miniextendr-api/CLAUDE.md` gotchas / MXL300 rationale). Quantify/bound it — e.g. a loop of N unwinds under a memory check — or at least document it as expected and non-fatal with a reference.

Open design question (lower priority, may warrant its own issue): *should* a worker-job R error carry the `rust_error` tagged class rather than a bare `simpleError`? Today it can't, because the unwind resumes before the transport runs — changing that would require intercepting the longjmp on the worker side. Capturing here so the decision is tracked.

Refs #733. Overlaps the worker/shutdown coverage in #734 (#929).

CGMossa:
Assert worker thread re-arms after R-longjmp + bound the per-unwind memory leak.

Closes #931

<details><summary>AI-generated summary</summary>

- **New fixture + re-arm test**: `gc_stress_worker_roundtrip_after_longjmp()` (a `#[miniextendr(worker)] -> i32` fixture in `gc_stress_fixtures.rs`, gated on `worker-thread`) forces a fresh job onto the worker thread and asserts via `!is_r_main_thread()` that the worker's `recv()` loop re-armed. `test-worker-longjmp.R` exercises it directly after a single `gc_stress_with_r_thread_stop()` longjmp and again after a 5x loop of longjmps, proving the worker is re-usable.
- **Leak-bound test**: a second `test_that` block measures RSS growth across 2000 longjmp-through-worker unwinds (after a 2000-unwind warmup to settle allocator arenas / resident pages), and asserts growth stays under 8 KB/unwind. Steady-state measured ~1.8 KB/unwind, dominated by page/arena granularity rather than the much smaller heap leak. `skip_on_cran()` + `skip_on_os("windows")`.
- **Corrected docs**: the prior "`with_r_unwind_protect` leaks ~8 bytes" note was wrong. There are two distinct R-longjmp leak profiles, now documented in `unwind_protect.rs` (Leaks section), `worker.rs::dispatch_to_worker` (Leak note), and `miniextendr-api/CLAUDE.md`:
  - **Plain path** (body run inline): the leaked payload is an `RErrorMarker` ZST, so `Box::new` doesn't heap-allocate → effectively **~0 bytes**.
  - **Worker/dispatch path**: `R_ContinueUnwind` skips destructors for `worker_rx` / `response_tx` / the parked `Done(Err)` message / a `job_tx` refcount → **fixed-per-unwind, ~hundreds of bytes, linearly bounded, non-compounding**.
- **Stale fn name fixed** in the doc gotcha.

</details>


CGMossa:
Filed #989 to track the follow-up design question: worker-job R errors currently surface as bare `simpleError` rather than carrying the `rust_error` class. Tagging them requires `R_tryCatch` interception, widening `run_on_worker`'s error type, and `R_PreserveObject`+`Sendable` SEXP handoff — multi-day cross-cutting, tracked separately.

CGMossa:
Closing as duplicate of #993, which covers the same issue (#931) and has been rebased + had its wasm_registry content-hash regenerated. Please merge #993 instead.

Issue #938, opened 2026-06-09
r!() build-time R-grammar validation + Rf_lang* call-tree lowering (follow-up to #687)


CGMossa:
## Background

#687 shipped the tractable core of the R-eval macros:

- **`r_str!("...")`** — runtime string parse + eval. `R_ParseVector` →
  status check → `Rf_eval`, with every intermediate SEXP protected via
  `OwnedProtect`. Parse failures and eval failures both come back as
  `Err(String)` (never a longjmp / segfault). Backed by the new public
  `miniextendr_api::expression::r_eval_str(code, env) -> Result<SEXP, String>`.
- **`r!(...)`** — *static-string-lowering* variant. Takes R syntax as a Rust
  token stream, `stringify!`s it to a `&'static str` at build time, and reuses
  the same protect-safe `r_eval_str` path. The compile-time benefit it
  delivers **today** is only what the Rust front-end already enforces on the
  token stream: unbalanced delimiters and lexically invalid tokens fail at
  `cargo check`. The R source is still parsed at first eval.

This issue tracks the genuinely novel parts that were deferred from #687.

### 1. True build-time R-grammar validation

`r!` currently does *not* validate R **grammar** at build time — only Rust
tokenization. We want `r!(summarize(x +)` or `r!(if (x) )` to fail at
`cargo check`, not at the user's first runtime call.

Options (from the #687 sketch):
- Embed a Rust-side R-grammar parser (subset) in the proc macro.
- Invoke the host R's `R_ParseVector` from `build.rs` when an R is on PATH
  (graceful fallback to no-check when absent — must not become a hard build
  dependency; CRAN/offline farms have no guaranteed R-with-parser contract
  beyond the build itself).
- Promote `r!` from `macro_rules!` to a real proc macro (required for any of
  the above — it has to walk the token tree and re-emit).

### 2. Direct `Rf_lang*` call-tree lowering

For the call-shaped subset (`f(a, b, ...)`, nested), lower `r!` straight to a
pre-built `Rf_lang*` / `Rf_cons` call tree with `Rf_install`ed symbols and
auto-inserted protects — **no runtime R parser cost at all**. This is the
"cheaper at runtime" branch from the #687 API sketch. Reuse the existing
`expression::RCall` builder as the lowering target where possible.

### 3. (lower priority) Interpolation

`r!(summary(${x}))` to splice a Rust `SEXP` into the call tree. Called out as
[...]

PR #1014, opened 2026-06-12
feat(macros): promote r! to a proc-macro with conservative R-grammar validation

CGMossa:
## Summary

- Promotes the `r!` macro from a `macro_rules!` (in `miniextendr-api`) to a proc-macro (in `miniextendr-macros`), re-exported via `miniextendr-api`'s `pub use`.
- Expansion is byte-identical at runtime to the old `macro_rules!`: the proc-macro emits `::core::stringify!(#tail_tokens)` — the same built-in that `macro_rules!` used, so whitespace normalisation is identical.
- Adds a conservative R-grammar validator (`miniextendr-macros/src/r_macro/grammar.rs`) that rejects only known-bad constructs; accepts anything ambiguous (reject-only-known-bad, no false positives):
  - Expression ending with a binary operator (e.g. `x *`)
  - Consecutive binary operators (e.g. `* *`)
  - Empty function-call arguments in non-index context (e.g. `f(, 1)`)
  - `if`/`while` with empty condition or missing body
  - `for` loop missing `in`
- Adds four UI (compile-fail) tests and one compile-pass test covering the above rules.

Addresses item 1 of #938. Items 2+3 (Rf_lang* call-tree lowering) remain open.

## What was not changed

- `rpkg/R/miniextendr-wrappers.R`, `NAMESPACE`, `man/*.Rd` — the `r!` macro does not appear in the R surface; no wrapper regeneration required.
- `registry.rs` — no new `#[distributed_slice]` entries.

## Notes on the validator

Single-quoted strings and backtick-quoted names can't reach the validator (Rust's lexer rejects them before macro expansion). `%op%` custom operators are accepted without analysis — they tokenise as `%` + `op` + `%` and would need their own special-casing; not implemented (conservative choice).

## Test plan

- [x] `just check` — exit 0
- [x] `cargo test -p miniextendr-macros` — 381 passed
- [x] `just test` — all suites pass (includes `r_eval_suite` with 4 `r!()` calls)
- [x] `just fmt` — clean
- [x] `just clippy` — clean
- [x] `cargo clippy --workspace --all-targets --locked --features full-codegen,...` (clippy_all subset from ci.yml) — clean
- [x] `just lint` — clean
- [x] `TRYBUILD=overwrite cargo test -p miniextendr-macros --test ui` — all UI snapshots match
- [ ] `just rcmdinstall` — not runnable in this worktree (rv library uninitialised); no R-surface change expected

Refs #938

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/miniextendr-api/src/lib.rs
+++ b/miniextendr-api/src/lib.rs
@@ -576,13 +576,25 @@ macro_rules! r_str {
 /// source is lowered to a `&'static str` (`stringify!`), so there is no
 /// `format!` allocation at the call site.
 ///
+/// This proc-macro additionally validates a conservative subset of known-bad
+/// R syntax constructs (trailing binary operators, empty function call
+/// arguments, bare `if`/`while`/`for` without a body, etc.) and emits a
+/// precise compile error pointing at the offending token.
+///
 /// # What is deferred
 ///
-/// True build-time R-grammar validation (embedding R's parser / invoking
-/// `R_ParseVector` from `build.rs`) and direct `Rf_lang*` call-tree lowering
-/// (skipping the runtime parser entirely) are tracked as a follow-up — see the
-/// issue referenced in the crate docs. Until then `r!` parses its static
-/// string at first evaluation, exactly like `r_str!`.
+/// Direct `Rf_lang*` call-tree lowering (skipping the runtime parser entirely)
+/// is tracked as a follow-up — see issue #938 (item 2). Until then `r!` parses
+/// its static string at first evaluation, exactly like `r_str!`.
+///
+/// # Non-goals
+///
+/// A complete R grammar validator is not achievable over Rust tokens:
+/// - Single-quoted strings (`'hello'`) and backtick-quoted names (`` `foo` ``)
+///   already die at the Rust lexer — nothing to validate.
+/// - `%op%` tokenises as `%`, ident, `%` and is accepted without analysis.
+/// - Anything the validator cannot confidently classify as wrong passes through
+///   unvalidated (conservative reject-only-known-bad design).
 ///
 /// # Forms
 ///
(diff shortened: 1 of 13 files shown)


CGMossa:

<details>

## Summary
- Review pass found the validator rejecting valid R: empty (missing) call arguments are legal in every call form, not just `[` indexing. `f(, x)`, `f(x,,y)`, and `matrix(, 2, 2)` all parse (verified against R 4.6). R's `sublist` grammar production is shared between `(`-calls and `[`-indexing, so the `IndexGroup` carve-out was built on a wrong premise. The rule, its `SequenceContext` plumbing, and the `r_macro_empty_call_arg` UI test are removed in 09541ab3.
- Second false positive: `r!(env::foo())` (R namespace access on a package named `env`) was misparsed as the `env: <expr>;` head, producing a spurious "missing `;` separator" error. The old `macro_rules!` fell through to the code arm here. Fixed by requiring the `:` after `env` to be Alone-spaced (a Joint `:` is half of `::`).
- One accidental correctness: the Joint-spacing skip in the consecutive-operator check means `2 ** 3` is accepted, and that is right, R's parser takes `**` as an undocumented synonym for `^`. The comment claimed the opposite ("clearly invalid R"); prose corrected and the behavior locked in with a test.

## Notes
- The lesson for future validator rules: every "known-bad" claim needs a `parse(text=...)` check against real R first. Two of the five shipped rules were wrong about R's grammar, and both were in the reject direction, which is exactly what the reject-only-known-bad design promises never to do.
- The remaining rules (trailing binary op, `* *` doubles, empty `if`/`while` condition, `for` without `in`, missing bodies) were all verified as genuine parse errors.

---

</details>

PR #1015, opened 2026-06-12
feat(macros): lower call-shaped r!() subset to Rf_lang* via RCall, string fallback otherwise

### LINKED ISSUE #938: r!() build-time R-grammar validation + Rf_lang* call-tree lowering (follow-up to #687)

CGMossa:
## Summary

Implements `r!()` call-tree lowering (issue #938 item 2), stacked on #1014 (G7-PR2 grammar validation).

After grammar validation in PR-2, `expand_inner` now classifies the tail before falling back to `r_eval_str`. Lowering is attempted for the call-shaped subset; everything else gets the existing string path — no accepted input can start failing.

**Lowerable subset**:
- Simple calls: `ident(args)` and R dot-names (`is.null(NULL)`, `as.character(x)`)  
- Namespaced calls: `pkg::fn(args)` via new `RCall::namespaced()` method  
- Atoms: string/integer/float literals, `TRUE`/`FALSE`/`NULL`/`NA`/`NA_integer_`/`NA_real_`/`NA_character_`/`Inf`/`NaN`, bare identifiers (symbols), named args (`name = atom`)  
- Nested lowerable calls as arguments (`c(1L, c(2L, 3L))`)  
- Anything outside this set (operators, assignments, control flow, `[`/`[[`/`@`, etc.) → falls back to `r_eval_str`

**What's deferred** (filed as #1019):
- Token-interpolation in the call-shaped path (`r!(f(x))` where `x` is a Rust variable)
- `NA_complex_` lowering (complex type not in `RCall` API yet)

## Changes

- `miniextendr-macros/src/r_macro/lowering.rs` (new): classifier (`LowerCall`/`LowerAtom` AST) + emitter; 22 unit tests for all classifier paths including `42L` single-token detection and R dot-name handling
- `miniextendr-api/src/expression.rs`: `RCall::namespaced(pkg, fun)` resolves `pkg::fun` at runtime via `R_tryEvalSilent`
- `miniextendr-api/tests/r_eval.rs`: `r_lowering_suite` with 10 equivalence tests (lowered vs string-path produce `R_compute_identical` results)
- `rpkg/src/rust/gc_stress_fixtures.rs`: `gc_stress_r_macro_lowering()` no-arg GC-stress fixture (convention #430)
- Regenerated: `miniextendr-wrappers.R`, `NAMESPACE`, `man/gc_stress_fixtures.Rd`, `wasm_registry.rs`

## Non-obvious implementation details

- `42L` is a **single** `proc_macro2` token (`Literal { lit: 42L }`), not `42` + `L`. Detection via `s.ends_with('L')` + strip + `parse::<i64>()`.
- `is.null` tokenizes as `[Ident, Punct('.'), Ident]` — `try_parse_r_dotname()` reconstructs the full name.
- `?` propagation from `RCall::namespaced()?` inside `unsafe {}` is resolved by wrapping the entire expansion in an IIFE closure `(|| -> Result<SEXP, String> { unsafe { ... } })()`.

## Test plan

- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (clippy_default) passes
- [x] `cargo clippy --workspace --all-targets --locked --features full-codegen,rayon,...` (clippy_all) passes
- [x] 935 workspace tests pass, including 2 new integration test suites (`r_eval_suite`, `r_lowering_suite`)
- [x] 22 new unit tests in `lowering.rs` for classifier edge cases
- [x] `just force-document` zero roxygen2 warnings; `NAMESPACE` + `wasm_registry.rs` regenerated
- [x] `cargo fmt --all --check` passes

Closes #938 item 2. Stacks on #1014.


CGMossa:

<details>

## Summary
- Rebased onto the updated #1014 base (which removed the empty-call-arg rule) and review-fixed the lowering path in d8791361. Two semantic divergences from the string path were live, both in the "silently produces a different value" direction the PR's equivalence promise forbids:
  - `r!(identity(NA))` lowered bare `NA` as `NA_integer_`. In R, `typeof(NA)` is `"logical"`. `identical(NA, NA_integer_)` is FALSE, so lowered and string paths disagreed. Now emits `scalar_logical_raw(NA_LOGICAL)`.
  - `r!(identity(42))` lowered unsuffixed integers as INTSXP. R parses unsuffixed numeric literals as double, only `42L` is integer. Now lowers as double.
- Neither case was in the 10-test equivalence suite (it used `42L`, `TRUE`, `NULL`, strings). Both are now locked in as lowered-vs-string `R_compute_identical` tests.
- Also guarded: `pkg : : fn(x)` (Alone-spaced colons, an R parse error) no longer classifies as a namespaced call, it falls back so the string path rejects it at runtime. And the post-#1014 interaction is pinned by a test: `matrix(, 2, 2)` now reaches the classifier, where the empty segment correctly fails `classify_atom` and falls back rather than lowering with the slot dropped.
- Cleanup: the duplicated NA_real_ bit-pattern const is gone, the emitter references `altrep_traits::NA_REAL` directly.

## Notes
- Same lesson as the #1014 comment, from the other direction: the validator's false positives rejected valid R, and the lowering's false "knowledge" changed R's literal semantics. Every claim about what R parses or types needs a check against a live R session, not intuition.
- No CI runs on this PR yet because it targets `feat/r-macro-proc`, checks will fire once it retargets to main after #1014 merges.

---

</details>

CGMossa:

<details>

Sync Checks failed on the rebased head with committed Rd drift: `gc_stress_r_macro_lowering`'s export was in NAMESPACE but `man/gc_stress_fixtures.Rd` was never regenerated to include it. This predates the rebase — CI simply never ran on this branch before (the PR targets `feat/r-macro-proc`, and checks only started reporting after the recent pushes). Regenerated via the full loop (`rcmdinstall` + `force-document`) and pushed as abb3ef23; the diff is exactly the four lines the sync check printed.

</details>

CGMossa:

Merged main after #1057 (name-as-title) landed there. The only conflict was `man/gc_stress_fixtures.Rd`: this branch's committed copy predated #1057, so its description block still had the degenerate repeated placeholder lines. Resolved to regeneration-truth by taking main's full-description side and inserting `gc_stress_r_macro_lowering`'s rustdoc paragraphs at its registration position (between `gc_stress_map_to_dataframe` and `gc_stress_worker_roundtrip`). Sync Checks will verify the hand-resolution against the actual regen.

</details>

Issue #1032, opened 2026-06-13
build: extend the leaked-tarball configure guard to the monorepo template


CGMossa:
## Background

PR for #1029 added a leaked-vendor-tarball guard to `rpkg/configure.ac` and the
standalone `minirextendr/inst/templates/rpkg/configure.ac`. The guard keys on the
existing `SOURCE_IS_GIT` walk: if `inst/vendor.tar.xz` is present AND
`SOURCE_IS_GIT=true` AND `MINIEXTENDR_BOOTSTRAP` is unset, configure errors out
instead of silently flipping into tarball mode.

## Gap

`minirextendr/inst/templates/monorepo/rpkg/configure.ac` is the *simpler*
standalone variant and **does not compute `SOURCE_IS_GIT`** (no `.git` walk, no
auto-vendor block). So the guard could not be ported there without first
introducing the `SOURCE_IS_GIT` machinery — a larger structural change than the
targeted #1029 fix warranted.

The monorepo template's `bootstrap.R` already sets `MINIEXTENDR_BOOTSTRAP=1` on
its configure invocation (added in the #1029 PR for uniformity), so the bypass
signal is in place; only the guard itself (and the `SOURCE_IS_GIT` walk it
depends on) is missing.

## Work

1. Add the `SOURCE_IS_GIT` `.git`-ancestor walk to
   `minirextendr/inst/templates/monorepo/rpkg/configure.ac` (mirror the
   standalone `rpkg/` template).
2. Add the same leaked-tarball guard block (scaffold-appropriate wording,
   no `just` references).
3. `just templates-approve` to refresh `patches/templates.patch`; confirm with
   `just templates-check`.

m4 gotcha learned in #1029: keep the `AC_MSG_ERROR([...])` message body free of
`#` characters and `()` empty-parens-adjacent-to-`(` — they break macro
expansion (`AS_MESSAGE_LOG_FD` undefined). Spell out "issue NNNN", not "#NNNN".

PR #1033, opened 2026-06-13
fix(build): guard bare configure / R CMD INSTALL against a leaked vendor tarball (#1029)

### LINKED ISSUE #1029: build: guard bare R CMD INSTALL against a leaked vendor tarball (T5.6/T5.7)
The `just` recipes and `miniextendr_doctor()` both catch a leaked `inst/vendor.tar.xz` in a dev tree, but a bare `R CMD INSTALL rpkg/` from the shell doesn't — `configure` just sees the tarball and silently flips to tarball mode, building against vendored sources and ignoring local workspace edits. `configure` already knows it's in a git tree (`SOURCE_IS_GIT`); it should use that to fail loud.

<details>

## Summary
- Source: `analysis/build-system-tests-2026-05-11.md` §T5.6/T5.7 ("§15 follow-up"): T5.6 (bare `R CMD INSTALL rpkg/`) "currently silently goes into tarball mode → BROKEN"; T5.7 (`miniextendr_build()`) "hard fail (R-side guard — needs adding to match `just`)".
- Current coverage: `justfile` `_assert-no-vendor-leak` guards `rcmdinstall`/`devtools-*`; `minirextendr_doctor()` (`minirextendr/R/doctor.R:213`) *detects* the leak; `miniextendr_clean_vendor_leak()` cleans it; `workflow.R:289` clears the latch before `miniextendr_build()`. The **bare-shell `R CMD INSTALL` path has no guard** — `configure` treats tarball-presence as the sole mode signal.

## The hook already exists
- `rpkg/configure.ac` computes `SOURCE_IS_GIT` and already uses it to **skip auto-vendor in dev trees** (`test "$SOURCE_IS_GIT" = "false"` gate, configure.ac:80-82). A symmetric guard is natural: `inst/vendor.tar.xz` present **and** `SOURCE_IS_GIT = true` ⇒ leaked tarball in a working tree ⇒ error out pointing at `just clean-vendor-leak` / `miniextendr_clean_vendor_leak()`, instead of silently entering tarball mode.

## Open design question
- `just vendor` deliberately produces the tarball **inside** the git tree as an intermediate for `R CMD build`/`check` (those recipes trap-clean it on exit, and the staged build dir consumed by `R CMD build` has no `.git` ancestor). A blanket "git + tarball = error" at configure time must not break that produce-then-consume flow. The guard likely needs to fire only on a *direct* `configure`/`R CMD INSTALL` against the source tree, not on the recipe-staged path — scope carefully.

## Test plan
- [ ] Add §T5.6/T5.7 coverage as `expect_failure` tests (per the doc, they land red until the guard exists, then flip green).
- [ ] Verify `just vendor` → `just r-cmd-build`/`r-cmd-check` still succeed (the guard must not trip the legitimate produce-then-consume path).

---

</details>

CGMossa:
Closes #1029.

A leftover `rpkg/inst/vendor.tar.xz` in a git-tracked source tree silently flipped `configure` into tarball mode — building against frozen vendored sources and ignoring local workspace edits. The `just` consume-recipes (`_assert-no-vendor-leak`) and `miniextendr_doctor()` already catch this, but a bare `bash ./configure` / `R CMD INSTALL rpkg/` from the shell had no guard. This PR adds that guard, reusing the existing `SOURCE_IS_GIT` signal as the issue suggested.

<details>
<summary>AI-generated details (reviewed by me)</summary>

## Guard condition and message

In `rpkg/configure.ac`, right after the `SOURCE_IS_GIT` `.git`-ancestor walk:

```sh
if test -f "$abs_top_srcdir/inst/vendor.tar.xz" \
   && test "$SOURCE_IS_GIT" = "true" \
   && test -z "${MINIEXTENDR_BOOTSTRAP+x}"; then
  AC_MSG_ERROR([leaked vendor tarball detected in a git-tracked source tree.
  ...
  Remove it before building from source:
    just clean-vendor-leak
  or:
    rm -f inst/vendor.tar.xz
  In a scaffolded package, run miniextendr_clean_vendor_leak, or
  minirextendr::miniextendr_doctor to diagnose. ...])
fi
```

Tarball present **AND** `SOURCE_IS_GIT=true` **AND** the bootstrap-bypass env var unset ⇒ hard error pointing at `just clean-vendor-leak` / `rm -f inst/vendor.tar.xz` / `miniextendr_doctor`.

## Why produce-then-consume is provably unaffected

The risk is the recipes that deliberately produce the tarball inside the git tree then build. Traced each:

* **`just r-cmd-build` / `r-cmd-check`** use raw `R CMD build`. Verified empirically that raw `R CMD build` does **not** honour `Config/build/bootstrap` (that is a pkgbuild-only convention) — so it never re-runs `./configure` in the git source tree. `R CMD build` stages to a temp dir with no `.git`, so any configure run there has `SOURCE_IS_GIT=false` and the guard cannot fire.
* **`just devtools-build` / `devtools-check`** go through pkgbuild, which **does** run `bootstrap.R` in the *original* git source dir (verified empirically: `getwd()` is the source dir, `.git` present) **after** `just vendor` has produced the tarball. To keep the guard silent on this legitimate path, `bootstrap.R` now exports `MINIEXTENDR_BOOTSTRAP=1` on its `./configure` invocation (`system2(..., env = "MINIEXTENDR_BOOTSTRAP=1")`). The guard skips when that var is set.
* **End-user tarball installs** extract to a dir with no `.git` (`SOURCE_IS_GIT=false`), so the guard never fires for them.

So the only path that trips the guard is a *direct* `bash ./configure` / `R CMD INSTALL rpkg/` against the git source tree with a leftover tarball — exactly the leak case.

## Scope / templates

* Ported the guard + bootstrap signal into the standalone `minirextendr/inst/templates/rpkg/` template (which mirrors rpkg's `SOURCE_IS_GIT` structure).
* The `monorepo/rpkg/` template's `configure.ac` lacks the `SOURCE_IS_GIT` machinery entirely, so adding the guard there is a larger change — tracked as #1032. Its `bootstrap.R` already carries the bypass signal for forward-compat.
* Regenerated `rpkg/configure` via `autoconf` and `patches/templates.patch` via `just templates-approve`.

## m4 gotcha

`AC_MSG_ERROR` message bodies must avoid `#` characters and `()`-empty-parens immediately followed by `(` — both broke macro expansion (`AS_MESSAGE_LOG_FD undefined`). Spelled out "issue 1029" instead of "#1029".

## Verification

| # | Check | Result |
|---|---|---|
| 1 | `cd rpkg && autoconf` regenerates cleanly; configure diff additive (20 lines) | PASS |
| 2 | Leak case: dummy `inst/vendor.tar.xz` in git tree → `bash ./configure` exits 1 with the loud error pointing at `just clean-vendor-leak` | PASS |
| 2b | Bootstrap bypass: `MINIEXTENDR_BOOTSTRAP=1 bash ./configure` with tarball present → guard does NOT fire (0 matches), proceeds to tarball mode | PASS |
| 3 | Source mode: no tarball → `bash ./configure` succeeds, selects source/monorepo mode, writes source-mode `.cargo/config.toml` | PASS |
| 4 | Produce-then-consume: `just test-bootstrap-vendor` (bootstrap-produces-vendor + vendor-loud-fail + vendor-cross-surface-rename) all pass; bootstrap runs configure with tarball absent then vendors — no spurious guard | PASS |
| 5 | `just templates-check` passes after `templates-approve` | PASS |

Standalone template `configure.ac` also `autoconf`s cleanly (exit 0).
</details>

### DIFF
--- a/minirextendr/inst/templates/monorepo/rpkg/bootstrap.R
+++ b/minirextendr/inst/templates/monorepo/rpkg/bootstrap.R
@@ -8,12 +8,16 @@
 # `R CMD build`), configure builds in tarball/offline mode. Otherwise,
 # source/network mode is used.
 
+# MINIEXTENDR_BOOTSTRAP=1 tells configure's leaked-tarball guard (#1029) that
+# this ./configure was invoked by bootstrap, not directly, so a deliberate
+# tarball-in-a-git-tree (produced by the vendor step that precedes
+# devtools::build()/check()) is not mistaken for a leak.
 if (.Platform$OS.type == "windows") {
   if (file.exists("configure.ucrt")) {
-    system("sh configure.ucrt")
+    system2("sh", "configure.ucrt", env = "MINIEXTENDR_BOOTSTRAP=1")
   } else if (file.exists("configure.win")) {
-    system("sh configure.win")
+    system2("sh", "configure.win", env = "MINIEXTENDR_BOOTSTRAP=1")
   }
 } else {
-  system2("bash", "./configure")
+  system2("bash", "./configure", env = "MINIEXTENDR_BOOTSTRAP=1")
 }
--- a/minirextendr/inst/templates/rpkg/bootstrap.R
+++ b/minirextendr/inst/templates/rpkg/bootstrap.R
@@ -37,14 +37,21 @@
 # pkgbuild-only). The bundled inst/vendor.tar.xz from step 2 is what
 # configure detects to build offline.
 
+# MINIEXTENDR_BOOTSTRAP=1 tells configure's leaked-tarball guard (#1029) that
+# this ./configure was invoked by bootstrap, not directly. The vendor step that
+# precedes devtools::build()/check() leaves inst/vendor.tar.xz in the
+# (git-tracked) source dir on purpose; without this signal configure would
+# (correctly, for a *direct* invocation) treat that tarball-in-a-git-tree as a
+# leak and abort. We pass it inline to the configure call only, so it never
+# leaks into the cargo-revendor step or the surrounding R session.
 if (.Platform$OS.type == "windows") {
   if (file.exists("configure.ucrt")) {
-    system("sh configure.ucrt")
+    system2("sh", "configure.ucrt", env = "MINIEXTENDR_BOOTSTRAP=1")
   } else if (file.exists("configure.win")) {
-    system("sh configure.win")
+    system2("sh", "configure.win", env = "MINIEXTENDR_BOOTSTRAP=1")
   }
 } else {
-  system2("bash", "./configure")
+  system2("bash", "./configure", env = "MINIEXTENDR_BOOTSTRAP=1")
 }
 
 if (!file.exists("inst/vendor.tar.xz")) {
--- a/minirextendr/inst/templates/rpkg/configure.ac
+++ b/minirextendr/inst/templates/rpkg/configure.ac
@@ -77,6 +77,33 @@ while test -n "$_walk_dir" && test "$_walk_dir" != "/"; do
   _walk_dir=`dirname "$_walk_dir"`
 done
 
+dnl ---- leaked-tarball guard (#1029) ----
+dnl A vendor tarball sitting in a git-tracked source tree is almost always a
+dnl LEAK: a previous build that produced inst/vendor.tar.xz was interrupted
+dnl before its cleanup ran. configure would silently flip to tarball mode and
+dnl build against the frozen vendored sources, ignoring local source edits.
+dnl Fail loud instead. bootstrap.R exports MINIEXTENDR_BOOTSTRAP=1 when it
+dnl invokes configure (after deliberately producing the tarball for the build),
+dnl so this guard stays silent on that legitimate produce-then-consume path.
+dnl End-user tarball installs extract to a dir with no .git, so the guard never
+dnl fires for them (SOURCE_IS_GIT=false).
+if test -f "$abs_top_srcdir/inst/vendor.tar.xz" \
+   && test "$SOURCE_IS_GIT" = "true" \
+   && test -z "${MINIEXTENDR_BOOTSTRAP+x}"; then
+  AC_MSG_ERROR([leaked vendor tarball detected in a git-tracked source tree.
+
+  inst/vendor.tar.xz is present AND this is a git working tree, so configure
+  would silently enter tarball mode and build against frozen vendored sources,
+  ignoring any local source edits. This is almost always a leftover from an
+  interrupted build.
+
+  Remove it before building from source:
+    rm -f inst/vendor.tar.xz
+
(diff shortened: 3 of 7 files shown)


CGMossa:
Pushed `eac35d9f` to fix the red CI on this PR.

**Root cause.** The leaked-tarball guard misfired on every R CMD check job. CI runs `just vendor` (writes `rpkg/inst/vendor.tar.xz`), `R CMD build` bundles it into the tarball, and `R CMD check` extracts that tarball into `<pkg>.Rcheck/<pkg>/` — which lives **inside the repo working tree**, so the `SOURCE_IS_GIT` ancestor-walk found the repo `.git` and returned `true`. The guard then treated a legitimate built-tarball install as a source-tree leak and aborted with `configure: error: leaked vendor tarball detected in a git-tracked source tree`. The `MINIEXTENDR_BOOTSTRAP` bypass only covered the pkgbuild/`bootstrap.R` path, not raw `R CMD build` + `R CMD check` (`r-lib/actions/check-r-package`) nor CRAN-style checks.

**Fix.** Added a `SOURCE_IS_BUILT` discriminator. `R CMD build` stamps an unconditional `Packaged:` field into the sealed DESCRIPTION (`R CMD INSTALL` adds `Built:`); a checked-out source tree has neither. The guard now fires only when `inst/vendor.tar.xz` present **AND** `SOURCE_IS_GIT=true` **AND** `SOURCE_IS_BUILT=false` **AND** the bootstrap bypass is unset. This separates a genuine leak (`R CMD INSTALL rpkg/` from the working tree → no `Packaged:` → fires) from a legitimate built-tarball install extracted under a git repo (`Packaged:` present → suppressed).

The `MINIEXTENDR_BOOTSTRAP` bypass is **retained**: `just devtools-build` runs `just vendor` (tarball present), then `bootstrap.R` runs `./configure` in the original, not-yet-built source dir where `Packaged:` is absent — so `SOURCE_IS_BUILT` can't suppress it and the env-var bypass is still required there.

**Verified locally (sandbox-disabled):**
- `autoconf` clean; `configure` diff is the 6-line shell addition only.
- Genuine leak still caught: dummy `inst/vendor.tar.xz` in the source tree → `configure` errors. ✓
- Source mode (no tarball) → exit 0, source install. ✓
- **CI path reproduced faithfully**: `just r-cmd-build` produced the real 22 MB tarball (DESCRIPTION has `Packaged:`, ships `inst/vendor.tar.xz`), extracted it under the repo `.git`, ran `bash ./configure` in the extracted dir → guard silent, entered tarball mode, unpacked vendor, lock-shape OK, exit 0. Inverse control (strip `Packaged:`) → guard fires. ✓
- `just test-bootstrap-vendor` → all 3 scripts pass (bootstrap bypass path intact). ✓
- `just templates-check` → passes (template configure.ac updated, `patches/templates.patch` re-approved). ✓

Issue #346, opened 2026-05-02
Structured `data = list(...)` payloads on `error!()` / `condition!()`


CGMossa:
## Context

PR #344 added `error!()` / `warning!()` / `message!()` / `condition!()` macros with optional `class = "..."` extension for programmatic catching:

```rust
error!(class = "my_error", "missing field: {name}");
```

```r
tryCatch(fn(), my_error = function(e) e\$message)
```

The [`rlang::abort`](https://rlang.r-lib.org/reference/abort.html) family additionally supports a `data` argument that attaches a named list to the condition, accessible to handlers as `e\$<name>`:

```r
tryCatch(
  rlang::abort(\"oh no\", class = \"my_error\", data = list(value = 42, code = 7L)),
  my_error = function(e) c(e\$value, e\$code)
)
# [1] 42  7
```

This enables programmatic recovery: a handler can inspect typed fields rather than parsing the message string.

## Proposal

Extend the macros with an optional `data = ...` form that takes a Rust value implementing `IntoR` (or specifically `HashMap`/`Vec<(&str, T)>` for named-list construction):

```rust
// Single named field
error!(class = \"missing_field\", data = (\"name\", name), \"field missing\");

// Multiple fields
error!(
    class = \"validation_error\",
    data = vec![(\"value\", value), (\"min\", 0), (\"max\", 100)],
    \"value {value} out of range\"
);
```

The extra fields are spliced into the `structure(list(...), class = ...)` call alongside `message` / `call` / `kind`.

## Implementation sketch

- `RCondition` enum gains a `data: Option<Vec<(String, ???)>>` field. The value type needs to be Send (panic_any payload) and IntoR-compatible. Could be `SexpHandle` (a Send-safe wrapper) created on the calling thread before `panic_any`, or just a serialized form like JSON if simpler.
- `make_rust_condition_value` accepts the data list and writes it as a 5th element of the tagged SEXP (`error`, `kind`, `class`, `call`, `data`).
- `condition_switch_lines` reads `.val\$data` and splices its named fields into the `list(...)` inside `structure()`.

## Open questions

- **Send-safety.** `SEXP` is not `Send`, so a `panic_any` payload can't carry SEXPs across the worker→main thread boundary. Either (a) restrict `data = ...` to Rust-native types (numbers, strings, bools, vectors) and convert at the unwind path, or (b) require the macro to run on the main thread and document the worker-thread limitation.
[...]

PR #1044, opened 2026-06-14
feat(condition): richer data payload value types + keyed builder sugar (#995)

### LINKED ISSUE #995: Condition data payloads: richer value types and keyed builder sugar (follow-up to #346)
## Context

#346 (PR pending) shipped structured `data = ...` payloads on `error!()` / `warning!()` / `message!()` / `condition!()`. The v1 value set is intentionally narrow: scalars and `Vec`s of `i32` / `f64` / `bool` / `String` (plus `&str` / `Vec<&str>` converted to owned), carried through `panic_any` as the Send-safe `ConditionDataValue` enum and materialised on the R main thread in `make_rust_condition_value_with_data`.

## What's deferred

1. **Richer value types.** Candidates, roughly in order of expected demand:
   - `Option<T>` of the supported scalars → `NA` (the natural R idiom for "field present but missing")
   - `Vec<Option<T>>` → vectors with `NA` elements
   - `i64` / `u32` etc. via the existing strict/coerce conversion ladder (mind `i32::MIN` = `NA_integer_`)
   - nested named lists (`HashMap<String, ConditionDataValue>` → R list)
   - a `Debug`-stringify fallback variant so any `T: Debug` can ride along explicitly (`ConditionDataValue::debug(x)`)

2. **Builder sugar.** The issue sketch floated `error_data!(value = 42, code = 7)`-style keyed construction. Today the grammar is `data = ("name", value)` / `data = [("a", v1), ("b", v2)]`. A keyed form `data = { value = 42, code = 7 }` would read better and could be added backward-compatibly as a third `__mx_condition_data!` arm.

3. **Runtime-built payloads.** The macros only accept literal tuple/list forms. Power users constructing `ConditionData` (`Vec<(String, ConditionDataValue)>`) at runtime must `panic_any(RCondition::Error { ... })` by hand (same situation as runtime `class` strings — see `demo_error_custom_class`). A `data = @expr` escape hatch (or just documenting the enum-literal pattern) would close this.

## Constraint to preserve

Whatever the sugar, the payload must stay `Send` (panic_any transport; may cross worker→main). Arbitrary `IntoR` values cannot be supported without materialising off the main thread — that's a hard no. The enum + `From` impls is the pattern; extend it, don't replace it.

## Pointers

- `miniextendr-api/src/condition.rs` — `ConditionDataValue`, `__mx_condition_data!`, macro arms
- `miniextendr-api/src/error_value.rs` — `make_rust_condition_value_with_data` (PROTECT discipline notes)
- `rpkg/src/rust/condition_demo.rs` + `rpkg/tests/testthat/test-conditions.R` — fixture/test pattern to extend
- `rpkg/src/rust/gc_stress_fixtures.rs::gc_stress_condition_data` — add new value types here too

CGMossa:
Follow-up to #346 (merged). Extends the structured `data = ...` condition payload with richer value types and a keyed builder grammar, keeping the payload `Send` (panic_any transport, may cross worker→main) — the enum is extended, not replaced.

Everything in the issue landed; nothing deferred, so no follow-up issue.

<details>
<summary>AI-generated detail</summary>

## What landed

**Richer `ConditionDataValue` variants** (all materialise on the R main thread via existing `IntoR` impls):
- NA-aware scalars: `Option<i32>` / `Option<f64>` / `Option<bool>` / `Option<String>` / `Option<&str>` → R `NA` ("field present but missing").
- NA-aware vectors: `Vec<Option<i32>>` / `…<f64>` / `…<bool>` / `…<String>` → vectors with per-element `NA`.
- Wide-integer ladder: `i64` (smart INTSXP/REALSXP, `i32::MIN` excluded as `NA_integer_`) and `u32` (lossless widening to `i64`).
- Nested named lists: `Vec<(String, ConditionDataValue)>` → R `list()`, recursive. `list_into_sexp` carries its own PROTECT scope mirroring the top-level builder (protect VECSXP, protect names STRSXP, root each child before the next allocation).
- `Debug` fallback: `ConditionDataValue::debug(x)` rides any `T: Debug` along as a `character(1)` of `format!("{x:?}")`.

**Keyed builder sugar**: third `__mx_condition_data!` arm — `data = { value = 42, code = 7 }` (bare-ident keys, stringified by the macro). Backward-compatible with `("name", v)` and `[("a", v1), ("b", v2)]`.

**Runtime-built payloads** (issue item 3): covered by the existing enum-literal pattern — fixtures construct `ConditionDataValue::List(...)` / `OptInt(opt)` from runtime values directly. No `data = @expr` escape hatch was needed, so nothing deferred.

**NA-faithful cross-package re-panic path**: `RCondition::from_tagged_sexp` previously dropped any NA-bearing field. It now reverse-maps NA scalars to `Opt*(None)`, NA-bearing vectors to `Opt*Vec`, and nested `VECSXP`s into `List` (via two new helpers `reconstruct_condition_data` / `sexp_to_condition_value`), so `e$field` survives a trait-ABI boundary as `NA`.

## Fixtures + tests

- New `#[miniextendr]` exports in `condition_demo.rs`: `demo_error_data_{option,na_vector,long,nested,debug,keyed}`.
- `gc_stress_condition_data` extended from 8 to 13 fields (adds Option scalar/vector, wide int, Debug, nested list — the nested list allocates a fresh VECSXP + names under torture).
- Rust unit tests in `condition.rs`: keyed-builder arm (idents + trailing comma + mixed types), every new `From`/variant, and an end-to-end macro test carrying NA + nested fields.
- R round-trip tests in `test-conditions.R`: Option present-vs-NA, NA-in-vector, i64 within/beyond i32 range, nested `e$details$min`, Debug `e$range == "0..=100"`, keyed builder, and a no-arg `gc_stress_condition_data()` smoke.
- Regenerated `miniextendr-wrappers.R` + `NAMESPACE` + `man/condition_demo.Rd` + `wasm_registry.rs` in lockstep.

## Verification

| Command | Result |
|---|---|
| `just rcmdinstall` | OK (compiles, regenerates wrappers) |
| `just force-document` | OK |
| `cargo test -p miniextendr-api condition_macro_tests` | 8 passed |
| `just devtools-test` | `FAIL 0 \| WARN 0 \| SKIP 19 \| PASS 6548` |
| gctorture(TRUE) over condition-data path (5× `gc_stress_condition_data` + all 6 demo fixtures) | clean |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` (clippy_default) | OK |
| clippy_all (curated feature list from ci.yml) `-D warnings` | OK |
| `cargo fmt --all` | no changes |

`just wrappers-sync-check` reports drift pre-commit only because it diffs against committed HEAD; the regenerated artifacts are committed here in lockstep.

No UI snapshot regen needed (no proc-macro error wording changed). `rpkg/src/rust/Cargo.lock` git-pin dev-drift discarded; no `inst/vendor.tar.xz` committed.

</details>

Closes #995


CGMossa:

<details>

## Summary
- The failing CI Rust test was `condition_data_roundtrip_na_field_is_dropped`, a pin of the pre-#995 v1 behavior that landed on main via #1021 with a comment deferring NA fidelity "to the Option-variant PR", which is this PR. The rescue rebase pulled the pin in without flipping it, so CI failed on exactly the behavior change this branch exists to make. Fixed in 4e100822: renamed to `condition_data_roundtrip_na_field_survives_as_option_none`, asserting the NA field now reconstructs as `OptInt(None)` instead of being dropped. Also rebased onto current main (clean; the #1129-extended `gc_stress_fixtures.rs` merged coherently, 8 → 13 fields with matching assertions).
- Content review came back clean: `list_into_sexp` follows the documented PROTECT recipe (protect VECSXP, protect names, root each child into the protected container before the next allocation, return unprotected for the caller to root); the reconstruction covers int/real/logical/character scalars and vectors plus nested VECSXP recursion with unknown-type drop as safe degradation; the keyed-builder macro arm is a straightforward third pattern over the existing `From` set.
- Full `cargo test -p miniextendr-api` and clippy pass locally post-rebase.

## Notes
- Sequencing with #1051: that PR folds `ConditionDataValue` into the new `RValue` tree, so these two are on a collision course over the same enum. This one is older and self-contained, and #1051 has to absorb the richer variants either way — merging #1044 first and rebasing #1051 over it is the cheaper order.
- Process observation: a behavior-change PR that sits unmerged while a "pin the current behavior" test lands on main will always fail this way after a rebase. The pin's comment even named this PR as its expiry condition, which made the fix unambiguous.

---

</details>

CGMossa:

<details>

The red CI on the rebased head is not this PR's fault: the only R-side failure is `gc_stress_typed_dataframe` under gctorture ("DataFrame always carries a names attribute"), a latent rooting bug in main's fixture (`List::from_raw_pairs` drops its protection guards on return; the fixture then lets `as_data_frame()` allocate against the unprotected container). It fires stochastically when a branch's extra allocations shift R's node-reuse schedule, which is why main's own CI stayed green.

Fix is up as #1145. Once that merges, rebase this branch over it (or just re-run the failed jobs after merging main in) and CI should go green.

</details>

CGMossa:

**Diagnosis — the failing check is a latent `main` bug, not something this PR introduced (and not the gc_stress issue the worklist guessed).** The `R tests / Linux` (valgrind) job fails inside `load_all()`:

```
Error in `load_all()`:
! Failed to load 'R/miniextendr-wrappers.R'
Caused by error in `methods::setMethod()`:
! no existing definition for function 's4_get_value'
```

`s4_get_value` is a **shared** S4 generic exposed by several `#[miniextendr(s4)]` classes (`CounterTraitS4`, `S4Counter`, `S4TraitCounter`). Its `setGeneric` and the per-class `setMethod`s are emitted as separate entries in the generated `wrappers.R`; installed `library()` load tolerates this, but `load_all()`'s top-to-bottom `source()` errors if any `setMethod("s4_get_value", ...)` is ordered before the `setGeneric`. `git diff origin/main...HEAD` on this branch touches only `condition.rs` and condition demo/test files — **nothing** in `class_system_matrix.rs`, the S4 generators, or `collect_r_wrappers()` — so the `s4_get_value` emission is byte-identical to `main`. `main`'s own `R tests / Linux` job has been *cancelled* (superseded, ~54 min runtime) on every run since the last success at `61d4854c` (#1116), which is why the regression is masked there and only surfaces on a PR run that completes the job.

**Action taken here (per `plans/2026-07-05-ci-failures-open-prs.md`, which scopes this PR to "merge main + rerun, do NOT re-fix, do NOT touch condition.rs"):**
- Branch is already current with `main` (0 commits behind) — the merge-main step is a no-op.
- Filed #1158 for the underlying `load_all` S4-generic ordering bug (with root-cause candidates + a suggested fix on `main`).
- Re-ran the failed job (run 28709918893) — if the wrapper ordering is nondeterministic this clears; if deterministic it will re-fail and the fix belongs on `main` per #1158.

Did **not** touch `condition.rs` / the `#1044`×`#1051` reconciliation, which is owned separately.

</details>

CGMossa:

Merged main twice today: first to pick up #1145 (the `gc_stress_typed_dataframe` rooting fix that this branch's `R tests / Linux` failure was hitting), then again after #1057/#1162 landed mid-flight and re-conflicted the branch. The second merge's `condition_demo.Rd` conflict could not be hand-merged (rerere even replayed a stale placeholder resolution, which was purged): a take-theirs would drop this branch's six new demo aliases, a take-ours would drop #1057's name-titles. Resolved by running the full regen cycle (`rcmdinstall` + `force-document`) in the branch worktree — the regenerated page has both.

Remaining risk: this branch does not yet contain #1163's S4 guard fix, so the valgrind leg may still fail on #1158. If it does, one more main-merge after #1163 lands clears it.

</details>

Issue #989, opened 2026-06-11
Worker-job R errors carry bare simpleError not rust_error class — design question


CGMossa:
## Context

R errors raised from a worker job (via `with_r_thread` routing back to the main thread and calling into the R API, which then longjmps) surface in R as a bare `simpleError` rather than carrying the framework's `rust_error` condition class. Surfaced while working on #931 (`test-worker-longjmp.R` documents the current behavior: "R error inside `with_r_thread` from a worker job surfaces as a bare simpleError").

## Why this is non-trivial

Implementing `rust_error` tagging on the worker-job longjmp path is a multi-day, cross-cutting change:

- **`R_tryCatch` interception** — to re-tag the condition we'd need to intercept the R unwind (before `R_ContinueUnwind` resumes it to top level) with `R_tryCatch`, inspect/wrap the condition object, and re-raise it with the `rust_error` class attached.
- **Widening `run_on_worker`'s error type** — today the dispatcher returns `Result<T, String>`. Carrying a structured condition (class + fields, not just a message string) back across the worker boundary means widening the error variant and every caller that pattern-matches on it.
- **`R_PreserveObject` + `Sendable` SEXP handoff** — the condition is a live SEXP allocated on the main thread. To move it across the worker channel safely it must be `R_PreserveObject`'d and wrapped in `Sendable<SEXP>`, with disciplined release on the consuming side to avoid leaking or GC'ing it mid-flight.

Each of these touches a different layer (FFI unwind, worker channel ABI, GC/protect discipline). It's a deliberate, well-scoped follow-up rather than something to bolt onto the #931 leak/re-usability work.

## Ask

Track the design here. Decide whether `rust_error` tagging on the worker-job error path is worth the cross-cutting cost, and if so, sequence the three changes above.

CGMossa:
Scoped the broader "remove the worker thread entirely" question (PR #1048, doc at `analysis/2026-06-14-worker-thread-removal-scoping.md`). Short version: I think we remove it, and doing so dissolves this issue by construction. Posting the exec summary + decision points here so the discussion lives on the issue.

<details>
<summary><h4>AI-written details</h4></summary>

## Summary
- **Recommendation: remove the worker thread.** It is vestigial. Every guarantee the framework advertises (panic safety, Rust-errors-as-conditions, R-error capture + early-terminate + resurface) is delivered by the main-thread `with_r_unwind_protect` + `catch_unwind` + `extern "C-unwind"` path, independent of the worker.
[...]

PR #1048, opened 2026-06-14
docs(analysis): scope worker-thread removal (#989)

CGMossa:
Do we really need this worker-thread thing? I'd remove it entirely and stick to the solution where Rust errors become R condition objects and R errors get captured, the Rust call terminated early, and the error resurfaced through the R wrapper. This is a scoping doc (no code changes) to let me decide whether to pull the trigger and how. The decision and implementation stay open.

<details>

## Summary
- Adds `analysis/2026-06-14-worker-thread-removal-scoping.md` (doc-only, single file).
- Inventory: the worker lives almost entirely in `miniextendr-api/src/worker.rs` behind `#[cfg(all(feature = "worker-thread", not(target_family = "wasm")))]`; the cargo feature is in `miniextendr-api/Cargo.toml:128` + rpkg mirror; codegen picks worker vs main at `miniextendr-macros/src/lib.rs:910-915`.
- Key finding: the worker is **opt-in** (`#[miniextendr(worker)]`) and structurally barred from any SEXP-carrying signature. Production usage is one `#[cfg]`-gated function (`rng_tests.rs:55`) plus test fixtures. **Zero CI jobs** enable `worker-thread`.
- The worker is **vestigial**: panic safety comes from `extern "C-unwind"` + `catch_unwind` (`unwind_protect.rs:281`), Rust-errors-as-conditions from the tagged-SEXP path, and R-error capture/early-terminate/resurface from `R_UnwindProtect` + `panic_any` + `R_ContinueUnwind` (`unwind_protect.rs:293-351`). All main-thread. The worker cannot even call R (R's `R_CStackStart`/`Limit` checks segfault off-main per `docs/THREADS.md`); it routes back to main via `with_r_thread`.
- The proposed end-state is **already shipping on wasm** (#758): worker compiled-out, everything inline. Removal generalises that to all targets.
- #989 (bare `simpleError` from worker jobs) is **dissolved by construction**: no worker boundary, no `R_ContinueUnwind` bypass of the tagged-condition transport.
- Recommendation: **remove**. Net deletion ~1,200-1,500 lines, collapses a duplicate `R_UnwindProtect` dance, and reduces the threading model to single-thread (extendr-like).
- Flat removal plan (delete codegen branch -> resolve `Sendable` in `externalptr.rs` -> delete worker machinery -> demote `is_r_main_thread` guards -> drop features -> delete fixtures + regen wasm registry -> docs/skills sweep).

## Notes
- Top 3 decision points: (1) `is_r_main_thread` checked-FFI guards in `sys.rs` — demote to `debug_assert!` (recommended) or delete? (2) Is "call R from a user-spawned rayon/std thread, routed back to main" a real requirement to preserve? (recommend no). (3) `Sendable<T>` — delete and inline, or keep a private no-op shim? (recommend delete).
- Biggest risk: silently dropping the off-main-thread *safety net*. Today a rayon-thread R call gets a clean panic from the `sys.rs` guard; deleting (rather than demoting) those guards turns that footgun into a segfault. Mitigation is trivial — keep `is_r_main_thread()`, demote guards to `debug_assert!` — but must be a conscious choice.
- The `Sendable`/`externalptr` provenance edit needs a `gctorture(TRUE)` pass per the `gc_stress_*` convention.

Refs #989

---

</details>

### DIFF
--- a/analysis/2026-06-14-worker-thread-removal-scoping.md
+++ b/analysis/2026-06-14-worker-thread-removal-scoping.md
@@ -466,3 +466,26 @@ demoting them, that footgun becomes a segfault (`R_CheckStack` false overflow pe
 `THREADS.md`). The mitigation is trivial — keep `is_r_main_thread()` and demote
 the guards to `debug_assert!` — but it must be a conscious decision, not an
 incidental casualty of deleting the worker.
+
+---
+
+## Addendum 2026-07-04 — what changed since this was written
+
+Two merges landed after this analysis that bear on the decision; the body
+above is left as written (it was accurate on 2026-06-14).
+
+1. **"Zero CI jobs enable `worker-thread`" is no longer true.** #1116
+   (merged 2026-07-02, audit A5/A10) added scheduled feature-runtime legs to
+   `ci.yml`, including a `worker-default` leg that rebuilds rpkg with the
+   worker feature and runs the full R suite. Coverage is scheduled-only (not
+   per-PR), so the "effectively untested" argument is weakened but not gone —
+   the leg exists precisely because the feature sat in the untested denylist.
+   If removal proceeds, that leg gets deleted with the feature (one more line
+   in the R6 step of the plan).
+2. **#1131 (merged 2026-07-04) added rayon thread-pool control** —
+   parallelism configuration from R, `ensure_pool`, cgroup-aware defaults.
+   This is adjacent to decision point (2) ("call R from a user-spawned
+   rayon/std thread"): the rayon integration's design keeps R contact on the
+   main thread (pools do pure-Rust work), which is consistent with the
+   recommendation here — removal does not conflict with #1131, it removes the
+   one mechanism that pretended off-main R calls were routine.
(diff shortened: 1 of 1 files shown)


CGMossa:

<details>

## Summary
- Reviewed the analysis against the codebase as of today. The inventory, the vestigiality argument (panic safety via `extern "C-unwind"` + `catch_unwind`, tagged-condition transport, `R_UnwindProtect` all main-thread), the wasm precedent, and the flat removal plan all still hold.
- One headline claim went stale while the PR sat: "zero CI jobs enable `worker-thread`" was true on 2026-06-14, but #1116 (merged 2026-07-02) added a scheduled `worker-default` feature-runtime leg that rebuilds rpkg with the feature and runs the full R suite. Rather than rewriting the historical analysis, 6ac99923 appends a dated addendum covering that, plus the #1131 rayon-pool merge, which is adjacent to decision point (2) and consistent with the remove recommendation (rayon pools do pure-Rust work; R contact stays on main).
- Rebased onto current main so CI re-runs against today's tree.

## Notes
- The scheduled-only leg weakens but does not reverse the "effectively untested" argument — the leg exists because the feature was in the untested denylist. If removal proceeds, the leg is one more deletion in the plan's R6 step.
- The decision itself (remove vs keep) remains open for you; nothing found in review argues against the doc's recommendation.

---

</details>

Issue #1052, opened 2026-06-16
Guard the Cargo.lock source-mode dev-drift (tarball-shape stripping)


CGMossa:
## Summary

Every dev-mode `R CMD INSTALL rpkg` / `just rcmdinstall` (and `just check / clippy / build / test / devtools-*`) mutates `rpkg/src/rust/Cargo.lock` as a side effect, dropping the canonical `source = "git+https://github.com/A2-ai/miniextendr#<sha>"` lines from the three framework crates. The committed lockfile must keep those lines ("tarball-shape"). The result is a persistent papercut: after almost any local build, `git status` shows `rpkg/src/rust/Cargo.lock` modified, and the dev must remember to `git checkout` / `just cargo-lock-restore` / `just vendor` before committing — with the pre-commit hook as the only backstop, and only if hooks were installed.

This isn't a new discovery (it's documented in `docs/CARGO_LOCK_SHAPE.md` and trapped in MEMORY across multiple sprints), but there is no open issue tracking a *guard/fix* for the DX cost. #680 (closed) only fixed the hook's silent-reject failure mode; it didn't prevent the drift.

## Mechanism (with file:line cites)

Source-mode `configure` emits a git-source-URL patch override into the generated `.cargo/config.toml`:

- `rpkg/configure.ac:851-866` — the `monorepo` branch of `AC_CONFIG_COMMANDS([cargo-config])` writes:
  ```
  [patch."https://github.com/A2-ai/miniextendr"]
  miniextendr-api    = { path = "<root>/miniextendr-api" }
  miniextendr-lint   = { path = "<root>/miniextendr-lint" }
  miniextendr-macros = { path = "<root>/miniextendr-macros" }
  ```
  (generated artifact: `rpkg/src/rust/.cargo/config.toml:9-12`). Note it is a `[patch."<git-url>"]` form, **not** `[patch.crates-io]`.

When cargo resolves under that patch, it records the framework crates as local (no-`source`) entries — stripping the `git+...#<sha>` line from each in `rpkg/src/rust/Cargo.lock`.

- `docs/CARGO_LOCK_SHAPE.md:130-151` — enumerates every drift trigger (`just check/clippy/test`, `R CMD INSTALL` source mode, `devtools::*`, `just devtools-document`).
[...]

PR #1053, opened 2026-06-16
build: gitignore generated wrappers.R + wasm_registry.rs (regenerate into tarball)

CGMossa:
Stop tracking the two large generated files (`R/miniextendr-wrappers.R` ~32k lines, `src/rust/wasm_registry.rs` ~13k lines) that caused constant merge conflicts — treat them like `inst/vendor.tar.xz`: regenerated on disk by the host cdylib pass, shipped in the tarball from disk, never committed.

<details>
<summary>🤖 Implementation details (AI-written)</summary>

### Why
Both files are generated byte-deterministically by the host cdylib pass on every `R CMD INSTALL`. Committing them produced merge conflicts on nearly every macro/wrapper change — the exact churn that got the vendor tarball gitignored.

### Theory verification (done in a worktree)
- A single host `R CMD INSTALL` regenerates **both** files **byte-identically** (same `content-hash`) — fully deterministic.
- A built tarball **contains both** despite being gitignored (`R CMD build` honors `.Rbuildignore`, not `.gitignore`; neither path is excluded). Confirmed via `tar tzf`.
- The two-step host→wasm build already exists in `.github/workflows/webr.yml` (Phase 1 native install regenerates the snapshot; Phase 2 cross-compiles).
- The rewritten `just wrappers-sync-check` passes.

### Changes
- **`rpkg/.gitignore` + all 3 scaffolding templates**: ignore `R/*-wrappers.R` and `src/rust/wasm_registry.rs`. The templates previously ignored only `src/*-wrappers.R` — a stale path that never matched the real generated files.
- **`justfile`**: `r-cmd-build` / `r-cmd-check` now depend on `r-cmd-install`, so the host pass regenerates the untracked files into the source tree (source mode) before `vendor` seals the tarball. This is load-bearing: **wasm-from-tarball has no regeneration fallback** — the wasm cdylib is a `SIDE_MODULE` that host R cannot `dyn.load`, so a missing/stub `wasm_registry.rs` silently breaks the wasm install. `wrappers-sync-check` now asserts the files are present + non-stub (content-hash != all-zero) instead of git-diffing them, and git-diffs only the still-tracked `NAMESPACE` + `man/`.
- **`build.rs`** (+ both template copies): clearer panic when `wasm_registry.rs` is absent on a wasm32 build — explains the crate is not directly installable and the two-step host→wasm flow. Template copies use `R CMD INSTALL` (not `just`, which is maintainer-only).
- **`.github/workflows/ci.yml`** "Generated Files Check": enforce both stay untracked (prevents accidental re-commit reintroducing the conflict).
- **`.githooks/pre-commit`**: removed the now-dead "wrappers staged without NAMESPACE" branch (gitignored files can't be staged); NAMESPACE/man drift is covered by CI's `wrappers-sync-check`.
- **Docs**: root `CLAUDE.md`, `AGENTS.md`, `rpkg/CLAUDE.md`, `wasm_registry_writer.rs` docstring.
- **`patches/templates.patch`**: re-locked for the build.rs message delta (`just templates-check` passes).

### Follow-up
- #1052 — guard the Cargo.lock source-mode dev-drift (separate papercut surfaced during this work).

</details>


CGMossa:
I'm tired of the constant merge conflicts stemming from the unmergable artifacts that are generated when installs are made. Therefore, this PR should be a step into removing the generated artifacts from git, and I hope this cements workflows so that they ensure the artifacts exists in r packages.

Issue #1055, opened 2026-06-17
Make DataFrame::builder available without the rayon feature


CGMossa:
## Summary

Make `DataFrame::builder` (`RDataFrameBuilder`) available without the `rayon` feature.

## Motivation

A consumer wants to build a `data.frame` column-by-column from Rust, deciding each column's name and type imperatively. The current options both fall short for this case:

- `vec_to_dataframe(&rows)` infers the schema from serde. When a row type has a nested enum field (e.g. an externally-tagged `enum Action { Add {...}, Init {...} }`), the columnar serializer emits that field as a nested **list-column** rather than flattening it. The consumer does not want a list-column.
- Writing a flattened serde *view* type (a mirror struct/enum + `From` impl) works, but forces the consumer to reach into and re-encode another crate's internal type layout. The whole point of serde inheritance is to *not* do that.

The natural escape hatch is a builder: push the columns you want, by name and type, straight from the deserialized values. `DataFrame::builder(nrow).column_str(...).column::<f64>(...).build()` is exactly right — but it is gated behind `#[cfg(feature = "rayon")]`, so consumers that don't otherwise want rayon (and the crossbeam-* deps it pulls in) can't reach it.

## Ask

Provide a builder path that does not require the `rayon` feature. Some options:

- Split the API so the column-registration surface (`column`, `column_str`, `build`) is available unconditionally, and only the *parallel* fill strategy is gated behind `rayon` (serial fill otherwise).
- Or expose a lightweight, always-available column-push builder that takes already-materialized column vectors / SEXPs.

## Notes

`List::as_data_frame()` already works without rayon, but it requires manually converting each column to a SEXP and assembling a named list by hand. A first-class builder available outside `rayon` would be the clean answer.

Found while building `dvs_audit_log()` in the dvs R bindings, which materializes an audit log (`Vec<AuditEntry>`, where `AuditEntry.action` is an externally-tagged enum) as a data.frame.

Issue #1054, opened 2026-06-17
Function with a doc comment but no roxygen tags gets no @title (so roxygen2 generates no .Rd)


CGMossa:
### Summary

An `#[miniextendr]` **function** whose doc comment is plain prose with **no roxygen tags** (`@param` / `@return` / `@name` …) does not get an `@title` in the generated R wrapper. roxygen2 then skips generating an `.Rd` for a title-less block, so the exported function ends up with **no reference page** even though it has a perfectly good doc comment.

### Reproduction

```rust
/// Version of the bundled DVS Rust core crate.
#[miniextendr]
pub fn dvs_version() -> String {
    dvs::VERSION.to_string()
}
```

Generated wrapper — no `@title`:

```r
#' @source Generated by miniextendr from Rust fn `dvs_version`
#' @export
dvs_version <- function() { ... }
```

→ roxygen2 emits no `man/dvs_version.Rd`.

Compare a function that happens to take an argument, so it has `@param`:

```rust
/// Format a byte count as a human-readable size string.
///
/// @param size_bytes non-negative integer representing file sizes in bytes.
#[miniextendr]
pub fn format_byte_size(size_bytes: u64) -> String { ... }
```

→ gets `#' @title Format a byte count as a human-readable size string` and an `.Rd`.

So whether an exported function gets a reference page depends on whether it happens to carry another tag — which is surprising.

### Root cause

`miniextendr-macros/src/roxygen.rs` (current `main`, ~L143-152):

```rust
let has_any_tags = !tags.is_empty();

// Auto-generate @title from implicit title if:
// - We have @name but no @title, OR
// - We have any tags (like @param/@return) but no @title
if (has_name || has_any_tags)
    && !has_title
    && let Some(title) = implicit_title_from_attrs(attrs)
{
    tags.insert(0, format!("@title {}", title));
}
```

For a function with only a prose doc comment and no args, `has_name == false` and `has_any_tags == false`, so the implicit `@title` (which `implicit_title_from_attrs` would happily return) is never inserted.

There is already an analogous tagless fallback for **method** `@description` further down:

```rust
// Original auto_description behavior for methods without any tags
if auto_description && tags.is_empty() && !regular_docs.is_empty() {
    let description = regular_docs.join(" ");
    tags.push(format!("@description {}", description));
}
```

…but nothing emits an implicit `@title` in the tagless case.

### Suggested fix

[...]

PR #1062, opened 2026-06-17
docs: second-install for new exports + rig-default non-persistence traps

CGMossa:
Documents two operational traps hit during the #1054/#1055/#1056 sprint.

<details>
<summary>🤖 Details (drafted with Claude Code)</summary>

- **New export needs a second install** (CLAUDE.md "After Rust changes" + AGENTS.md): `just rcmdinstall` regenerates `wrappers.R` but installs against the *existing* `NAMESPACE`; `force-document` writes the new export to `NAMESPACE` on disk but doesn't reinstall — so a freshly added `#[miniextendr]` fn is missing from the *installed* package until you install again. To gctorture/testthat a new export: `rcmdinstall && force-document && rcmdinstall`. (Committing the regenerated docs needs only the single pass.) Hit twice in the sprint (#1055 fixture not found at runtime; #1054 undefined-export on a removed fn).
- **`rig default` may not stick across shells** (CLAUDE.md rv/rig section): a fresh shell can resolve `R` back to the system default (framework `Current` symlink flips), silently reverting mid-session and breaking installs with the `__rv_R_mismatch` safe-mode error. Re-verify `R --version` per R-recipe batch. R reverted 4.6 → 4.5.2 mid-sprint.

AGENTS.md mirrors only the export note (it has no rv/rig section).
</details>

Docs-only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -114,6 +114,8 @@ Use `just force-document` (not `just devtools-document`) after **anything** that
 
 `just devtools-document` (short-circuiting variant) is safe for **pure R/roxygen changes** — skips roxygenize when `needs_roxygenize()` returns `FALSE`.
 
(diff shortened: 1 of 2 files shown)


CGMossa:
Maybe this needs a better solution, but I am unsure.

PR #1057, opened 2026-06-17
fix(macros): title generated R wrappers by name, not rustdoc prose (#1054)

CGMossa:
**TODO: replace this line with your own framing, in your own voice.** Reworked per review: rustdoc-style doc-comment prose (intra-doc links, code spans) isn't valid roxygen `@title` text, so the title now comes from the structural name (function / class) and the prose is demoted to a sanitized `@description`.

<details>
<summary><h3>AI-written details</h3></summary>

## Summary
- Root problem: a prose-only `#[miniextendr]` fn got no `@title`, so roxygen2 skipped its `.Rd` (#1054). The first attempt promoted the doc-comment's first sentence to `@title`, but `rpkg` has `Config/roxygen2/markdown: TRUE`, so roxygen2 reads rustdoc `[`Foo`]` / `[x][crate::y]` links as R `\link{}` to help topics and fails ("could not resolve link to topic" / "refers to un-installed package"). That was the `Sync Checks` failure.
- New approach: `@title` is always the structural name, never prose.
  - Standalone functions: `@title` = the R wrapper name (`lib.rs`).
  - Class blocks: `@title {Name} Class`, already emitted by `ClassDocBuilder` (it now fires because the core no longer synthesizes a prose title).
  - Leading prose (all paragraphs before the first `@tag`) is promoted to a single `@description`, with rustdoc link syntax neutralized by a new `sanitize_roxygen_links` helper. Genuine `[text](url)` markdown links are preserved.
- Removed the core's prose `@title`/`@details` synthesis and the now-unused `implicit_details_from_attrs` helper.
- Regenerated `man/` (204 files). `NAMESPACE` unchanged (no export drift). S7 `@prop` text now keeps the full sentence instead of a truncated first-sentence title.

## Test plan
- [x] `cargo test -p miniextendr-macros` (348 unit + UI/trybuild + doctests) green; 4 impl snapshots refreshed and reviewed.
- [x] New unit coverage for `sanitize_roxygen_links` (shortcut / reference / markdown / UTF-8) and `leading_prose_from_attrs`.
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (default features) clean.
- [x] clippy with the `clippy_all` curated feature list clean.
- [x] `just wrappers-sync-check` passes (no "could not resolve link" / "un-installed package" warnings; `NAMESPACE` / `man/` in sync).
- [x] `just templates-check` passes.

## Notes
[...]

PR #1066, opened 2026-06-20
feat(rpkg): serde "satellite-crate" experiment harness

CGMossa:
<TODO: your framing paragraph>

<details>

## Summary
- Adds a miniextendr-FREE crate `rpkg/src/rust/satellite/` (serde only, sealed `[workspace]`) and a bridge module `satellite_bridge.rs` that exposes its types to R purely through `miniextendr_api::serde`. `cargo tree` for the satellite crate is `satellite -> serde` only; zero miniextendr in its dependency graph.
- Proves the satellite architecture for the *data* surface: a third-party serde crate gets R interop with no per-type conversion code. The bridge exercises `vec_to_dataframe` (columnar + nested flatten), `vec_to_dataframe_split` (enum), `AsSerialize`, and bidirectional `from_r`. A satellite type's own `serde = "1"` unifies with miniextendr-api's, so its derived impls satisfy the bridge bounds directly.
- The `satellite` cargo feature is a normal optional integration feature: auto-enabled by `tools/detect-features.R` like the other integration features, so every `R CMD INSTALL` (including CI's install + wrappers-sync-check jobs) compiles the bridge and generates R wrappers for the `satellite_*` exports. (An earlier revision denylisted it as dev-only; that was reverted in the "make it a normal optional feature" commit.)
- `satellite/CLAUDE.md` pins the single invariant (stay miniextendr-free) and the verify command.
- Documents the pattern in `docs/SERDE_R.md` (new "Satellite crates" section: layout, bridge code, capability table, and the data-vs-behaviour boundary — what serde alone cannot give you: ExternalPtr, R6/S3/S4/S7, ALTREP, connections, conditions, dots).
- Fixes a latent gate bug this feature surfaced: `json_string_tests` was gated on `serde` but uses `serde_json`-only items; it only ever built because `full` always pairs the two. `--features satellite` is the first combo to enable serde alone. Re-gated on `serde_json`.

## Test plan
- [x] `cd rpkg/src/rust && cargo check --features satellite` (clean)
- [x] `cd rpkg/src/rust && cargo clippy --features satellite -- -D warnings` (clean)
- [x] `cd rpkg/src/rust/satellite && cargo tree` shows no miniextendr crate (satellite -> serde only)
- [x] Full CI green on the feature-enabled revision (R CMD INSTALL jobs compile the bridge via detect-features; Sync Checks verifies the generated NAMESPACE/man)

## Notes
- The satellite library's own `Cargo.lock` is gitignored (libraries do not pin). The committed `rpkg/src/rust/Cargo.lock` gains the `satellite` path-dep entry (optional deps are locked regardless of feature activation; without the entry every local resolve dirtied the lockfile).
- `--features satellite` enables `serde` without `serde_json` (the first combo to do so), which is what surfaced the `json_string_tests` gate bug; it is now re-gated on `serde_json`.

---

</details>


CGMossa:

<details>

Reviewed, rebased onto current main (clean, no conflicts), and fixed one real defect. Pushed as 5744f995.

## The fix: `Cargo.lock` was missing the `satellite` entry

Cargo locks optional dependencies regardless of feature activation, so once `satellite = { path = "satellite", optional = true }` entered the manifest, every `cargo` resolve added the `satellite 0.1.0` entry to `rpkg/src/rust/Cargo.lock` and dirtied VCS on every build. The PR's "committed Cargo.lock is unchanged" note was true for the original dev-only design, but went stale when the later commit made `satellite` a normal auto-detected feature. Committed the two-hunk lock addition (the package entry plus `satellite` in rpkg's dep list).

Related: the PR body still described the abandoned dev-only design (denylisted in `detect-features.R`, "CI does not build `--features satellite`"), both inverted by the "make it a normal optional feature" commit. Updated the body to match the code.

## Review assessment

The architecture is sound and the invariant is real: `cargo tree` from `satellite/` is `satellite -> serde` only, verified locally. The sealed `[workspace]` plus `exclude = ["satellite"]` pairing is the right way to keep the path dep out of both workspaces. The `json_string_tests` re-gate (`serde` to `serde_json`) is a genuine latent-bug fix: `--features satellite` is the first combination that enables `serde` without `serde_json`. The `docs/SERDE_R.md` section is the strongest part of the PR; the data-vs-behaviour boundary table (what serde gives you vs what needs miniextendr-native code) answers the question end users will actually ask.

Local verification: `cargo check --features satellite`, `cargo clippy --features satellite -- -D warnings`, `cargo fmt --check` (rpkg and satellite), all clean. First attempt failed with stale `group_by` resolution errors: the worktree was missing `.cargo/config.toml`, so cargo resolved `miniextendr-api` from a stale git pin instead of the rebased siblings. `just configure` repaired it; not a PR defect.

## One observation for a follow-up

The generated `man/satellite_bridge.Rd` drops the doc text for four of the five bridge functions: their `\description` entries collapse to the group placeholder "satellite bridge", and only `satellite_echo_reading` keeps (the first sentence of) its rustdoc. This is grouped-page implicit-title handling in the wrapper emitter, exactly the territory #1057 touches, not something to fix in this PR. Worth regenerating this page after #1057 merges.

## Insight

A PR body is part of the diff: when a later commit inverts the design (dev-only became auto-enabled here), the stale prose actively misleads reviewers about what ships to CRAN. And the concrete cargo fact worth remembering: optional path deps are locked unconditionally, so "the feature is off by default" never means "the lockfile is unaffected".

</details>

PR #1074, opened 2026-06-22
Vendor path-dependency monorepo siblings (freeze leaves git deps as git)

CGMossa:
<TODO: user's framing paragraph — e.g. why we want path-dependency monorepo siblings to ship self-contained tarballs, and that this unblocks dvs-rpkg dropping its dvs git pin.>

<details>

## Summary

Lets a miniextendr R package depend on a monorepo sibling core crate by path (`core = { path = "../../../core" }`) and still ship a self-contained, offline-installable tarball. Two changes:

- **`fix(cargo-revendor)`**: `freeze_manifest` rewrote every crate in `local_pkgs` to a `vendor/` path, including git deps that resolve to a local crate via a `.cargo/config.toml` `[patch."git+<url>"]` (the monorepo dev override). Those deps are declared `git =` and are external by declaration: they must stay `git =` in the frozen manifest and resolve offline via source replacement (the `git+<url>#<sha>` shape that `lock-shape-check.R` enforces). The rewrite changed a working, tested shape and dirtied the manifest. Freeze now rewrites only deps the manifest declares as path deps, and adds `[patch.crates-io]` only for those. Git deps are left as git and surface in the existing remaining-git warning, matching what the README's freeze section already documents.
- **`feat(templates)`**: `bootstrap.R` and the `just vendor` recipes pass `--freeze`. A path dep is not source-replaceable, so without freeze the sealed `Cargo.toml` references a sibling that does not travel inside the tarball. `cran-prep` restores `src/rust/Cargo.{toml,lock}` after the build, since freeze rewrites them in place and the working tree must return to its source-shape form.

## Why it is safe for existing packages

For a git-only package, or a package whose framework crates resolve via the monorepo `[patch]` override, freeze now touches no declared-git dep, so it is inert (no manifest rewrite, just the usual lock normalisation).

## Test plan

- [x] `cargo test` in `cargo-revendor` (106 unit tests pass, including a new regression test: a manifest with both a git-patched-local dep and a path dep in `local_pkgs` keeps the git dep as git and freezes only the path dep)
- [x] `cargo fmt --check` and `cargo clippy` clean
- [x] Full `bootstrap.R` (with `--freeze`) on the model `rpkg/` with the gated binary: succeeds, leaves `miniextendr-{api,lint}` as `git =`, manifest stays clean, lock via the "copied from vendored sources" path
- [ ] CI: `bootstrap-produces-vendor.sh`, `vendor-*` and `cross-package` suites
- [ ] Downstream: dvs-rpkg adopts a `dvs = { path = ... }` dependency and installs via `install_github` / a clean-room tarball (verified locally against this branch's behavior)

## Notes

- `--freeze` alone is sufficient in `bootstrap.R`; cargo-revendor auto-detects the source root from `cargo metadata`, so no `--source-root` is needed.
- This is the upstream half of dropping dvs-rpkg's `dvs` git pin in favor of a path dependency on the in-repo core crate.

---

</details>


CGMossa:
Pushed a follow-up commit (`e9769c9f`) from a review pass over this branch. It finishes the path-dep-sibling story across the install methods and cleans up two things the `--freeze` change introduced. Details below.

<details>

## Summary

Five fixes, all verified end-to-end by scaffolding fresh packages and installing them:

- **cargo-revendor `--freeze` `--help` corrected.** It still said "Rewrites git deps to vendor/ path deps", the opposite of this branch's behavior (and it contradicted the `--strict-freeze` doc right below it). Rewritten to describe the path-only rewrite + git-stays-git.
- **`cran-prep` trap split into two `git checkout`s** (both template justfiles). `git checkout -- Cargo.toml Cargo.lock` is atomic, so an untracked `Cargo.lock` (the case on a fresh scaffold) aborted the whole restore and left the FROZEN `Cargo.toml` (path sibling rewritten to `vendor/`) in the working tree, which then breaks the next source build since `vendor/` is gitignored.
- **Monorepo `bootstrap.R` now auto-vendors `--freeze`** (the fallback the single-package bootstrap already had). The monorepo bootstrap previously only ran `./configure`, so `remotes::install_git` / `devtools::install` / `pkgbuild` / CRAN built the package subtree without ever sealing the sibling and failed with `failed to read .../my-core/Cargo.toml`. The sibling is reachable at `R CMD build` time (whole-repo clone or in-place build), so vendoring there makes the tarball self-contained.
- **Monorepo template now wires `rpkg` to the core sibling by default** (path dep plus a `core_greeting()` that calls it), with `crate_name_rs` (dash to underscore) added in `create.R`. Previously the template scaffolded `my-crate/` but nothing depended on it, so the feature was never exercised out of the box.
- **`miniextendr_build()` snapshots/restores `src/rust/Cargo.{toml,lock}` and drops a freshly-leaked `inst/vendor.tar.xz`.** A `build = TRUE` install runs bootstrap in the source tree, which (with the new auto-vendor) froze the in-source manifest. The dev loop now leaves a clean source-shape tree, and as a side benefit the single-package case no longer leaks a tarball either.

## Verified

- Monorepo + sibling: `R CMD INSTALL <dir>`, `pkgbuild` build + offline-isolated install (sibling physically moved away, compiles from `vendor/`), `remotes::install_git`, `devtools::install`, and `miniextendr_build` all install and run (`rust_gcd`, `lcm`, `core_greeting`).
- Single package (git-only, `--freeze` inert): full matrix still green; dev loop leaves no tarball leak.
- `just templates-check`, `templates-recipes-check.sh`, the affected minirextendr tests, and the cargo-revendor build all pass after rebasing onto the merged `main`.

## Notes

- **`pak` and `rv` still need `just vendor` first** for a monorepo + sibling. Both copy only the package subtree before building, so the out-of-tree sibling is gone before any bootstrap runs (and `rv` does not run the bootstrap hook at all). This is not fixable via an absolute-path override in `.cargo/config.toml`: `src/rust/.cargo` is `.Rbuildignore`d so it never travels into their build copy, configure runs in the copy where it cannot recover the sibling's original absolute location, and an absolute path would not be portable to CI/CRAN/other clones anyway. Pre-vendoring (which they both honor) is the portable answer. Verified: pak and rv both succeed once the package is pre-vendored.
- Not yet run: `R CMD check --as-cran` on a scaffolded monorepo.

---

</details>

CGMossa:
Split the scaffolding de-just out into #1078 to keep this PR focused on the `--freeze` path-dep-sibling work. Force-pushed `91d90ff9` (drops the combined de-just commit; `e9769c9f` is unchanged).

<details>

## What stayed here vs. moved to #1078

The `just`-removal work touched two kinds of content, so I split it along that line.

**Stays in this PR (freeze-coupled):**
- `vendor_crates_io()` now passes `--freeze`, so `miniextendr_vendor()` seals a local path-dependency sibling the way the workspace `just vendor` recipe did — without it the just-free vendor path wouldn't self-seal a monorepo sibling.
- `miniextendr_vendor()` closing guidance documents restoring the frozen `src/rust/Cargo.{toml,lock}` after `R CMD build`.
- De-just of the auto-vendor comments in the monorepo `bootstrap.R` block that this branch introduced.

**Moved to #1078 (de-just of content already on `main`):**
- Deleting the two template justfiles + the scaffolder's `copy_template("justfile")` call.
- De-just guidance in template `configure.ac` / `lock-shape-check.R` / README / vignette / `create.R` next-steps / `miniextendr_doctor()`.
- Removing the orphaned `templates-recipes-check` script + root recipe + CI step.
- The test assertion flip.

The split line was `git diff main...e9769c9f` — files the freeze work already touches are entangled and keep their de-just here; everything else is pure-`main` content and went to #1078.

## Rebase note

This branch still **modifies** the two template justfiles (the `cran-prep` trap split), while #1078 **deletes** them. After #1078 merges, rebasing this branch onto `main` will surface a modify/delete conflict on those justfiles (resolve → deleted) and a `patches/templates.patch` conflict (resolve → `just templates-approve`). Both branches currently pass `just templates-check` independently.

</details>

Issue #1087, opened 2026-06-26
Leash remaining `'static`-pinned borrowed TryFromSexp arg types (Cow, Vec<&'static str>, scalar refs)


CGMossa:
Follow-up to the StrVec/RCow leash work (PR for `fix/leash-rvec-views`), which made the custom view types lifetime-parameterised so a borrow into R-owned memory cannot escape the enclosing `.Call` (use-after-GC). The same `&'static`-from-R escape class still exists in several **std-type** `TryFromSexp` impls that hard-code `'static`.

## The hazard

A `TryFromSexp` impl that yields a borrow into R vector/CHARSXP data with lifetime `'static` lets the borrow be stored in an `ExternalPtr` / global / sent cross-thread. The data lives only while R keeps the source SEXP reachable (R protects `.Call` arguments for the call's duration). Stored past the call → dangling.

## The invariance distinction (why some are already safe)

- **Covariant bare refs** — `&[T]`, `&mut [T]`, `&str`: already leashed. The user fn's elided lifetime (`fn f(x: &[i32])`) is a fresh `'a`; even when the impl produces `&'static`, covariance narrows it to `'a`, so the borrow can't escape. `&[T]`/`&mut [T]` additionally have generic blanket impls (`from_r.rs:586`). `&str` (`strings.rs:47`) is `'static`-only but covariant ⇒ still leashed in the elided form. **Lower priority** — only an explicit `&'static` annotation by the user is unsound, and that's opt-in.
- **Invariant / forced-`'static` containers** — `Vec<&'static T>`, `Vec<&'static [T]>`, `Vec<&'static mut [T]>`, `Vec<&'static str>`, `Cow<'static, [T]>`, `Cow<'static, str>`, `Vec<Cow<'static, str>>` (+ `Option`/`mut` variants): **HIGH priority**. `Vec`/`Cow` are invariant in the element lifetime, and the only `TryFromSexp` impl is for the `'static` instantiation — so the user is *forced* to write `Vec<&'static str>` / `Cow<'static, [T]>` (there is no `Vec<&'_ str>` / `Cow<'_, …>` impl to resolve). That mandatory `'static` is escapable. This is the same forced-unsound shape RCow had before it was generalised.

## Affected impls

[...]

PR #1088, opened 2026-06-26
fix: leash StrVec and RCow borrows to a lifetime (kill the `&'static`-from-R escape)

CGMossa:
## Problem

`StrVec` and `RCow`'s borrowed arm hand out references straight into R-owned CHARSXP / vector data. Both hard-coded `'static`:

- `StrVec::get_str(self) -> Option<&'static str>`
- `impl TryFromSexp for RCow<'static, T>`

That `'static` is a lie. The data lives only while R keeps the source SEXP reachable (R protects `.Call` arguments for the call's duration). A borrow stowed in an `ExternalPtr`, a global, or sent cross-thread dangles the moment R GCs the source.

**Demonstrated empirically** during investigation: `StrVec::get_str` → stored in an `ExternalPtr` → `rm() + gc()` → the recreated string lands at a different address; the original reads freed memory. No `unsafe` required.

## Fix — leash to a real `'a`

- **`StrVec` → `StrVec<'a>`** (covariant `PhantomData<&'a str>`). Safe `get_str` / `get_cow` / `iter` / `iter_cow` / `IntoIterator` return `&'a` / `Cow<'a>`. `'static` is relegated to the new `unsafe get_str_static` / `get_cow_static` escape hatch (for callers who own the rooting obligation, e.g. having stashed the source SEXP in an `ExternalPtr`'s `prot` slot). `StrVecBuilder::into_strvec` leashes to the builder's protect scope. `ProtectedStrVec` wraps `StrVec<'static>` — sound, because its protect guard *is* the rooting obligation.
- **`RCow`** — `impl TryFromSexp for RCow<'static, T>` → generic `impl<'a, T> … RCow<'a, T>`. Users write `RCow<'_, T>`; the module's honor-system "do not store past the call return" note becomes **compile-enforced**.

At a `#[miniextendr]` boundary the user writes `StrVec<'_>` / `RCow<'_, T>`. The anonymous lifetime is a fresh generic the generated wrapper cannot widen to `'static`, so storing a borrow past the call is a compile error (E0521). The C wrapper stays `SEXP`-in/out — lifetimes erase at codegen, so this needs no macro change (anonymous `<'_>` already works with the current macro; explicit `<'a>` is enabled by #1086, on which this stacks).

Sound models already in tree, left unchanged: `Factor<'a>`, `ProtectedStrVec`.

## Test

`miniextendr-macros/tests/ui/view_type_lifetime_leash.rs` (trybuild compile-fail) reproduces the original use-after-GC PoC — storing a leashed `StrVec<'_>` / `RCow<'_, _>` borrow in a `'static` field — and asserts it no longer compiles:

```
error: lifetime may not live long enough
  ... this usage requires that `'1` must outlive `'static`
```

## Scope / follow-up

The remaining `'static`-pinned **std-type** `TryFromSexp` conversions (`Cow<'static, […]>`, `Cow<'static, str>`, `Vec<&'static str>`, `Vec<&'static [T]>`, scalar `&'static T`) are the same escape class but a broad, mechanical change across ~50 generated impls. Tracked in #1087. Note the bare covariant `&[T]` / `&mut [T]` / `&str` arg forms are *already* leashed by the user fn's elided lifetime (and `&[T]`/`&mut [T]` already have generic blanket impls), so the high-value targets there are the invariant `Vec<&'static …>` / `Cow<'static, …>` families.

## Verification

- `just check` — green across all manifests (incl. rpkg fixtures + cross-package).
- `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean.
- `cargo fmt --all` — clean.
- `cargo test -p miniextendr-macros --test ui` — passes (new compile-fail snapshot stable, no other UI snapshots drifted).

`clippy_all`'s feature matrix not separately re-run: the diff touches only `strvec.rs` / `rcow.rs` (always-compiled, no feature gates) plus a trybuild test, so the extra features cannot surface new lints in the changed code.

## Stacking

Based on `feat/miniextendr-accept-lifetimes` (#1086) for clean history; **no hard code dependency** on it (uses only anonymous `<'_>`). When #1086 merges, this auto-retargets to `main`.

### DIFF
--- a/miniextendr-api/src/rcow.rs
+++ b/miniextendr-api/src/rcow.rs
@@ -33,8 +33,10 @@
 //! # Lifetime contract
 //!
 //! As with `Cow<[T]>`, a borrowed `RCow` is valid only for the duration of the
-//! enclosing `.Call` (R protects the argument SEXP while Rust runs). Do not send
-//! it to another thread or store it past the call return.
+//! enclosing `.Call` (R protects the argument SEXP while Rust runs). Write
+//! `RCow<'_, T>` at a `#[miniextendr]` boundary: the `'a` leashes the borrow to
+//! the call, so sending it to another thread or storing it past the call return
+//! is a *compile error*, not an honor-system pitfall.
 
 use std::ops::Deref;
 
@@ -82,13 +84,13 @@ impl<T> RBorrow<'_, T> {
 /// ```ignore
 /// // Zero-copy in *and* out: the returned SEXP is the original R vector.
 /// #[miniextendr]
-/// pub fn passthrough(x: RCow<'static, f64>) -> RCow<'static, f64> {
+/// pub fn passthrough(x: RCow<'_, f64>) -> RCow<'_, f64> {
 ///     x
 /// }
 ///
 /// // Mutating forces a copy (copy-on-write), then materializes a fresh vector.
 /// #[miniextendr]
-/// pub fn doubled(mut x: RCow<'static, f64>) -> RCow<'static, f64> {
+/// pub fn doubled(mut x: RCow<'_, f64>) -> RCow<'_, f64> {
 ///     for v in x.to_mut() {
 ///         *v *= 2.0;
 ///     }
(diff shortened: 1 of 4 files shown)


CGMossa:
Accuracy note after empirically probing the related `'static`-pinned std-type conversions (tracked work, now closed as #1087):

The **StrVec** change here is the substantive soundness fix — `StrVec` was a lifetime-*less* `Copy` struct whose `get_str(self) -> &'static str` fabricated a `'static` borrow with nothing to leash to, so `fn f(v: StrVec)` (the only spelling) genuinely dangled.

The **RCow** change is cosmetic *for soundness*: `RCow`'s elided `RCow<'_, T>` arg form was already leashed on `main` (verified — the macro's C wrapper picks `'static` for the conversion, but the user fn's elided `'a` leashes the borrow; storing it past the call is E0521). It's still worth keeping because it makes `RCow<'_, T>` the canonical resolvable form and fixes the old doc examples that taught the escapable explicit `RCow<'static, …>` spelling. Same applies to `Cow<'_, …>` / `Vec<&str>` more broadly — all already leashed in the elided form, so #1087 needed no impl changes.

PR #1090, opened 2026-06-28
fix(api): protect self.sexp across CHARSXP alloc in DataFrame::rename/strip_prefix

CGMossa:
Hardens GC discipline in two `DataFrame` editing methods to match the rest of the file. From `audit/dataframe.md`.

<details>
<summary>🤖 Generated by Claude Opus 4.8 — details</summary>

**Finding.** `miniextendr-api/src/dataframe.rs` — `DataFrame::rename` and `DataFrame::strip_prefix` allocate a CHARSXP via `SEXP::charsxp(...)` while `self.sexp` is not GC-protected.

**Root cause.** `SEXP::charsxp` calls `Rf_mkCharLenCE`, which can allocate and trigger R's GC. `self.sexp` lives only on the Rust stack (R's GC does not scan it), so it — and the `names` attribute we mutate, reached through it — could be collected mid-operation. The sibling editing methods `drop`/`select` already guard with `crate::OwnedProtect`; these two did not.

**Fix.** Root `self.sexp` for the duration of each `unsafe` block with `let _guard = crate::OwnedProtect::new(self.sexp);`, placed before `self.sexp.get_names()`. Rooting `self.sexp` keeps its `names` attribute reachable across the `SEXP::charsxp` allocation. Mirrors the idiom already used in `drop`/`select` in the same file. Diff is limited to these two functions.

**Impact.** No R-visible output change; internal GC-discipline hardening only.

**Verification.** `cargo fmt --all`, `cargo check -p miniextendr-api`, and `cargo clippy -p miniextendr-api --all-targets -- -D warnings` all pass.

This came from `audit/dataframe.md`.
</details>

### DIFF
--- a/miniextendr-api/src/dataframe.rs
+++ b/miniextendr-api/src/dataframe.rs
@@ -272,6 +272,9 @@ impl DataFrame {
     /// Rename a column. No-op if `from` doesn't match any column name.
     pub fn rename(self, from: &str, to: &str) -> Self {
         unsafe {
+            // Root `self.sexp` so its `names` attribute survives the
+            // `SEXP::charsxp` (Rf_mkCharLenCE) allocation below, which can GC.
+            let _guard = crate::OwnedProtect::new(self.sexp);
             let names_sexp = self.sexp.get_names();
             if names_sexp == SEXP::nil() {
                 return self;
@@ -290,6 +293,9 @@ impl DataFrame {
     /// Strip a prefix from all column names that start with it.
     pub fn strip_prefix(self, prefix: &str) -> Self {
         unsafe {
+            // Root `self.sexp` so its `names` attribute survives the
+            // `SEXP::charsxp` (Rf_mkCharLenCE) allocation below, which can GC.
+            let _guard = crate::OwnedProtect::new(self.sexp);
             let names_sexp = self.sexp.get_names();
             if names_sexp == SEXP::nil() {
                 return self;


CGMossa:
Why is the guard necessary here? If `self.sexp` was protected before the call, moving `self` into the method wouldn't drop that protection, so what is it actually guarding against?

<details>

## Summary
- The instinct about the move is right, but it rests on the wrong GC model. R's collector has no idea Rust owns `self.sexp`. It only scans its own roots: the `PROTECT` stack, the preserved-object list, and whatever is transitively reachable from those. A `SEXP` sitting in a `DataFrame` on the Rust stack is **not** a root just because Rust owns it. "Protected before" only holds if some external C-level root still points at it, and yes, a Rust move can't disturb that root. By-value vs by-reference `self` is irrelevant; neither touches the protect stack.
- The catch is the call site. These are consuming builder methods (see the "Post-assembly editing, absorbed from the old serde columnar assembler" region comment) meant to be chained on a freshly assembled DataFrame, e.g. `DataFrame::from_rows(..).rename(..).strip_prefix(..)`. That intermediate VECSXP is a brand-new allocation living only in a temporary Rust value: never `PROTECT`ed, never `R_PreserveObject`d, not yet returned to R. It has **no root**.
- Inside the loop, `names_sexp.set_string_elt(i, SEXP::charsxp(to))` allocates via `Rf_mkCharLenCE`, which can fire GC. With nothing rooting `self.sexp`, the collector is free to reclaim it, and `names_sexp` is `self.sexp`'s `names` attribute, reachable only through it, so it goes too. The STRSXP being written into gets freed mid-loop. `gctorture(TRUE)` would surface this if such a call site exists.
- Rooting `self.sexp` (rather than `names_sexp`) is sufficient because GC marks transitively: protecting the parent keeps its `ATTRIB` chain, including the names STRSXP, alive. One guard covers both. Same idiom `drop`/`select` already use a few lines down.
- Net: the guard is not restoring protection lost on the move. A method that allocates has to root whatever must survive the allocation unless it can prove every caller already did, and a `pub` consuming builder cannot prove that. The redundant case (caller already rooted it) costs one PROTECT/UNPROTECT pair; the unprotected case is a segfault under R-devel's GC.

---

</details>

PR #1094, opened 2026-06-28
fix(api): RHash returns a hex String (R-native); doc alloc_array 2-slot protect cost

CGMossa:
This started life as a doc-only PR noting that `RHash::hash()` can come back negative. Auditing that turned up the real problem: `i64` isn't faithfully representable in R at all, so the value was silently *lossy*, not just sign-flipped. So I've turned finding 1 into a fix — `hash()` now returns a hex `String` — rather than documenting a wart. The `alloc_array` 2-slot protect note (finding 2) is genuinely doc-only and rides along unchanged.

<details>

## Summary
- **Finding 1 — `RHash` now returns a hex `String` (was `i64`).** `hash() -> i64` routed through the lax `IntoR` path; a `DefaultHasher` output is almost always `> i32::MAX`, so it landed in R as an `f64` (REALSXP). An f64 has a 52-bit mantissa, so the low ~11 bits of a 64-bit hash were rounded away crossing into R — distinct Rust values **collide** in R, breaking the trait's stated jobs (dedup, environment keys). The negative-sign behavior the first revision documented was the harmless symptom of the same root cause (R has no unsigned/64-bit integer).
- Fix: `format!("{:016x}", hasher.finish())` — a 16-char lowercase hex string. All 64 bits survive, it's directly usable as an R environment key (env keys must be character anyway), and it works with `duplicated()` / `match()`. Hex over base64: marginal size difference for 8 bytes, and hex is the universal R hash idiom (`digest`, `rlang::hash`, `openssl`) with zero deps.
- **Finding 2 — `alloc_array` consumes two protect slots.** Unchanged from the first revision: `ProtectScope::alloc_array` allocates a dims INTSXP (1 slot) and the array (a second slot) but returns only the array's `Root`, so `scope.count()` rises by 2 while the caller sees one handle. Added a `# Protect-stack budget` doc section. Doc-only.

## Test plan
- [x] `cargo fmt --all` / `cargo check -p miniextendr-api` / `cargo clippy -p miniextendr-api --all-targets -- -D warnings` — clean
- [x] `cargo test -p miniextendr-api --lib test_rhash` — passes
- [x] `just rcmdinstall && just force-document` — succeeds; `NAMESPACE` and `man/` unchanged (only generated reference is the `Point$RHash$hash` alias)
- [x] `just devtools-test adapter-traits` — FAIL 0 | PASS 93

## Notes
- Updated the rpkg testthat expectation from `is.numeric` to `is.character`.
- Session-instability of `DefaultHasher` (random per-session seed) is inherent and orthogonal to representation; separately fixable by pinning a fixed-seed hasher if cross-session stability is ever wanted.

---

</details>

### DIFF
--- a/miniextendr-api/src/adapter_traits.rs
+++ b/miniextendr-api/src/adapter_traits.rs
@@ -112,20 +112,19 @@ impl<T: Display> RDisplay for T {
 ///
 /// # Methods
 ///
-/// - `hash()` - Returns a 64-bit hash as i64
+/// - `hash()` - Returns the 64-bit hash as a 16-character hex string
 ///
 /// # Note
 ///
 /// Hash values are deterministic within a single R session but may vary
 /// between sessions due to Rust's hasher implementation.
 ///
-/// Returned hashes may be **negative**. `DefaultHasher::finish()` yields a
-/// `u64`, which this trait returns as `i64` (R has no unsigned 64-bit type).
-/// The cast reinterprets the bit pattern, so any hash in `[2^63, 2^64)`
-/// surfaces in R as a negative integer. This is expected, not an error —
-/// the full 64-bit value is preserved; only the sign is a side effect of the
-/// reinterpretation. The sign bit is deliberately not masked off, as that
-/// would discard a bit of entropy.
+/// The hash is returned as a hex `String` (e.g. `"a1b2c3d4e5f60718"`), not a
+/// number. R has no faithful 64-bit integer type, so returning the `u64` as a
+/// numeric would round away its low bits above `2^53` — distinct Rust values
+/// would collide in R — and surface half the range as negative. A hex string
+/// preserves all 64 bits, is directly usable as an R environment key, and
+/// works with `duplicated()` / `match()` for deduplication.
 ///
 /// # Example
 ///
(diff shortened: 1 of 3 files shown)


CGMossa:
Returning `i64` from `RHash` isn't R-native — R has no faithful 64-bit integer, so the value this PR documents as "negative but full-precision" is actually *lossy* once it reaches R. I'd rather fix the root cause than document the wart, so I've switched `hash()` to return a hex `String` and pushed it onto this branch (`fb049975`). That makes the negative-values note this PR adds obsolete.

<details>

## Summary
- `hash() -> i64` returns through the lax `IntoR` path. A `DefaultHasher` output is almost always `> i32::MAX`, so `i64::into_sexp` falls to the `else` branch and lands in R as an `f64` (REALSXP). `into_r.rs` says this in its own words: values above `i32::MAX` land as an inexact `f64` and "IDs/counters larger than `2^53` start colliding."
- Consequence: an f64 has a 52-bit mantissa, so the low ~11 bits of a 64-bit hash are rounded away crossing into R. Distinct Rust values whose hashes differ only in the low bits become **equal** in R, breaking the trait's stated jobs (deduplication and environment keys). The negative-sign behavior this PR documents is the harmless symptom of the same root cause (no unsigned/64-bit int in R).
- Fix: return a 16-character lowercase hex `String` (`format!("{:016x}", hasher.finish())`). All 64 bits survive, it is directly usable as an R environment key (env keys must be character anyway), and it works with `duplicated()` / `match()` for dedup. Hex over base64: for a fixed 8-byte value the size win is marginal and hex is the universal R hash idiom (`digest`, `rlang::hash`, `openssl`) with zero dependencies.

## Notes
- The `alloc_array` 2-slot protect note (finding 2) is unrelated and stays.
- Updated the rpkg testthat expectation from `is.numeric` to `is.character`. `NAMESPACE` and `man/` are unchanged (the only generated reference is the `Point$RHash$hash` alias).
- Verified: `cargo fmt`/`check`/`clippy -D warnings` clean, `test_rhash` passes, and the adapter-traits R suite passes (FAIL 0, PASS 93) after `rcmdinstall` + `force-document`.
- Session-instability of `DefaultHasher` (random per-session seed) is inherent and orthogonal to representation; separately fixable by pinning a fixed-seed hasher if cross-session stability is ever wanted.

---

</details>

PR #1107, opened 2026-06-28
docs: add patches/templates.patch to the rebase-conflict regenerate rule

CGMossa:
Adds `patches/templates.patch` to the "regenerated artifacts" carve-out in the rebase-conflict rule (CLAUDE.md + AGENTS.md).

<details>

`patches/templates.patch` is a generated diff (the rpkg→templates delta). On rebase its hunks conflict, and hand-merging a diff-of-a-diff produces garbage — exactly what just happened rebasing #1089, where all three commits conflicted only on this file. The correct fix is to take either side and regenerate, not to reconcile hunks.

The rule already documented this for the binary vendor tarball (`rpkg/inst/vendor.tar.xz` → `just vendor`). This restructures that one bullet into a shared "regenerated artifacts — never hand-merge their hunks" carve-out and adds `patches/templates.patch` → `just templates-approve` (verified by `just templates-check`).

Docs-only; mirrored verbatim in both CLAUDE.md and AGENTS.md per the project rule to keep them in sync.

---
_Drafted by Claude (claude-opus-4-8)._

</details>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,7 @@ configure.ac eol=lf
 cleanup      eol=lf
 cleanup.ucrt eol=lf
 *.sh         eol=lf
+
+# Generated diff — never 3-way merge (produces diff-of-a-diff garbage on rebase).
+# On conflict, pick a side and regenerate with `just templates-approve`. See CLAUDE.md.
+patches/templates.patch -merge
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -256,7 +256,9 @@ Some agent sandboxes block compilation. For any compiling command (`just force-d
 - If agent didn't commit, commit in the worktree first.
 - Don't delete a worktree until its branch is pushed or merged.
 - **Clean up after push**: `git worktree remove -f -f <agent-worktree-path>` + `git worktree prune`. Each worktree holds a full `target/` (2–3 GB/agent).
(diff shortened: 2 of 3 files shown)


CGMossa:
Added the mechanical half: `patches/templates.patch -merge` in `.gitattributes`. Re-evaluated the other generated artifacts too — only the patch file qualifies.

<details>

## Summary
- The doc rule tells humans not to hand-merge the patch; the `.gitattributes` `-merge` attribute makes git skip the 3-way line merge that produces the diff-of-a-diff garbage in the first place. On conflict git leaves one side intact and marks it conflicted; you pick a side and regenerate with `just templates-approve`.
- Zero per-clone config (unlike a `merge=ours` driver, which needs `git config merge.ours.driver true`). `.gitattributes` ships in the repo, so it applies to everyone automatically.
- Re-evaluated the other regenerated artifacts: vendor.tar.xz, wasm_registry.rs, and wrappers.R are all gitignored (a `.gitattributes` entry would be a no-op). NAMESPACE and the 306 `man/*.Rd` are tracked and roxygen-generated, but deliberately left alone.

## Notes
- The patch file is uniquely dangerous because line-merging a unified diff yields output that looks structurally valid but is semantically corrupt (misaligned `@@` hunk headers), and nothing yells. NAMESPACE/`.Rd` botches are loud (conflict markers are syntax errors, `R CMD check`/roxygen fail), and those files usually merge cleanly when two branches add different exports. Marking 306 files `-merge` would kill those free auto-merges and force a manual pick + full regen every time, net negative. So only the diff got armored.

---

</details>

Issue #1118, opened 2026-07-02
macros: impl-block method-tag warning consts collide (E0428) across multiple #[miniextendr] impl blocks of one type


CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h4>AI-written details</h4></summary>

## Summary
- `strip_method_tags` / `strip_method_tags_r6` (`miniextendr-macros/src/roxygen.rs:689` and `:762`) emit one `#[deprecated] const _MINIEXTENDR_IMPL_METHOD_TAG_WARN_<Type>_<idx>: () = ();` per stripped block-level tag, with `<idx>` restarting at 0 for every impl block.
- If a type has two `#[miniextendr]` impl blocks that both carry block-level method-only tags (for example `@param` on the inherent impl doc and `@param` on a trait impl doc), the generated consts collide and the user gets `error[E0428]: the name _MINIEXTENDR_IMPL_METHOD_TAG_WARN_Foo_0 is defined multiple times` instead of the intended gentle deprecation nudge.
- Hit while writing the rand adapter fixtures for #1119: `@param` docs on both the `SeededRng` inherent impl and its `RRngOps` trait impl produced exactly this E0428. Moving the tags onto the methods (the correct placement) sidesteps it, but the failure mode is a confusing hard error for what is defined as a soft warning.
- Separate question worth checking while in there: the warning mechanism relies on `#[deprecated]` on a const that nothing ever references, and rustc only fires deprecation warnings on use. The nudge may never actually surface at all.

## Notes
- Minimal fix: include a per-block disambiguator (trait ident, or a monotonic per-expansion counter) in the const name. Call sites are `miniextendr_impl.rs:2259/2265`, `miniextendr_impl_trait.rs:298`, and `miniextendr_impl_trait/vtable.rs:147`.
- If the deprecated-const nudge indeed never fires, consider replacing it with a `compile_error!`-free diagnostic that actually reaches the user (a `#[deprecated]` item that the generated code references once, or proc-macro `Diagnostic` once stable).

---
_Drafted by Claude (claude-fable-5). Reviewed by the author._

</details>

Issue #1117, opened 2026-07-02
macros: trybuild UI snapshots mismatch locally (unexpected_cfgs warnings from rayon-gated DataFrameRow expansion)


CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h4>AI-written details</h4></summary>

## Summary
- `cargo test -p miniextendr-macros --test ui` fails locally on macOS with rustc 1.95.0: 7 of 132 trybuild fixtures mismatch their `.stderr` snapshots.
- The mismatch is `unexpected_cfgs` warnings appearing in the actual stderr but not in the snapshots, e.g. on `tests/ui/derive_dataframe_enum_nested_enum_b1_custom_tag_sibling.rs` and `tests/ui/derive_dataframe_enum_nested_enum_no_derive.rs`: `warning: unexpected cfg condition value: rayon ... expected values for feature are: coerce-default, doc-lint, r6-default, s7-default, strict-default, vctrs, and worker-default`.
- The warning comes from `#[derive(DataFrameRow)]` emitting `#[cfg(feature = "rayon")]`-gated items (`miniextendr-macros/src/dataframe_derive.rs:2108`, introduced with the non-rayon builder work in #1055/#1058) into a crate (the macros test crate) that has no `rayon` feature declared.
- Verified pre-existing on a pristine `origin/main` worktree, so it is not caused by any in-flight branch. CI's `rust-tests` job (`cargo test --workspace --locked --no-fail-fast`, added in #1110) is green on the same code, which makes this a toolchain/environment-dependent snapshot drift.

## Notes
- Regenerating the snapshots locally with `TRYBUILD=overwrite` would bake the warnings in and likely turn CI red, so that is not the fix.
- Candidate real fixes: have the derive emit the rayon-gated code in a way that does not trip `unexpected_cfgs` at the call site (for example route the parallel variant through a feature-gated helper in miniextendr-api instead of a raw `#[cfg(feature = "rayon")]` on emitted items), or declare a `rayon` feature on the macros crate for the test context.

---
_Drafted by Claude (claude-fable-5). Reviewed by the author._

</details>

PR #1119, opened 2026-07-02
feat(fixtures): rand/rand_distr adapter + expression subsystem coverage (audit A6)

CGMossa:

<details>

## Summary
- Ledger item 04 (`plans/2026-07-01-rand-expression-fixtures.md`, audit finding A6): two subsystems shipped in every rpkg build with zero exemplar coverage.
- **rand adapters** (`rpkg/src/rust/rand_adapter_tests.rs` + `test-rand-adapter.R`): `RRng` driven through `rand::RngExt` (`random`, `random_range`) under `#[miniextendr(rng)]`, with `set.seed()` determinism tests proving it consumes R's RNG; an `RDistributions` fixture mirroring the trait rustdoc; `SeededRng` (`RefCell<StdRng>`) exposed through `RRngOps` and `SeededNormal` through `RDistributionOps<f64>` as `#[miniextendr]` trait impls; `rand_distr::Normal` sampled through `RRng` with loose-tolerance moment checks on a fixed seed.
- **expression subsystem** (`rpkg/src/rust/expression_tests.rs` + `test-expression.R`): `expr_eval_str` (value round-trips, multi-expression source, empty input, `stop()` and parse failures surfacing as R errors, never a longjmp), `expr_call_builder` (`RCall` positional + named args), `expr_env_lookup` (`REnv::base_namespace()` plus the `r_eval_str` env parameter).
- **GC discipline (#430)**: `RCall` stores SEXP args in a `Vec` across allocations, so a no-arg `gc_stress_expression_call()` fixture drives the builder + eval path, with a gctorture harness test in `test-gc-stress-fixtures.R`.
- **API drift fixed by dogfooding** (the plan's "treat compile errors against the doc's examples as doc bugs" clause):
  - `RRngOps` / `RDistributionOps` were documented as `#[miniextendr]`-exposable but never wired into the trait ABI. Added `#[miniextendr]` to both trait declarations.
  - Re-exported `rand` from miniextendr-api under the `rand` feature (matching the existing `rand_distr` re-export) so downstream crates can name `RngExt` / `SeedableRng` at the trait-coherent version.
  - Rewrote the `rand_impl.rs` rustdoc examples: the old ones double-implemented the trait (E0119) and advertised nonexistent `r_`-prefixed R method names. The real convention is `obj$TraitName$method()` and the vtable requires every trait method implemented (`#[miniextendr(skip)]` to hide one from R).
  - `rand_distr::Normal::new` only rejects a non-finite sd (negative sd mirrors the distribution); fixture docs and tests pin that contract.
  - `docs/EXPRESSION_EVAL.md` never mentioned `r_eval_str` / `dollar_extract`; both documented now, and the `REnv` example fixed (unsafe fns, `package_namespace` returns `Result`).

## Test plan
- [x] `cargo check` on the rpkg crate with no features, `rand,rand_distr`, and `full`
- [x] Double-install loop for the new exports (`rcmdinstall`, `force-document`, `rcmdinstall`)
- [x] `just devtools-test`: FAIL 0, PASS 6607 (19 pre-existing environment skips)
- [x] New test files run, not skipped (rand/rand_distr are in configure's default feature set): 38 + 14 + 15 passes
- [x] gctorture test for `gc_stress_expression_call` green
- [x] `cargo fmt --all`, clippy_default, clippy_all, clippy_all_s7 (CI feature lists from ci.yml) all clean
- [x] `just test -- --skip ui` green across all four workspace legs; the skipped `ui` target fails identically on a pristine `origin/main` worktree (pre-existing, see #1117)
- [x] `just templates-check` clean

## Notes
- Deferrals tracked as issues: #1117 (local trybuild snapshot drift from rayon-gated DataFrameRow expansion, pre-existing on main, CI green), #1118 (impl-block method-tag warning const E0428 collision, found while placing `@param` docs on these fixtures).
- Per the ledger collision map, this PR adds exports and regenerates NAMESPACE/man, so it should merge one-at-a-time relative to #1116 and later export-adding items, regenerating wrappers per merge.

---

</details>


CGMossa:

<details>

Reviewed and rebased onto current main (over #1131, clean, no conflicts). Pushed as 6539b6ed. This one came back clean: no defects found, nothing to fix beyond the rebase.

## Review assessment

The coverage itself is what the ledger asked for, but the real value is the dogfooding fallout it forced into the same diff: `RRngOps` / `RDistributionOps` were documented as R-exposable for a long time while the trait declarations were never actually wired into the trait ABI, and the old rustdoc examples could never have compiled (double trait impl, E0119) while advertising `r_`-prefixed R method names that never existed. Examples written from intention rather than against the compiler. The rewritten examples now match the real convention (`obj$TraitName$method()`, every vtable slot implemented, `#[miniextendr(skip)]` to hide).

Details that hold up under scrutiny:

- `gc_stress_expression_call` is the right #430 shape: `RCall` holds its argument SEXPs in a `Vec` while later args and the `build()` cons-chain allocate, and the fixture protects its own args (per the builder contract) so the sweep exercises the builder's internal discipline, not the caller's. The `OwnedProtect` declaration order respects LIFO unprotect.
- The `set.seed()` determinism tests are the load-bearing ones: they prove `RRng` actually consumes R's RNG stream rather than an OS entropy source, which is the whole point of the adapter.
- The parse-failure tests (`"incomplete"`, `"syntax error"`) grep miniextendr's own error strings from `expression.rs`, not R's locale-translated parser messages, so they are locale-robust. Verified against the source.
- `#[miniextendr(rng)]` bracketing is applied exactly on the fixtures that touch R's RNG and deliberately absent on the `StdRng`-backed types; the module doc spells the distinction out.

## Local verification (post-rebase)

`cargo fmt --all --check`; clippy with `-D warnings` on miniextendr-api (`--features rand,rand_distr`), on the rpkg crate (`--features rand,rand_distr`), and on the workspace with default features: all clean. The R-side jobs (install, testthat, Sync Checks wrapper regeneration) were fully green on the pre-rebase head and the rebase only crossed #1131 with disjoint hunks, so CI on the new head is the confirming signal for the R surface.

## Optional follow-up (not blocking)

`SeededNormal$new(0, Inf, 1L)` panics through `.expect("sd must be finite")`, which the framework converts to an R error, but no R test pins that path; `rand_distr_normal` covers the same contract through its `Result`. A one-line `expect_error()` would close it if you want symmetric coverage.

</details>

Issue #1130, opened 2026-07-03
miniextendr-parallel skill: reference the concrete thread-control API once #1124 and this PR both land


CGMossa:
Follow-up from the parallelism-control-from-r plan (`plans/2026-07-02-parallelism-control-from-r.md`).

This PR ships `miniextendr_api::optionals::parallel` (`effective_threads()`, `ensure_pool()`, `set_threads()`) plus R-callable `miniextendr_num_threads()` / `miniextendr_set_threads()`, and documents the full precedence table in `docs/RAYON.md` ("Controlling Parallelism from R").

PR #1124 (end-user Claude skills, not yet merged) already ships a `miniextendr-parallel` skill at `minirextendr/inst/claude/skills/miniextendr-parallel/SKILL.md` with generic guidance ("honor an env var or R option for thread count... cap under `_R_CHECK_LIMIT_CORES_`"). Since the two PRs are sibling branches off `main` and neither currently contains the other's work, that skill can't reference the concrete API this PR introduces yet.

Once both PRs are merged, update that skill's thread-control section to point at `miniextendr_api::parallel::ensure_pool()` / the precedence table directly, per plan item 5.

PR #1131, opened 2026-07-03
feat(rayon): thread pool control, CRAN compliance, cgroup-aware defaults

CGMossa:
## Summary

Implements `plans/2026-07-02-parallelism-control-from-r.md` (F3, external-feedback ledger): rayon's compute patterns were solid but had no R-side knob for pool size, no CRAN `_R_CHECK_LIMIT_CORES_` compliance, and no documented cgroup-awareness story.

- New `miniextendr_api::optionals::parallel` module (under `feature = "rayon"`): `effective_threads()` resolves `MINIEXTENDR_NUM_THREADS` env > `RAYON_NUM_THREADS` env > `_R_CHECK_LIMIT_CORES_` cap (2 cores) > `available_parallelism()`. `ensure_pool()` builds the global rayon pool lazily and exactly once; if a global pool already exists (user code built one first), it's left alone.
- `ensure_pool()` wired into every `rayon_bridge` entry point (`with_r_vec`/`with_r_matrix`/`with_r_array`, `reduce::*`, `ParCollectR::collect_r`, `par_collect_sexp`, the `FromParallelIterator` impl, and a new `RParallelIterator::par_iter_ready()` choke point used by all ~25 default trait methods), the `RDataFrameBuilder` par fill path, and the serde columnar par path (`par_iter_to_dataframe[_growing]`).
- New R exports (rpkg, double-installed): `miniextendr_num_threads()` (reports the effective count, building the pool if needed) and `miniextendr_set_threads(n)` (sets `MINIEXTENDR_NUM_THREADS`; errors clearly if the pool is already built, since rayon can't resize it).
- `docs/RAYON.md`: new "Controlling Parallelism from R" section (precedence table, R-level knobs, containers/cgroups caveats, and a decision guide comparing rayon vs R-level parallelism (future/mirai) vs worker-thread + channel task queues).
- `minirextendr::use_rayon()` hint updated to point at the new control surface (no new scaffolding mechanism — none exists yet for shipping example fixtures into scaffolded packages, see plan 11 `scaffold-feature-passthroughs`, still TODO).
- Along the way: regenerated a stale `man/miniextendr_build.Rd` ("cdylib pass" → "wrapper-gen pass", matching already-updated roxygen source), and fixed root `CLAUDE.md`/`AGENTS.md` claiming a nonexistent third `clippy_all_s7` CI job (CI only runs `clippy_default` + `clippy_all`; `s7-default` is covered by `clippy_default`'s rpkg default path).

Deliberately not doing (per plan): cgroup cpu.weight/shares heuristics, per-call `num_threads` overrides / scoped pools, async runtimes.

Deferral: #1130 — the `miniextendr-parallel` end-user skill (PR #1124, not yet merged) should reference this PR's concrete API once both land.

## Test plan

- [x] `cargo test -p miniextendr-api --features rayon,serde --lib` — 336 passed
- [x] `just check` — clean across root workspace, cross-package fixtures, rpkg, cargo-revendor
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` (clippy_default) — clean
- [x] `cargo clippy --workspace --all-targets --locked --features full-codegen,rayon,...` (clippy_all) — clean
- [x] `cargo fmt --all` — clean
- [x] `just devtools-test` (full rpkg suite) — 6682 passed, 0 failed, 20 skipped
- [x] `just minirextendr-test` — 559 passed, 0 failed
- [x] Manual: `MINIEXTENDR_NUM_THREADS=3 Rscript -e 'library(miniextendr); miniextendr_num_threads()'` → 3; `_R_CHECK_LIMIT_CORES_=TRUE ...` → ≤2; unset → `available_parallelism()`; `miniextendr_set_threads()` after first use errors with a clear message


CGMossa:

<details>
<summary><h4>AI-written details: 8-angle review + CI rescue (81e75faa)</h4></summary>

## Review findings

An 8-angle review (3 correctness, 3 cleanup, altitude, conventions; candidates independently verified) found the design faithful to the plan but the enforcement unsound. Confirmed findings, most severe first:

- `test-rayon.R`: callr merges `env` over the inherited environment, so under `R CMD check --as-cran` the inherited `_R_CHECK_LIMIT_CORES_=TRUE` capped the `uncapped` baseline at 2 while the explicit-FALSE children uncapped. That one bug was all four red check legs (3 failures each).
- Derive-generated par paths (`from_rows_par`, `try_from_dataframe_par`, enum counterparts) never called `ensure_pool()`; the macros crate was untouched by the PR.
- `par_mean` and `par_variance` (hence `par_std_dev`) still called `self.par_iter()` directly, three methods below the `par_iter_ready()` doc claiming every default method uses it.
- Six raw-rayon rpkg fixtures plus `rayon_num_threads()` (whose `current_num_threads()` itself builds the default pool) bypassed the resolver, so rpkg's own test suite built an all-core pool before any wired path ran.
- `set_threads` guarded on a flag that only tracked `ensure_pool`, so any pool built via the holes above (or by user code) made it return success while permanently ineffective, the exact silent no-op the plan said must error.
- The CLAUDE.md/AGENTS.md edit claiming `clippy_all_s7` is a nonexistent third job is contradicted by this PR's own lint run (`clippy_all_s7: success`) and by ci.yml, whose comment calls that step the only compile gate for the s7 mutex side.
- The shared `\item{n}` on the merged `rayon_tests.Rd` page was clobbered to the threads wording, breaking `rayon_with_r_dataframe(n)`'s row-count doc.
- `skip_if_not_installed("callr")` at file top level mid-file stranded the 10 pre-existing tests below it whenever callr is absent.

## What 81e75faa fixes

- All four hole classes wired: `par_mean`/`par_variance` route through `par_iter_ready()`; the derive emits `ensure_pool()` at the top of each generated parallel region; all raw fixtures and `ParBuf::par_extend` call it; `rayon_num_threads()` resolves before reporting.
- `set_threads` now eagerly builds the pool with `n` through the same `Once` (building is the only exact probe rayon exposes) and errors on any pre-existing pool, including ones built outside miniextendr. The `unsafe env::set_var` channel and the redundant `POOL_BUILT` atomic are gone.
- Tests pin all three resolver env vars per scenario; the skip moved into the helper; the pure-validation test runs in-process; rustfmt fixed via a `match` (also drops an `as` cast).
- Doc revert for `clippy_all_s7`; RAYON.md updated for the new coverage claim and `set_threads` semantics; the two exports moved to `thread_control.rs` with their own Rd page, restoring the `n` param doc.

## Verification

- 337 api + 353 macros unit tests pass; all three CI clippy legs clean with `-D warnings`; the 7 trybuild mismatches are the known local-rustc E0599 wording drift (CI toolchain is authoritative).
- `test-rayon.R`: 123 pass, 0 fail, both with and without `_R_CHECK_LIMIT_CORES_=TRUE` in the calling environment.
- Fresh-session probes under `_R_CHECK_LIMIT_CORES_=TRUE`: first call `rayon_parallel_sum()` then `miniextendr_num_threads()` reports 2; same for `flat_basic_par()` and `gc_stress_dataframe_par_reader()` (derive paths); post-hoc `miniextendr_set_threads(4L)` errors instead of silently no-opping.

## Notes

- Deferrals filed per convention: #1132 (scaffold the thread-control wrappers into end-user packages, the PR's template-port scope cut had no tracking issue) and #1133 (post-review cleanups: callr spawn consolidation, `run_isolated` dedup, precedence-table duplication, `is_truthy` vs `backtrace.rs` truthiness polarity, `par_iter_ready` override-hardening, `RAYON_NUM_THREADS` parse divergence).
- Semantic change to note: `miniextendr_set_threads(n)` now builds the pool immediately rather than stashing an env var, so a second pre-use call errors instead of overwriting. This is what makes the already-built detection exact.

---

</details>

PR #1043, opened 2026-06-14
test(gctorture): backfill no-arg reader fixtures for skipped exports (#1026)

CGMossa:
Backfills no-arg `gc_stress_reader_*` fixtures for the non-scalar `FromDataFrame` reader shapes whose only exports are arg-taking `reader_*_roundtrip(df)` entrypoints — which the fast `gctorture(TRUE)` no-arg sweep skips. Writing the fixtures surfaced three reproducible use-after-frees that only the no-arg torture path could catch; all are fixed by rooting the synthesised frame SEXP, the same way the #807 enum-reader fixtures already do.

<details>
<summary>AI-generated detail (reviewed by me)</summary>

### What the sweep skips
The fast per-function `gctorture(TRUE)` sweep iterates `ls("package:miniextendr")` and calls each export with no arguments; any export with a required formal is skipped. The non-scalar `FromDataFrame` readers (`reader_fixed_roundtrip(df)`, the pinned/auto/Box column-expansion variants, mixed-type struct-flatten, Box/multi opaque list-columns, HashMap map-column, and the rayon `_par` variants) all take a `df` argument, so their production `from_dataframe` / `from_dataframe_par` read paths were never exercised on the quick path.

### Fixtures added (14)
Sequential: `gc_stress_reader_fixed`, `_pinned`, `_box_pinned`, `_auto`, `_auto_box`, `_flatten_mixed`, `_list_box`, `_list_multi`, `_hashmap`.
Parallel (`feature = "rayon"`): `gc_stress_reader_fixed_par`, `_flatten_par`, `_list_vec_par`, `_map_par`, `_int_map_par`.

Each synthesises 16 rows, writes them with `into_dataframe()`, **roots the frame SEXP with `OwnedProtect`**, reads it back with the sequential / parallel reader, and asserts a value-exact round-trip. The root is the load-bearing part: a Rust `DataFrame` is `Copy` and does not protect its SEXP, so under `gctorture(TRUE)` the reader's intermediate sub-frame / typed-vector allocations reclaim the parent frame mid-read. In production the frame is an R-rooted call argument; the explicit root stands in for that.

### Bugs surfaced
[...]

Issue #908, opened 2026-06-08
minirextendr helper: wire a local miniextendr framework checkout for standalone dev (configure-managed absolute [patch], source-mode only)


CGMossa:
## Summary

Documented in #823 (docs PR): depending on a **local miniextendr framework checkout** (`miniextendr-api` / `miniextendr-macros` / `miniextendr-lint`) from a **standalone** scaffolded package is currently a hand-rolled, error-prone `[patch]` edit. A `minirextendr` helper should manage it the way the monorepo's `configure` does automatically.

## Background

Inside the monorepo, `configure` walks up from the package dir, finds a sibling `miniextendr-api/Cargo.toml`, and — only in source mode — writes an **absolute-path** `[patch."https://github.com/A2-ai/miniextendr"]` block into the **configure-managed** `.cargo/config.toml` (gitignored, regenerated, never `Cargo.toml`). Tarball mode always wins over this override.

A standalone package has no equivalent. Users currently hand-edit `[patch]` in `src/rust/Cargo.toml` and hit two failure modes (see #823):

- **Symptom A** — a *relative* path dies under `R CMD INSTALL` of a built tarball, because the framework checkout is `.Rbuildignore`d and not staged.
- **Symptom B** — a `[patch]` coexisting with `[source] replace-with = "vendored-sources"` in tarball mode is fragile (patch + source-replacement on the same git source).

Working manual workaround: absolute path + source/dev mode (no `inst/vendor.tar.xz`).

## Proposed design

A helper, e.g. `use_local_miniextendr(path = "/abs/path/to/miniextendr")`, that:

1. **Absolute path only.** Canonicalize the supplied path; reject / warn on relative input. On Windows, normalize to forward slashes and strip the `\\?\` prefix (see CLAUDE.md "Windows paths in TOML").
2. **Writes to a configure-managed location, not `Cargo.toml`.** Mirror the monorepo: emit the `[patch."https://github.com/A2-ai/miniextendr"]` block into the configure-generated `.cargo/config.toml` (the standalone template's `configure.ac` already has monorepo-detection + a `[patch]` emit branch — extend it to accept an explicit sibling path, e.g. via a `tools/*.R`-recorded setting or a `miniextendr.yml` key, rather than only ancestor auto-detection). This keeps `Cargo.toml` / `Cargo.lock` / `.rs` un-mutated at configure time (CLAUDE.md: "`configure.ac` never mutates sources").
[...]

PR #1134, opened 2026-07-03
fix(minirextendr): base on.exit in repair_lock; single severity-gated doctor vendor check

CGMossa:
Two correctness bugs from the 2026-07-03 sonnet-5 audit (`audit/_worklist-2026-07-03.md` BUG5, BUG6; evidence in `audit/2026-07-03-dogfooding-minirextendr-r.md` findings #1 and #3).

## BUG5 — `withr::defer()` against a Suggests-only dep

`miniextendr_repair_lock()` (exported) called `withr::defer()` at `repair-lock.R:87` to restore `.cargo/config.toml` after a failed `cargo update`, but `withr` is Suggests-only. In practice withr is always installed transitively (usethis, a hard Import, imports it), so the `there is no package called 'withr'` crash was latent rather than reachable today — but an unguarded `::` into Suggests leans on a transitive dependency we don't control and breaks the day usethis drops it. It was the sole `withr::` call in `R/`; the package convention (documented in `create.R`) is base `on.exit()`.

Fix: `on.exit(..., add = TRUE)` drop-in with identical restore semantics. New regression test mocks a failing `cargo update` and asserts the backup is restored on the error path (the guarantee `withr::defer()` gave, now without the dependency).

## BUG6 — doctor double-counts the vendor tarball with contradictory severity

`miniextendr_doctor()` checked `inst/vendor.tar.xz` twice, unconditionally: a hard **fail** in "Install-mode signal" ("remove it") and a **warn** in "Vendor tarball" ("fine if mid-CRAN-prep") — the same file double-counted in the summary with contradictory guidance.

Fix: one check, in the "Vendor tarball" section, severity gated on a `.git` ancestor (via the existing `find_root_with_file()` helper):
- dev source tree (`.git` present) → **fail** — this is the latch leak (#441), with the `miniextendr_clean_vendor_leak()` remediation
- no `.git` ancestor → **warn** — likely CRAN-prep staging or an offline install

New tests pin warn-vs-fail in both contexts and that the tarball is reported exactly once.

## Testing

`just minirextendr-test`: `[ FAIL 0 | WARN 0 | SKIP 3 | PASS 620 ]`

### DIFF
--- a/minirextendr/R/doctor.R
+++ b/minirextendr/R/doctor.R
@@ -111,20 +111,18 @@ miniextendr_doctor <- function(path = ".", webr = FALSE) {
     }
   }
 
-  # -- Install-mode signal --
   # inst/vendor.tar.xz is gitignored and only belongs in the source tree
   # transiently (during `miniextendr_vendor()` + `R CMD build`). Its presence
-  # is resolved once here; the actual pass/warn/fail report happens in the
-  # single "Vendor tarball" check below (#BUG6 -- this used to be reported
-  # twice, with contradictory severity).
-  cli::cli_h2("Install-mode signal")
-
+  # is resolved once here (the #908 marker check below needs it); the actual
+  # pass/warn/fail report happens in the single "Vendor tarball" check below
+  # (#BUG6 -- this used to be reported twice, with contradictory severity).
   vendor_tarball_path <- tryCatch(
     usethis::proj_path("inst", "vendor.tar.xz"),
     error = function(e) NULL
   )
 
   # -- Local miniextendr override marker (#908) --
+  cli::cli_h2("Local override marker")
   # .miniextendr-local is a dev-only marker written by use_local_miniextendr().
   # It is gitignored and Rbuildignored; warn loudly if still present so the
   # developer doesn't accidentally ship a package that requires a local path.
@@ -155,6 +153,7 @@ vendoring or distributing this package."
       results$warn <- c(results$warn, ".miniextendr-local override active — run unuse_local_miniextendr() before distributing")
     }
   } else if (!is.null(local_marker_path)) {
+    cli::cli_alert_success("No {.path .miniextendr-local} override marker")
     results$pass <- c(results$pass, "no .miniextendr-local override")
   }
 
--- a/minirextendr/R/doctor.R
+++ b/minirextendr/R/doctor.R
@@ -111,38 +111,18 @@ miniextendr_doctor <- function(path = ".", webr = FALSE) {
     }
   }
 
-  # -- Stale vendor tarball check --
+  # -- Install-mode signal --
   # inst/vendor.tar.xz is gitignored and only belongs in the source tree
-  # transiently (during `miniextendr_vendor()` + `R CMD build`). If it is
-  # left behind after an interrupted build, configure switches to tarball mode
-  # and the post-build cleanup in Makevars.in deletes src/rust/.cargo/ from the
-  # source tree -- silently breaking the monorepo [patch] override on the next
-  # `miniextendr_build()`.
+  # transiently (during `miniextendr_vendor()` + `R CMD build`). Its presence
+  # is resolved once here; the actual pass/warn/fail report happens in the
+  # single "Vendor tarball" check below (#BUG6 -- this used to be reported
+  # twice, with contradictory severity).
   cli::cli_h2("Install-mode signal")
 
   vendor_tarball_path <- tryCatch(
     usethis::proj_path("inst", "vendor.tar.xz"),
     error = function(e) NULL
   )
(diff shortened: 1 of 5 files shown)


CGMossa:

<details>

Reviewed, rebased onto current main (trivial, one commit behind), and pushed one fix plus a body correction. Head is now 807e11be.

## What I fixed

**Empty section header left behind by the BUG6 restructure.** Moving the tarball report from "Install-mode signal" down to "Vendor tarball" left the `cli_h2("Install-mode signal")` with nothing to say in the healthy path: the only remaining content under it is the #908 local-override marker check, which prints nothing when the marker is absent. A healthy `miniextendr_doctor()` run rendered a bare header followed by silence. Renamed the header to "Local override marker" (which is what the section now contains), moved the `vendor_tarball_path` computation out from under any header (it is shared state for two later checks, not a report of its own), and gave the marker-absent path a visible success line like every other section. No test pinned the old header text.

**Overstated severity claim in the body (corrected in place).** BUG5's framing said any user without withr hit `there is no package called 'withr'`. That crash was latent, not reachable: minirextendr hard-Imports usethis, and usethis Imports withr, so withr is always installed transitively. The fix is still clearly right — an unguarded `::` into a Suggests-only package leans on a transitive dependency we don't control, and it contradicts the package's own documented base-`on.exit()` convention — but the audit-to-PR pipeline copied the audit's severity without checking the dependency graph.

## What holds up

- The `on.exit(..., add = TRUE)` swap is ordering-safe. `with_project()` registers its project-restore on the same frame (prepended, via usethis), so the new appended handler runs after the project restore — but `cargo_cfg` / `cargo_cfg_bak` are absolute paths captured before registration, so the restore does not depend on the active project. Verified by reading, and the new mocked-failure regression test pins exactly the guarantee that matters (restore on the error path).
- The `.git`-ancestor severity gate is the right heuristic: it is the same signal `configure` itself uses for auto-vendor, so doctor's warn-vs-fail now agrees with the build system's own dev-vs-staging distinction. And `find_root_with_file()` uses `file.exists()`, which is TRUE for the `.git` *file* in git worktrees, not just the directory — so worktree checkouts classify as dev trees correctly.
- The five new doctor tests cover absent/warn/fail plus exactly-once in both contexts — the exactly-once assertions are the ones that would catch a regression to double-counting.

## Verification

`just minirextendr-test` post-rebase, post-fix: `[ FAIL 0 | WARN 0 | SKIP 3 | PASS 620 ]` — identical to the PR's baseline.

</details>

CGMossa:

<details>

The red "R CMD check / Linux devel" on the rebased head is not this PR's doing: this diff touches only `minirextendr/`, but the devel job also runs rpkg's testthat suite, which segfaulted with `*** recursive gc invocation` spam (R CMD check preserves only those last 13 lines, so the crashing test is not identifiable from the log). The same rebased-head reruns hit a related failure on #1043 and #1044: `gc_stress_typed_dataframe` panicking under gctorture because `List::from_raw_pairs` drops its protection on return and the fixture let `as_data_frame()` allocate against the unprotected container (fix: #1145). A use-after-free of that kind manifests as a clean panic or a segfault depending on what reuses the node, so the devel crash is plausibly the same bug on a different schedule. Suggest merging #1145 first, then re-running this PR's failed job; if devel still segfaults after that, it is a distinct R-devel-only bug worth its own issue.

</details>

CGMossa:

**Diagnosis.** Both failing checks were the same latent `main` bug, not this (minirextendr-only) PR:

- **Linux release**: `test-gc-stress-fixtures.R:143` — `gc_stress_typed_dataframe iteration 3: DataFrame always carries a names attribute`.
- **Linux devel**: the same fixture, surfacing as a crash — `*** recursive gc invocation` spam then `checking tests ... ERROR`.

The branch was already current with `main` (0 commits behind), so merging `origin/main` changes nothing. This is the unrooted-container UAF documented in PR #1145's review file: `gc_stress_typed_dataframe` builds a synthetic container via `List::from_raw_pairs`, whose internal `OwnedProtect` guards drop on return, leaving the VECSXP unprotected while `as_data_frame()` allocates the class/`row.names` attributes and `TheophDf::try_from_sexp` validates. Under gctorture the node is reaped; depending on what reuses the cell you get the clean "names attribute" assert (release) or a "recursive gc invocation" segfault (devel) — same one bug.

**Fix.** Root both halves of the fixture into their existing `ProtectScope` before the next allocation:

```rust
unsafe { scope.protect_raw(list.as_sexp()) };
...
unsafe { scope2.protect_raw(list2.as_sexp()) };
```

Byte-identical to PR #1145 (still open). Applied here rather than blocking on that PR landing, since merging `main` can't pick up an unmerged fix and a bare re-run only passes by allocation-schedule luck. Identical hunks merge cleanly once #1145 lands.

Pushed as `115c4f2e`.

</details>

Issue #1139, opened 2026-07-04
R_MAIN_THREAD_ID OnceLock blocks safe in-crate unit tests of new R-engine-dependent code


CGMossa:
## Context

Discovered while adding direct unit-test coverage for `charsxp_to_string_lossy`
(D3 of the 2026-07-03 dogfooding audit, `audit/_worklist-2026-07-03.md`).

`worker.rs`'s `R_MAIN_THREAD_ID` is a process-global, set-once `OnceLock<ThreadId>`
(`worker.rs:69`). `worker::tests` (the crate's own `#[cfg(test)]` module) calls
`miniextendr_runtime_init()` from several `#[test]` fns to exercise that
registration mechanism itself — those tests are written to tolerate "another
test already claimed it" (see the comments at `worker.rs:792`, `worker.rs:833`,
`worker.rs:907`: "If another test already called miniextendr_runtime_init...
skip").

That tolerance assumes no *other* test in the same `--lib` binary needs
`R_MAIN_THREAD_ID` to correctly and durably point at a real, live R engine
thread. That assumption broke the first time I tried to add one: a new
`#[cfg(test)] mod tests` in `from_r.rs` that spun up its own dedicated R
worker thread (mirroring `tests/r_test_utils.rs`'s pattern, but in-crate
because the target function was `pub(crate)`). When run in isolation
(`cargo test -p miniextendr-api --lib from_r::tests`), all 4 tests passed.
When run as part of the full `--lib` suite (`cargo test -p miniextendr-api
--lib`, 309 tests, default parallelism — 14 cores on the box I hit this on),
2 of the 4 consistently failed with:

```
with_r_thread called from a non-main thread without the `worker-thread` feature.
```

Root cause: one of `worker::tests`'s own test threads won the
`R_MAIN_THREAD_ID.set()` race first (registering *its own, throwaway* thread
ID), so my dedicated, live R-engine thread's later
`miniextendr_runtime_init()` call became a silent no-op (`OnceLock::set`
returns `Err` and the code does `let _ = ...set(...)`, discarding it). Every
subsequent checked FFI call from my (correct, live) R thread then saw
`is_r_main_thread() == false` and, since the test wasn't compiled with the
`worker-thread` feature, panicked instead of dispatching. The panic message
itself never surfaced in captured test output either (a second, minor gotcha
— see below), which made this much harder to diagnose than it should have
been.

This is architecturally different from the existing `tests/*.rs` integration
[...]

Issue #1138, opened 2026-07-04
re-add refcount_protect arena flavors (HashMap/ahash-backed) when a consumer appears


CGMossa:
## Context

Part of the D2 cleanup from the 2026-07-03 dogfooding audit
(`audit/2026-07-03-dogfooding-runtime-error-gc.md` finding #1,
`audit/_worklist-2026-07-03.md` D2). `miniextendr-api/src/refcount_protect.rs`
carried 6 GC-protection arena flavors ({BTreeMap, HashMap} x {RefCell,
thread-local} x {SipHash, ahash}) behind a generic `MapStorage` trait, but
repo-wide only `RefCountedArena` (BTreeMap + RefCell) had any consumer (a
single rpkg test fixture). The variant explosion — `HashMapArena`,
`ThreadLocalHashArena`, `FastHashMapArena`, `ThreadLocalFastHashArena`, the
`MapStorage` trait, `impl_hashmap_map_storage!` macro, `ThreadLocalArenaOps`
trait, `define_thread_local_arena!` macro, and the `refcount-fast-hash`
feature (+ `ahash` dependency) — was removed in favor of two concrete,
non-generic types: `RefCountedArena` (RefCell-backed) and `ThreadLocalArena`
(thread-local-backed), both BTreeMap-based.

## What was removed

- `HashMapArena`, `ThreadLocalHashArena` — HashMap-backed variants (SipHash).
- `FastHashMapArena`, `ThreadLocalFastHashArena` — ahash-backed variants,
  behind the `refcount-fast-hash` feature.
- `MapStorage` trait + `impl_hashmap_map_storage!` macro — the map-type
  abstraction that only existed to support >1 map flavor.
- `ThreadLocalArenaOps` trait + `define_thread_local_arena!` macro — the
  generic thread-local-arena-generator machinery (only one instantiation,
  `ThreadLocalArena`, survives; it's now a concrete struct with inherent
  methods, not a macro-generated trait impl).
- `RefCountedGuard` — a "legacy type alias for backwards compatibility" with
  zero consumers (dead per the project's no-backwards-compat policy).
- The `refcount-fast-hash` Cargo feature and the `ahash` dependency in
  `miniextendr-api/Cargo.toml` (also removed the forwarding feature in
  `rpkg/src/rust/Cargo.toml`, the `miniextendr-bench` forwarding feature, and
  all doc/CI/justfile mentions).

## What to do when a real consumer appears

[...]

PR #1140, opened 2026-07-04
refactor(api): collapse refcount_protect to used flavors; unify CHARSXP lossy readers

CGMossa:
Dogfooding cleanup for D2 and D3 from the 2026-07-03 audit: `refcount_protect.rs` had six GC-protection arena flavors behind a generic map-storage abstraction but only one flavor ever had a real consumer, and three CHARSXP-to-`String`/`str` helpers existed with a dead public one and ~7 hand-rolled reimplementations of it scattered around the crate. Both get collapsed here; no R-visible behavior change is intended beyond the deliberate NA-handling tightenings called out below.

<details>

## Summary

**D2 — `refcount_protect.rs`: 6 arena flavors → 2** (`audit/_worklist-2026-07-03.md` D2, evidence in `audit/2026-07-03-dogfooding-runtime-error-gc.md` finding #1)
- Removed `HashMapArena`, `ThreadLocalHashArena`, `FastHashMapArena`, `ThreadLocalFastHashArena`, the `MapStorage` trait + `impl_hashmap_map_storage!` macro, the `ThreadLocalArenaOps` trait + `define_thread_local_arena!` macro, and the dead `RefCountedGuard` alias. Kept `RefCountedArena` (RefCell-backed) and `ThreadLocalArena` (thread-local-backed) as concrete, non-generic structs — both BTreeMap-based, matching the single flavor that ever had a consumer (an rpkg test fixture).
- Removed the now-unused `refcount-fast-hash` Cargo feature and the `ahash` dependency from `miniextendr-api/Cargo.toml`, and pulled that thread through everywhere it was forwarded or mentioned: `rpkg/src/rust/Cargo.toml`'s feature forward (this one was load-bearing — leaving it in place would have broken `cargo check` on `rpkg/src/rust` with an "unknown feature" error), `rpkg/tools/detect-features.R` + its two scaffolding-template mirrors + `minirextendr/R/feature-detect-configure.R`'s generator source, the scheduled `feature-legs` CI job matrix, `justfile`'s `bench-features`/`bench-full` recipes, and several docs/README feature tables.
- `miniextendr-bench/benches/refcount_protect.rs` (958 → 559 lines) and its doc-comment plan lost the benchmark groups for the deleted flavors.
- `refcount_protect.rs` itself: 1150 → 855 lines.
- Tracking issue for re-adding a flavor when a real consumer shows up: #1138.

**D3 — CHARSXP→`String`/`str`: adopt the dead safe reader, add its owned sibling** (`audit/_worklist-2026-07-03.md` D3, evidence in finding #2)
- `SexpExt::r_char_str` (public, safe, zero callers) is now used at its two literal drop-in sites: `vctrs.rs:376` (`new_rcrd`'s field-name extraction) and `encoding.rs:74` (the UTF-8-locale-assertion helper's element-name comparison).
- Added `from_r::charsxp_to_string_lossy(SEXP) -> Option<String>` (NA/nil-defensive, `to_string_lossy` for non-UTF-8) next to `charsxp_to_str`, and routed the five owned-extraction sites through it: `s4_helpers::s4_class_name`, `expression::get_r_error_message`, `IndexMap<String, T>::try_from_sexp`, and jiff's 4 identical `tzone` attribute reads (`Timestamp`/`Zoned`/`Vec<Zoned>`/`Vec<Option<Zoned>>` `TryFromSexp` impls).
- Behavior preserved at every site's existing decision point (empty-name fallback, no-tzone-attribute fallback, etc.) with one deliberate, consistent tightening: an `NA_character_`/nil input now maps to the site's existing "no value" fallback (`None` / auto-generated `V<n>` name / UTC-default timezone) instead of literally reading `NA_STRING`'s C-string value (`"NA"`) as if it were real data. At the IndexMap site specifically, non-UTF-8 bytes are now lossily decoded (replacement chars) rather than discarded to an auto-generated name — this is the one place the plan's "add lossy owned reader" contract actually changes the *content* of an edge case, not just how it's computed.

## Test plan
- [x] `just check`, `just test`, `just clippy` — all pass except the pre-existing, documented `miniextendr-macros --test ui` trybuild snapshot mismatch (local rustc 1.95 vs. CI's `dtolnay/rust-toolchain@stable`; CLAUDE.md already says this is expected and not to overwrite snapshots locally). Nothing in this diff touches `dataframe_derive.rs` or its UI fixtures.
- [x] `cargo fmt --all` — clean.
- [x] All three CI clippy variants reproduced locally, `-D warnings`: default (`clippy_default`), `full-codegen` + integration list (`clippy_all`), `full-codegen-s7` + integration list (`clippy_all_s7`).
- [x] `cargo test -p miniextendr-api --test refcount_protect` — all 16 pre-existing `RefCountedArena` tests pass unmodified against the rewritten module.
- [x] New integration test `indexmap_degenerate_names_use_charsxp_to_string_lossy_fallbacks` (`miniextendr-api/tests/indexmap.rs`) hand-builds a list with an NA name, an empty name, a valid name, and a raw non-UTF-8 (`CE_BYTES`) name, and asserts `charsxp_to_string_lossy`'s exact behavior at each.
- [ ] rpkg testthat suite — per the plan, this is cargo-only work; the coordinator will run the R-side suite at the review checkpoint (shared install-slot serialization this session). `just rcmdinstall` / `R CMD INSTALL` were intentionally not run here.

## Notes
- Attempting a *direct* in-crate unit test for `charsxp_to_string_lossy` (NA / nil / valid-UTF-8 / invalid-UTF-8, as the plan asked for) surfaced a real, pre-existing test-isolation hazard: `worker.rs`'s process-global `R_MAIN_THREAD_ID` `OnceLock` races against any other `--lib`-target unit test that needs a live R engine on its own thread, because `worker::tests` itself claims that global from disposable test threads and is only "safe" for tests that don't need it to stay valid afterward. It reproduced consistently under default test parallelism (14 cores) and vanished at `--test-threads=4` or in isolation, which is the kind of failure that looks like flaky CI rather than a structural problem. Filed as #1139 with the root cause and remediation options; the coverage for this PR instead lands as the integration test above (a separate process, so it doesn't hit the hazard).
- `rpkg/configure` was regenerated via `autoconf` after the `configure.ac` edit, and `patches/templates.patch` was regenerated via `just templates-approve` (verified with `just templates-check`) after the `detect-features.R` edits — both per the standard template-sync workflow, not hand-patched.
- `rust-llm-docs/generated/miniextendr-api.md` (a periodically-regenerated LLM-doc snapshot) got a minimal one-line fix for the now-removed feature-table entry; the deeper staleness from deleted arena-type doc blocks elsewhere in that file is left for its next full regeneration pass (it's not CI-gated and the tool requires a heavier nightly-rustdoc-JSON regeneration step out of scope here).

---
_Drafted by Claude (claude-sonnet-5). Reviewed by the author._

</details>


CGMossa:
Coordinator follow-up per the test plan: rpkg testthat suite run against this branch after `just configure && just rcmdinstall` — `[ FAIL 0 | WARN 0 | SKIP 20 | PASS 6805 ]`. R-side gate green. (Two earlier install attempts failed due to a concurrent session racing the shared rv library — unrelated to this diff.)

Issue #1141, opened 2026-07-04
Trait-method emitter: unify per-class namespace shapes + Self-return re-wrapping


CGMossa:
## Context

Follow-up to `refactor/trait-method-emitter` (branch), which implements the
shared `TraitMethodContext` for trait-impl R-wrapper codegen
(`miniextendr-macros/src/miniextendr_impl_trait/method_context.rs`). That PR
lands responsibilities 1+2 from `plans/2026-07-04-audit-trait-method-emitter.md`
(structured receiver-ptr extraction fixing the S4/S7/R6 `str::replace`
substring-corruption bug, and the full 6-step method prelude / `Missing<T>`
forwarding). Responsibilities 3+4 are deferred here per the plan's explicit
fallback clause, to keep that PR reviewable.

Source: `audit/2026-07-03-dogfooding-macros-codegen.md` finding #1
("Recommendation for #1"), `audit/_worklist-2026-07-03.md`.

## 3. Per-class namespace-shape unification

The 5 trait generators (`miniextendr_impl_trait/r_wrappers.rs`) each
independently decided the R-visible spelling for static trait-method access,
and they disagree in ways that aren't all functionally motivated:

- Env / S3 / R6 **static** methods: `Type$Trait$method` (all three agree).
- S4 static methods: flat `Type_Trait_method()` standalone function —
  legitimate divergence (S4 objects intercept `$<-`, so `Type$Trait$method`
  can't be assigned onto them).
- S7 static methods: `attr(Type, "Trait")$method()` via a local env attached
  through `attr()` — also a legitimate divergence for the same `$<-`
  interception reason, but a *different* workaround shape than S4's.
- **R6 instance vs static within the same class system**: instance methods
  are exported as standalone `r6_trait_<Trait>_<method>(x)` functions, but
  static methods use `Type$Trait$method()` — i.e. R6's own two branches
  disagree with each other, and there's no functional reason for it (both
  blocks could use the same shape); it's just two independently hand-rolled
  code paths making different arbitrary choices.

None of the 5 generators share a "put a static method into the per-class
trait namespace" helper — each re-decides the shape inline. Fix: define one
namespace-shape policy per class system (keeping the S4/S7 `$<-`-interception
workarounds, since those are real constraints) and route every generator
through it via `TraitMethodContext`, closing the R6 instance/static
inconsistency in the same pass. This is a **user-visible breaking rename**
[...]

PR #1142, opened 2026-07-04
refactor(macros): shared trait-method emitter — fixes substring receiver corruption and prelude parity

CGMossa:
## Summary

Unifies trait-method R-wrapper codegen behind a shared `TraitMethodContext`
(`miniextendr-macros/src/miniextendr_impl_trait/method_context.rs`), mirroring
the `MethodContext` layer the 6 inherent-impl class generators already share.
Closes two live correctness bugs plus the architecture finding, per
`audit/2026-07-03-dogfooding-macros-codegen.md` finding #1,
`audit/_worklist-2026-07-03.md` (BUG1, BUG2, architecture-level finding), and
`plans/2026-07-04-audit-trait-method-emitter.md`.

- **BUG1 (receiver-ptr substring corruption)**: the S4/S7/R6 trait generators
  built the `.Call()` invocation assuming `self = "x"`, then ran
  `call.replace(", x", ", .ptr")` to swap in the real receiver expression.
  `str::replace` rewrites *every* match of that substring, so a parameter
  whose R name starts with `x` (e.g. `x_factor`) was silently corrupted into
  `.ptr_factor` — a runtime `object '.ptr_factor' not found` error.
  `TraitMethodContext::instance_call(self_expr)` passes the receiver
  expression directly to `DotCallBuilder::with_self` (mirroring
  `MethodContext::instance_call`), so no string surgery happens at all. All
  three `.replace` sites are gone.
- **BUG2 (prelude parity)**: trait methods skipped the
  `precondition_checks` (`stopifnot()`) and `match_arg_prelude`
  (`base::match.arg()`) steps that inherent methods get, and built `.Call()`
  arguments via `collect_param_idents`, which has no `Missing<T>` handling (a
  truly-missing R argument forwarded as a bare binding errors on lookup, per
  #1129). `TraitMethodContext::emit_method_prelude` gives all 5 trait
  generators the same prelude steps as `MethodContext::emit_method_prelude`,
  and `.Call()` args now go through `build_r_call_args_from_sig` (the same
  function the inherent path uses), which forwards `Missing<T>` inline.
- **New attribute surface needed to make BUG2 meaningful**: trait methods
  previously had *no* `match_arg`/`choices` attribute support at all
  (`TraitMethod` carried no per-param map). Added
  `choices(...)`/`choices_several_ok(...)` parsing, mirroring the
  inherent-impl parser (`split_choice_list` moved to `r_wrapper_builder.rs`
  and shared between the two parsers so they can't drift). While wiring this
  up, found and fixed a second live bug: `base::match.arg(arg)` (no explicit
  `choices=`) reads its choice list from the *formal default* of the calling
  function's parameter — `TraitMethodContext` initially built formals from
  the raw `param_defaults` map, so `match.arg()` failed at runtime with
  `argument is missing, with no default` even when the caller passed a valid
  value. `effective_r_defaults` (`r_class_formatter.rs`) is now a shared free
  function both `MethodContext` and `TraitMethodContext` use.
- **Design chosen**: a parallel `TraitMethodContext` (not direct
  `MethodContext` reuse) — `TraitMethod` (flat fields) and `ParsedMethod`
  (`MethodAttrs` bag) have different shapes, so forcing structural
  unification would ripple through all 6 inherent generators for a much
  larger, riskier diff. Instead, `TraitMethodContext` shares the actual
  **computation** with `MethodContext` via new free functions in
  `r_class_formatter.rs` (`build_match_arg_prelude`, `match_arg_skip_set`,
  `build_method_precondition_checks`, `effective_r_defaults`,
  `match_arg_doc_placeholder_map` now keyed on `per_param`), so future fixes
  to prelude/formal-default logic can't drift between the two paths again.
  Per the audit's finding #1(b), adapter traits (`RDebug`/`ROrd`/`RClone`)
  share this same code path, so they are fixed for free.

## Deferred scope (fallback per the plan)

Per `plans/2026-07-04-audit-trait-method-emitter.md`'s explicit fallback
clause, plan responsibilities 3+4 are **not** in this PR:

- **Per-class namespace-shape unification** (the 5 divergent static-method
  spellings, including R6's internally inconsistent
  `r6_trait_<Trait>_<method>` instance vs `Type$Trait$method` static split)
  — a user-visible breaking rename across all 5 generators and every
  downstream fixture.
- **Self-return re-wrapping** for trait factory methods (`clone`/`default`/
  `from_vec` currently return a bare unclassed ExternalPtr) — needs the
  constructor-ptr-wrapping logic factored out of the inherent S4/S7/R6
  generators first.

Both are tracked with full design context in #1141.

Also descoped there: `match_arg` (choices derived from a `#[derive(MatchArg)]`
enum's `CHOICES` const) is not yet exposed as a trait-method attribute — it
additionally needs the `__match_arg_choices__<param>` C-wrapper helper +
`MX_MATCH_ARG_CHOICES` registration that only the inherent path emits
(`generate_method_match_arg_helpers` in `miniextendr_impl.rs`); accepting it
today would ship an unresolvable write-time placeholder as the formal
default. `choices(...)` (literal list, no derivation) is fully supported and
covered end-to-end. Documented at the attribute-parser site in `vtable.rs`.

## Tests

- **New unit tests** (`miniextendr_impl_trait/tests.rs`): `x_factor`
  receiver-corruption regressions across S4/S7/R6 (generic body + S7
  fast-path shortcut), precondition checks, `choices` formal default +
  `match.arg()` prelude, `Missing<T>` inline forwarding.
- **New rpkg fixture + testthat** (`rpkg/src/rust/class_system_matrix.rs`
  `Scaler` trait + `ScalerS4`/`ScalerS7`/`ScalerR6`;
  `rpkg/tests/testthat/test-trait-method-emitter.R`): BUG1 + BUG2 exercised
  at runtime, 16 assertions, all passing. The runtime tests caught the
  `match.arg()` formal-default bug the codegen-string tests missed.
- `cargo test -p miniextendr-macros`: 358 lib tests pass. UI suite: the 7
  pre-existing `derive_dataframe_*` local-rustc-drift mismatches documented
  in CLAUDE.md remain untouched; 1 legitimate `.stderr` update for the
  extended trait-method attribute list (`choices`, `choices_several_ok`).
- `just check` (root workspace + cross-package consumer/producer + rpkg +
  cargo-revendor): clean.
- `just devtools-test` (full rpkg suite): **FAIL 0 | WARN 0 | SKIP 20 |
  PASS 6821** (all skips are pre-existing feature gates).
- `just cross-install && just cross-test` (mandatory trait ABI coverage):
  **producer.pkg 52 PASS / 0 FAIL, consumer.pkg 335 PASS / 0 FAIL**.
  Regenerated tracked cross-package wrappers.R committed (they now carry the
  stopifnot preludes on trait methods).
- All three CI clippy variants reproduced locally with `-D warnings`
  (`clippy_default`; `clippy_all` = `full-codegen` + integration list;
  `clippy_all_s7` = `full-codegen-s7` + integration list): clean.
- `cargo fmt --all`: clean. NAMESPACE/man regenerated via
  `just rcmdinstall && just force-document` and committed in sync.

Do not merge — awaiting review.


CGMossa:

The `R CMD check / Linux release` job was red for two independent reasons; both are fixed in `baebe4a5`.

**1. Rd `\usage` WARNING (real regression from this refactor).**

```
Undocumented arguments in Rd file 'ScalerR6.Rd'
  ‘x_factor’ ‘amount’ ‘mode’
Documented arguments not in \usage in Rd file 'ScalerS7.Rd':
  ‘"slow")’
```

Both trace to how the trait-method emitters auto-document formals. The S7 fast-path shortcut and the shared `MethodContext` auto-`@param` loop derived parameter names with `formals.split(", ")`, which shreds a `match_arg` default like `mode = c("fast", "slow")` into two bogus formals (`mode = c("fast"` and `"slow")`) — hence the spurious `@param "slow")`. Separately, the R6 standalone trait functions (`r6_trait_Scaler_scale(x, x_factor)` etc.) emitted `@param x` only, never documenting the method's own formals — hence the "Undocumented arguments" list.

Fix: a shared paren/quote-aware splitter `split_r_formals` (+ `formal_name`) in `roxygen.rs`, unit-tested for the `c("fast", "slow")` and `list(a=1, b=2)` cases. The S7 shortcut loop and the shared `MethodContext` loop now route through it, and the R6 trait wrapper gets an equivalent auto-`@param` loop. Regenerated `man/ScalerR6.Rd` (adds the three params) and `man/ScalerS7.Rd` (drops `"slow")`).

**2. `gc_stress_typed_dataframe` gctorture failure** (`iteration 2: DataFrame always carries a names attribute`) — the latent unrooted-container UAF from PR #1145. Rooted both halves of the fixture into their `ProtectScope`; identical hunks to #1145 merge cleanly.

**Notes on what I deliberately did *not* commit:**
- `devtools::document` (load_all-based) also flipped ~35 `derived_*` vctrs entries in `NAMESPACE` from `S3method(...)` to `export(...)`. That is a load_all-vs-full-install S3-generic-detection artifact, not a product of the `@param` change (which cannot affect exports), and this branch's `wrappers-sync-check` was already green — so the committed `NAMESPACE` matches CI's full-install regeneration. I reverted it to avoid breaking sync-check.
- `rpkg/src/rust/Cargo.lock` (configure stripped the `git+` source lines) — reverted.

Verified: `cargo test -p miniextendr-macros` (new splitter tests pass), `cargo clippy -p miniextendr-macros --all-targets --locked -- -D warnings` clean, `cargo fmt --all --check` clean. The `ui`/trybuild mismatches present locally are the known rustc-1.95-vs-CI-stable `derive_dataframe_*` snapshot drift (unrelated files); snapshots left untouched since CI's toolchain is authoritative.

Pushed as `baebe4a5`.

</details>

Issue #1143, opened 2026-07-05
Batched diagnostics for string-parse vector conversions (uuid/url/regex/num-bigint)


CGMossa:
## Context

PR (refactor/api-str-parse-tryfromsexp branch, audit D4) deduplicates the ~20
hand-written string-parse `TryFromSexp` impls across
`optionals/{uuid,url,regex,num_bigint}_impl.rs` into one
`try_from_sexp_via_str_parse!` macro (`miniextendr-api/src/from_r.rs`).

The macro preserves the pre-existing bail-on-first-failure behavior for
`Vec<T>` / `Vec<Option<T>>`: the first NA or parse failure aborts the whole
conversion with a single indexed error.

## The gap

CLAUDE.md's project principle says: "Collect all errors in vectorized ops:
don't bail on first failure — give users one batched diagnostic." None of the
string-parse vector conversions ever did this (pre- or post-macro). With the
macro in place, batching is now a one-site change instead of eight.

## Sketch

In the macro's `Vec<T>` / `Vec<Option<T>>` arms, replace the short-circuiting
`.collect::<Result<_,_>>()` with a partition that accumulates
`(index, error-string)` pairs and, when non-empty, returns a single
`SexpError::InvalidValue` listing all failures (capped at some reasonable
count, e.g. first 10 plus "and N more"). Error-message shape change must be
coordinated with any rpkg testthat expectations (today only "invalid UUID" is
grepped, in rpkg/tests/testthat/test-feature-adapters.R:61).

Consider whether the same batching should extend to the numeric-parse vector
conversions elsewhere in from_r.rs while at it.

PR #1144, opened 2026-07-05
refactor(api): try_from_sexp_via_str_parse! — dedupe string-parse TryFromSexp across uuid/url/regex/num_bigint

CGMossa:
Audit D4 (`audit/_worklist-2026-07-03.md`, evidence `audit/2026-07-03-dogfooding-optionals-integrations.md` finding #1): the "parse a scalar type out of an R string vector" `TryFromSexp` set was copy-pasted ~20× across uuid/url/regex/num_bigint — with url and uuid hand-rolling the STRSXP validation `String`'s own impl already performs — while the mirror-image `IntoR` direction has had `into_r_infallible!` all along. This adds the missing counterpart macro and migrates all four modules onto it.

## What changed

- **New `try_from_sexp_via_str_parse!(Type, "label", |s| parse(s))`** in `miniextendr-api/src/from_r.rs` (helper-macros region), expanding to the four delegating impls (`T` / `Option<T>` / `Vec<T>` / `Vec<Option<T>>`) over `Option<String>` / `Vec<Option<String>>` — type, length, and NA checks live in exactly one place.
- **Migrated:** uuid (4 impls → 1 invocation), url (4 → 1, deleting ~140 lines of hand-rolled `unsafe` STRSXP loops), regex (4 → 1), num_bigint (8 → 2, deleting the now-unused `parse_bigint`/`parse_biguint` helpers). Net **−314 LoC** in the four modules; `IntoR` sides untouched.

## NA-policy standardization (deliberate behavior changes)

Previously each module had its own NA story; now uniform:

| Shape | Policy |
|---|---|
| `T` | NA / NULL → `SexpError::Na` |
| `Option<T>` | NA / NULL → `None` (unchanged everywhere) |
| `Vec<T>` | NA → `InvalidValue("NA at index {i} not allowed for Vec<T>")` |
| `Vec<Option<T>>` | NA → `None`; parse failures error with index (unchanged) |

Concretely:
- **regex NA bug fixed:** `NA_character_` used to flow through `String`'s NA→`""` mapping and *silently compile as an empty pattern that matches everything* (scalar and `Vec<Regex>`). Now rejected as a missing value. Regression test included.
- uuid scalar NA: was a confusing `"invalid UUID: invalid length…"` parse error → now `SexpError::Na`.
- url scalar NULL: was `SexpTypeError` → now `SexpError::Na` (consistent "missing value" story).
- num_bigint `Vec<T>` NA: was un-indexed `SexpNaError` → now the indexed `InvalidValue` message (matches what uuid/url already said, preserved verbatim via `stringify!`).
- BigInt/BigUint parse errors gain the type label ("invalid BigInt: …") and, in vectors, the index. The only testthat message-grep on these paths (`test-feature-adapters.R:61`, `"invalid UUID"`) is unaffected; `test-conditions.R`'s `"invalid digit"` grep targets a different fixture and remains a substring regardless.

## Scope notes

- ordered_float / rust_decimal / num_complex / jiff / bytes are **not** migrated — their conversions are numeric/other shapes, not string-parse (per the audit finding).
- Vector conversions keep the pre-existing bail-on-first-failure shape; batched diagnostics (CLAUDE.md "collect all errors in vectorized ops") tracked in #1143 — now a one-site change instead of eight.

## Test plan

- [x] New integration test `miniextendr-api/tests/str_parse.rs` (7 tests): valid/invalid parse with label+index assertions, scalar NA → `Na` error, `Option` NA → `None`, `Vec<T>` NA rejection with index, `Vec<Option<T>>` NA → `None`, and the regex empty-pattern NA regression. Integration test (separate process) rather than `--lib` unit test to avoid the `R_MAIN_THREAD_ID` test-isolation hazard (#1139).
- [x] Existing `tests/{uuid,regex,num_bigint}.rs` — 29 tests pass unmodified.
- [x] `just check`, `just clippy` clean; `just test` passes everywhere except the known pre-existing `derive_dataframe_*` trybuild drift (local rustc 1.95 vs CI stable; untouched by this diff).
- [x] `cargo fmt --all --check` clean; all three CI clippy variants (`default` / `full-codegen`+integrations / `full-codegen-s7`+integrations) clean with `-D warnings`.
- [ ] rpkg testthat suite: cargo-only session (shared install slot held elsewhere) — coordinator runs the R-side suite at review checkpoint; uuid/url/regex/num-bigint are all in the auto-detected default feature set.

### DIFF
--- a/miniextendr-api/src/from_r.rs
+++ b/miniextendr-api/src/from_r.rs
@@ -1784,4 +1784,127 @@ macro_rules! impl_vec_option_try_from_sexp_list {
         }
     };
 }
+
+/// Implement the four string-parse `TryFromSexp` impls (`T`, `Option<T>`,
+/// `Vec<T>`, `Vec<Option<T>>`) for a type parsed from an R character vector.
+///
+/// Sibling of [`into_r_infallible!`](crate::into_r) for the reverse direction:
+/// every "parse a scalar type out of an R string" integration (uuid, url,
+/// regex, num-bigint) used to hand-write these four impls — some reinventing
+/// the STRSXP validation `String`'s own `TryFromSexp` already performs.
+/// This macro delegates to `Option<String>` / `Vec<Option<String>>`, so type,
+/// length, and NA checks live in exactly one place.
+///
+/// Semantics:
+/// - `T`: `NA_character_` / `NULL` → `SexpError::Na`; parse failure →
+///   `InvalidValue("invalid <label>: <err>")`.
+/// - `Option<T>`: `NA_character_` / `NULL` → `None`.
+/// - `Vec<T>`: any NA element →
+///   `InvalidValue("NA at index <i> not allowed for Vec<T>")`; parse failure →
+///   `InvalidValue("invalid <label> at index <i>: <err>")`.
+/// - `Vec<Option<T>>`: NA elements → `None`; parse failures error with index.
+///
+/// The parse body is a closure-style `|s| expr` where `s: &str`, returning
+/// `Result<T, E>` with `E: Display`.
+///
+/// ```ignore
+/// try_from_sexp_via_str_parse!(Uuid, "UUID", |s| Uuid::parse_str(s));
+/// ```
+#[macro_export]
+macro_rules! try_from_sexp_via_str_parse {
+    ($ty:ty, $label:literal, |$s:ident| $parse:expr) => {
+        impl $crate::from_r::TryFromSexp for $ty {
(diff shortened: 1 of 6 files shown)


CGMossa:
Test-plan follow-up: rpkg testthat suite run against this branch after `just configure && just rcmdinstall` — `[ FAIL 0 | WARN 0 | SKIP 20 | PASS 6805 ]`. R-side gate green (uuid/url/regex/num-bigint are all in the auto-detected default feature set).

PR #1145, opened 2026-07-05
fix(gc): root the synthetic data.frame container in gc_stress_typed_dataframe

CGMossa:

<details>

## The bug

`gc_stress_typed_dataframe` failed on the CI reruns of two unrelated PRs (#1043 at iteration 3, #1044 at iteration 5) with `DataFrame always carries a names attribute`, while passing on main's own CI.

`List::from_raw_pairs` protects its list and names vectors only *during* construction; the `OwnedProtect` guards drop on return. The fixture then called `as_data_frame()` on the unprotected container, which allocates the `data.frame` class STRSXP and `row.names` INTSXP, and then drove `TheophDf::try_from_sexp` over it. Under `gctorture(TRUE)` the container node is reaped at the first of those allocations. The read-after-free is silent until R reuses the node, at which point the `names` attribute reads back nil and `DataFrame::named_list()`'s expect fires.

The node-reuse dependence is why it was invisible on main and stochastic on branches: extra fixtures/tests shift the allocation schedule. Both halves of the fixture (with and without the optional `flag` column) had the pattern; both are now rooted into the existing `ProtectScope` before `as_data_frame()` runs.

Production callers are unaffected: real `#[miniextendr]` entry points receive their SEXP from R, already rooted in the call frame. Only the fixture synthesises the container itself. The neighbouring fixtures were audited for the same shape: `gc_stress_as_named_list_deferred` does an allocation-free readback (safe by design), and `gc_stress_factor_labels` consumes its container immediately by intent.

## Verification

- 100/100 iterations of `gc_stress_typed_dataframe()` under `gctorture(TRUE)` post-fix (the pre-fix failures hit by CI at iterations 3 and 5).
- `cargo fmt --check` clean.

## Impact

This unblocks the CI reruns on #1043 and #1044 (their only R-side failure is this fixture) and removes a stochastic red-CI source for every future PR.

</details>

### DIFF
--- a/reviews/2026-07-05-gc-typed-dataframe-unrooted-container.md
+++ b/reviews/2026-07-05-gc-typed-dataframe-unrooted-container.md
@@ -1,0 +1,64 @@
+# 2026-07-05 — gc_stress_typed_dataframe: unrooted container UAF flaking unrelated PRs
+
+## What was attempted
+
+Post-sweep CI verification of the rebased open PRs (all had been force-pushed
+onto current main during the 2026-07-04/05 review sweep).
+
+## What went wrong
+
+Three unrelated branches went red on the same rerun wave:
+
+- #1043 and #1044: `gc_stress_typed_dataframe` failed the gctorture sweep with
+  `DataFrame always carries a names attribute` (iterations 3 and 5
+  respectively). Neither PR touches typed dataframes.
+- #1134 (minirextendr-only diff): rpkg testthat segfaulted on the R-devel
+  check job with `*** recursive gc invocation` spam — plausibly the same bug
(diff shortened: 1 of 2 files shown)


CGMossa:
Additional evidence this flake class is live on main: #1134 (a minirextendr-only diff) just hit a segfault with `*** recursive gc invocation` spam in rpkg's testthat on the R-devel check job. A use-after-free of the fixture's unprotected container can surface as a segfault rather than the clean names-attribute panic, depending on node reuse — so that crash is plausibly this same bug. If #1134's devel job still segfaults after this merges and its branch picks the fix up, that points to a second, distinct devel-only issue.

Issue #1151, opened 2026-07-05
minirextendr: unify .Rbuildignore/.gitignore scaffolding between monorepo and standalone paths


CGMossa:
Follow-up to the D8 scaffold-enumeration dedup (see audit
`audit/2026-07-03-dogfooding-minirextendr-r.md` finding #4 and
`plans/2026-07-05-audit-d8-scaffold-enumeration-dedup.md`).

That PR unified the standalone (`use_miniextendr_*()`) and monorepo
(`create_rpkg_subdirectory()`) scaffold paths for every file whose content
only depends on the package name: configure.ac, bootstrap.R, cleanup*,
configure.win/ucrt, config.guess/config.sub + lock-shape-check.R, Makevars.in,
win.def.in, stub.c/r_shim.h, mx_abi.h, package.R, build.rs — plus lifted the
LICENSE body, minimal NAMESPACE content, and `Config/build/*` field set into
shared constants (`mx_license_content()`, `mx_minimal_namespace()`,
`MX_CONFIG_BUILD_FIELDS` in `minirextendr/R/utils.R`).

**Left un-unified, deliberately, because the two paths do genuinely different
things today:**

- `use_miniextendr_rbuildignore()` / `use_miniextendr_gitignore()` (standalone)
  read the template, filter blank/comment lines, and call
  `usethis::use_build_ignore()` / `usethis::use_git_ignore()`, which
  append-and-dedupe into any existing `.Rbuildignore`/`.gitignore`.
- `create_rpkg_subdirectory()` (monorepo) instead does a raw
  `use_template("Rbuildignore", ...)` / `use_template("gitignore", ...)` —  a
  full overwrite, no append/dedupe semantics.

Unifying these safely requires confirming the `usethis::use_build_ignore()` /
`use_git_ignore()` output is byte-identical to a straight template copy for a
brand-new file (it may reorder/dedupe/reformat lines), which the D8 PR did not
attempt in order to keep its "byte-identical before/after" acceptance bar
tight. Worth revisiting: either prove the usethis path produces identical
output for the fresh-scaffold case and route the monorepo path through
`use_miniextendr_rbuildignore()`/`use_miniextendr_gitignore()` too, or extract
just the "read template, filter comments" prep step into a shared helper.

PR #1152, opened 2026-07-05
refactor(minirextendr): single scaffold file-set enumeration across standalone/monorepo/inline paths

CGMossa:
Addresses audit `2026-07-03-dogfooding-minirextendr-r.md` finding #4 (HIGH):
the miniextendr scaffold file set was enumerated independently in three
places — standalone `use_miniextendr_*()` (`create.R:418-446`, reused by
`upgrade.R`), monorepo `create_rpkg_subdirectory()` (`create.R:191-292`), and
inline `scaffold_inline_package()` (`rust-source.R:257-391`) — so a dropped
file or changed field in one could silently diverge the others. Plan:
`plans/2026-07-05-audit-d8-scaffold-enumeration-dedup.md`.

## Unification boundary

- **Standalone ↔ monorepo — fully unified for anything that only depends on
  the package name.** `use_miniextendr_configure()`, `_bootstrap()`,
  `_cleanup()`, `_configure_win()`, `_config_scripts()`, `_makevars()`,
  `_stub()`, `_mx_abi()`, `_package_doc()`, and `_build_rs()` each gained a
  `subdir` argument (threaded through to `use_template()`/`template_path()`).
  `create_rpkg_subdirectory()` now calls these with `path = <rpkg subdir>` and
  `subdir = "rpkg"` instead of re-implementing the sequence by hand — a
  dropped file in one path is now structurally impossible in the other,
  because there's only one implementation.
- **LICENSE / NAMESPACE / `Config/build/*` — lifted into shared constants**
  in `utils.R`: `mx_license_content()`, `mx_minimal_namespace()`,
  `MX_CONFIG_BUILD_FIELDS`. Both `use_miniextendr_description()` (standalone)
  and `create_rpkg_subdirectory()`'s hand-written DESCRIPTION literal
  (monorepo) now build from the same source.
- **config.guess/config.sub copy loop — one helper, three callers.**
  `copy_config_scripts()` in `utils.R` is now used by
  `use_miniextendr_config_scripts()` (standalone), `create_rpkg_subdirectory()`
  (via the same function, monorepo), and `scaffold_inline_package()` (inline,
  quiet mode).
- **DESCRIPTION construction stays separate, deliberately.**
  `use_miniextendr_description()` updates-or-delegates-to
  `usethis::use_description()`; `create_rpkg_subdirectory()` still writes a
  hand-rolled literal (placeholder Title/Authors@R shape, never an existing
  file to update). Only the `Config/build/*` fields are shared between them.
- **`scaffold_inline_package()` stays independent** (staticlib crate-type, no
  `Config/build/*` fields, git-URL deps, dynamic NAMESPACE exports built from
  the parsed code) — it only adopts `copy_config_scripts()`, per the plan's
  explicit scope (do not force it through `use_miniextendr_*()`).

## Verification

Wrote a temp-dir hash-manifest harness (not committed — ad hoc verification
script) that exercises all three public scaffold entry points
(`create_miniextendr_package()`, `create_miniextendr_monorepo()`,
`scaffold_inline_package()`) via `devtools::load_all()`, hashes every produced
file, and diffs the manifest before/after the refactor. Result: **byte-identical
across all 90 produced files** (standalone: 37, monorepo: 40, inline: 13),
confirmed on two separate pre-refactor runs (to rule out nondeterminism in
`autoconf`/`autom4te.cache`) and again post-refactor.

`just minirextendr-test`: **`FAIL 0 | WARN 0 | SKIP 3 | PASS 608`** (the 3
skips are pre-existing, unrelated `test-use-native.R` cases gated on tooling
not present in this sandbox).

`devtools::document("minirextendr")` regenerated 9 man pages with the new
`@param subdir` entries (roxygen2 clean — no unresolved-link warnings; fixed
two I introduced myself by using `[fn()]` autolink syntax against `@noRd`
functions, now plain code font). `NAMESPACE` is unchanged (no export changes).
`inst/templates/` untouched (`git status` confirms — this is an R-code-only
change).

## Every test/doc change

- No `tests/testthat/*.R` files needed changes — the existing
  `test-monorepo-gaps.R` B1 test (`create_rpkg_subdirectory()` called
  directly with a hand-built `data` list, asserting `rpkg/src/r_shim.h`
  placement) and `test-exports-exist.R` (existence-only check, tolerant of
  new optional params) both still pass unmodified against the new call-through
  implementation.
- 10 man pages regenerated (`use_miniextendr_{bootstrap,build_rs,cleanup,
  config_scripts,configure,configure_win,makevars,mx_abi,package_doc,stub}.Rd`)
  — additive `@param subdir` documentation only, no signature-breaking changes
  (all new params default `NULL`, all existing call sites in `upgrade.R` /
  `create.R` call with zero args and are unaffected).

## Known cosmetic-only behavior changes (file content unaffected)

- The monorepo path's "Copied" CLI bullets for `tools/config.guess` /
  `tools/config.sub` / `tools/lock-shape-check.R` now read `tools/<file>`
  instead of `<rpkg_name>/tools/<file>` (message is relative to the
  now-active rpkg sub-project instead of the workspace root).
- `create_rpkg_subdirectory()` now also prints the standalone
  `use_miniextendr_package_doc()`'s "Run `devtools::document()`..." info
  bullet, previously silent in the monorepo path.

## Deferred scope

`.Rbuildignore`/`.gitignore` still diverge between the standalone path
(`usethis::use_build_ignore()`/`use_git_ignore()`, which append+dedupe into
an existing file) and the monorepo path (raw template overwrite via
`create_rpkg_subdirectory()`) — unifying them requires confirming the usethis
functions produce byte-identical output for a fresh file, which this PR did
not attempt in order to keep the "byte-identical before/after" bar
unambiguous. Filed as A2-ai/miniextendr#1151.

---
<details>
<summary>AI-assisted PR</summary>

This PR description and the underlying refactor were drafted by an AI
coding agent (Claude) under human supervision, following a pre-written plan
(`plans/2026-07-05-audit-d8-scaffold-enumeration-dedup.md`) derived from a
prior audit pass.

</details>


CGMossa:
Coordinator follow-up (install-based gate): `just minirextendr-check` fails on this branch — but identically on #1153 and at a site this diff never touches (`test-templates.R:769`: pak dependency resolution inside R CMD check's sandbox, plus two standing WARNINGs). Pre-existing, tracked in #1154. The branch-level gate that is attributable to this diff is green: `just minirextendr-test` → `[ FAIL 0 | WARN 0 | SKIP 3 | PASS 608 ]` + the byte-identical 90-file scaffold diff.

Issue #1161, opened 2026-07-05
Stray tracked, dead file: rpkg/src/rust/miniextendr_wrappers.R


CGMossa:
While working audit finding #5 (PR renaming rpkg_has_feature/rpkg_enabled_features
to miniextendr_*, see plans/2026-07-06-audit-a3-rpkg-prefix-exports.md), found that
`rpkg/src/rust/miniextendr_wrappers.R` is tracked by git (last touched in commit
902c3b83, "Add minirextendr scaffolding..." from Jan 2026) but is NOT the real
generated wrapper file — that's the gitignored `rpkg/R/miniextendr-wrappers.R`
(hyphen, different directory) documented in CLAUDE.md.

This `src/rust/` copy (underscore, ~3.4K lines) appears to be a stale one-time
snapshot from before the wrapper-generation pipeline moved to `R/`. It still
contains the old `rpkg_*`-prefixed function names (`rpkg_altrep_unsafe`,
`unsafe_rpkg_altrep_from_doubles`, etc.) and is not referenced by any Makevars,
build.rs, or Rust source — `grep -rn "miniextendr_wrappers.R" rpkg` outside the
file itself returns nothing.

Recommend: verify it is truly unreferenced/dead, then delete it (untracked artifact
policy in CLAUDE.md — generated files should be gitignored, not committed once).
Not fixed in the A3 PR since it's out of scope for the rpkg_/miniextendr_ prefix
rename and deserves its own verification pass.

CGMossa:
Fixed in PR #1162 (commit 529f4505) rather than deferred — the file is dead and carried the old `rpkg_*` stubs that PR renames. Will close on merge.

PR #1162, opened 2026-07-05
fix(rpkg): rename rpkg_-prefixed feature helpers to miniextendr_* (audit A3)

CGMossa:
The exemplar package loads as `library(miniextendr)`, but its exported feature-detection helpers were named `rpkg_has_feature()` / `rpkg_enabled_features()` — the source-directory name (`rpkg/`) leaking into the public surface. Since `rpkg/` is the template end-user packages get cloned from, that awkward prefix would have propagated into every scaffolded package. This renames them to match the installed package name.

<details>

**Source:** `audit/2026-07-03-api-sense-externalptr-gc-misc.md` finding #5, via `plans/2026-07-06-audit-a3-rpkg-prefix-exports.md`.

**Changes:**
- `rpkg_enabled_features()` (Rust `#[miniextendr]` fn, `rpkg/src/rust/lib.rs`) → `miniextendr_enabled_features()`, including its roxygen `@name`/`@examples`.
- `rpkg_has_feature()` (hand-written R, `rpkg/R/feature_helpers.R`) → `miniextendr_has_feature()`, including its roxygen examples and the internal call inside `skip_if_missing_feature()`.
- Updated all ~16 callers across `rpkg/tests/testthat/*.R` (used exact-identifier `\b`-anchored substitution, not a blanket `s/rpkg_//`, per the known sed-substring-rename gotcha).
- Regenerated `rpkg/NAMESPACE` and `rpkg/man/*.Rd` via `just rcmdinstall && just force-document`; the old `man/rpkg_has_feature.Rd` / `man/rpkg_enabled_features.Rd` were removed (git recorded these as renames to the new filenames).
- No back-compat aliases — unreleased project, per project convention.

**Scaffolder check:** `minirextendr/R/use-feature.R`'s `use_feature_detection()` already derives the prefix from the target package's actual `DESCRIPTION` `Package:` field (`package_rs <- gsub("[.-]", "_", package_name)`), not a hard-coded `rpkg_`/`miniextendr_` string. `minirextendr`'s own tests (`test-use-feature.R`) already exercise this with `testpkg_*` / `my_pkg_*` derived names. So `rpkg/`'s helpers were hand-written directly with the `rpkg_` prefix rather than generated by this scaffolder — no scaffolder changes were needed. No copy of `feature_helpers.R` exists under `minirextendr/inst/templates/`, and `patches/templates.patch` has no reference to the old names, so nothing to re-approve there.

**Deferred (filed as issue):** while investigating, found `rpkg/src/rust/miniextendr_wrappers.R` (underscore path, distinct from the real gitignored `rpkg/R/miniextendr-wrappers.R`) is a stale tracked file from Jan 2026 containing old `rpkg_*`-prefixed function stubs, unreferenced by any build path. Out of scope for this rename — filed as A2-ai/miniextendr#1161.

**Test plan:**
- `just check` — clean, no warnings.
- `just clippy` — clean.
- `cargo fmt --all -- --check` — clean.
- `just configure && just rcmdinstall && just force-document && just rcmdinstall` (second install so renamed exports are runtime-callable).
- `just devtools-test` → `[ FAIL 0 | WARN 0 | SKIP 20 | PASS 6805 ]`.
- `minirextendr` package itself was not touched (scaffolder already correct), so its test suite was not re-run.

</details>


CGMossa:
Follow-up commit 529f4505: deleted the stray tracked `rpkg/src/rust/miniextendr_wrappers.R` (the #1161 finding) in this PR instead of deferring — it's confirmed dead (the only repo references are a scaffolder test asserting its *absence* and two doc comments about the real `R/`-level generated file), and it still carried the old `rpkg_*` stubs this PR renames. #1161 closes on merge.

PR #1200, opened 2026-07-07
fix(api): ExternalPtr<T> arguments unwrap class-wrapped handles (audit A9)

CGMossa:
Passing a class object where a standalone fn expects `ExternalPtr<T>` used to fail with a bare type-mismatch error — the test suite itself had to document the workaround ("use the low-level constructor, the R6 handle would fail conversion"). That asymmetry is a trap: the ergonomic constructor and the bare-pointer constructor both hand you "the same object" from R's point of view, but only one of them was accepted at the conversion boundary. This PR makes `TryFromSexp for ExternalPtr<T>` unwrap class-wrapped handles transparently (R6 `private$.ptr`, S4 `ptr` slot, S7 `.ptr` attribute, plus any env binding `.ptr` directly), fixing audit finding #5 (A9, `audit/2026-07-03-api-sense-conversions-dataframe-errors.md`). `Any::downcast` stays the type-safety authority — a handle wrapping the wrong `T` still fails with the existing mismatch error naming both types, so this loosens accepted shape, not safety.

## Summary

- `miniextendr-api/src/externalptr.rs`: new main-thread helper `unwrap_class_handle(sexp) -> Option<SEXP>`. For ENVSXP: a direct `.ptr` binding, then the R6 chain `.__enclos_env__` -> `private` -> `.ptr` (each step guarded via `R_UnboundValue`, newly declared in `sys.rs`). For S4 (`isS4()`-gated, which excludes S7 despite the shared OBJSXP type): the `ptr` slot via `s4_helpers::s4_get_slot` (`methods::slot()`). Otherwise: the `.ptr` attribute (covers S7 property storage). Returns `Some` only if the inner value is an EXTPTRSXP; one unwrap level, no recursion. The returned SEXP is reachable from the argument (binding/slot/attribute), which the `.Call` frame keeps protected — **no new SEXP storage across allocations, so no gc-stress fixture is needed** (unwrap only reads existing bindings).
- `miniextendr-api/src/from_r.rs`: both `try_from_sexp` and `try_from_sexp_unchecked` for `ExternalPtr<T>` try `unwrap_class_handle` on the non-EXTPTRSXP branch. The container impls (`Option<ExternalPtr<T>>`, `Vec<ExternalPtr<T>>`, `Vec<Option<ExternalPtr<T>>>`) delegate element-wise, so they gain the behavior uniformly — no macro changes. Remaining failures self-explain: `expected an external pointer or a miniextendr class object wrapping one, got <SEXPTYPE>`. The old duplicated error-mapping closure is factored into `type_mismatch_to_sexp_error`.
- Per class system, what actually happens: **Env / S3 / Vctrs-style S3 objects** are bare classed EXTPTRSXPs — they always passed and never reach the unwrap. **R6** unwraps through the enclos-env chain. **S4** unwraps the `ptr` slot. **S7** unwraps the `.ptr` attribute. **Vctrs** has nothing to unwrap by design — vctrs constructors return a payload vector, not `Self`; there is no Rust struct behind the object.
- `typed_list!` alignment (flagged in the plan as a risk): verified no misalignment exists. `TypeSpec` has no ExternalPtr variant (only `Class(&str)` via `Rf_inherits`, which already accepts class handles), and `TypedList::get::<ExternalPtr<T>>()` delegates to `TryFromSexp`, so dots/typed_list access gains the unwrap automatically. Nothing to file.
- rpkg: three `noexport` fixtures (`s3/s4/s7_raiser_id_via_externalptr` in `condition_class_system_tests.rs`, reusing existing Raiser types); Env and R6 reuse existing fixtures (`ptr_identity`, `sidecar_consumer_panic`). The `test-conditions-comprehensive.R:457` workaround comment is gone — both the bare-pointer and R6-handle paths are now tested.
- `docs/EXTERNALPTR.md`: new "Passing Class Objects to Standalone Functions" section with the per-class-system table. Rustdoc on the helper and impl is standalone (no docs/*.md links).

## Test plan

- [x] `miniextendr-api/tests/externalptr_class_handle.rs` (with_r_thread, per #1139): env `.ptr` binding unwraps; `.ptr` attribute unwraps; plain env errors with the new hint; wrong-`T` handle still reports the downcast mismatch naming both types.
- [x] testthat (`test-conditions-comprehensive.R`): R6 handle accepted by `sidecar_consumer_panic()`; S4 slot / S7 attribute unwrap assertions; Env/S3 bare-handle regression checks; wrong-`T` mismatch names both types; non-handle env gets the hint message.
- [x] `just check` / `just test` / `just clippy` / `just lint` — clean (only pre-existing `derive_dataframe_*` trybuild drift, untouched per convention).
- [x] All three CI clippy variants (`clippy_default`, `clippy_all`, `clippy_all_s7`) locally with `-D warnings` — clean. `cargo fmt --all -- --check` clean.
- [x] `just configure && just rcmdinstall && just force-document && just rcmdinstall`, then full `just devtools-test` — `[ FAIL 0 | WARN 0 | SKIP 23 | PASS 6977 ]` (all skips pre-existing feature/Rd_db gates; no transient reruns needed). Targeted `devtools::test(filter = "conditions-comprehensive")` after the final doc pass: `[ FAIL 0 | PASS 97 ]`.
- [x] Smoke-tested every unwrap path against the *installed* package from a plain `library(miniextendr)` session (R6 handle -> `sidecar_consumer_panic`, S4 slot, S7 attribute, S3 bare handle, wrong-`T` mismatch, non-handle hint).

</details>


CGMossa:
Reviewed — the unwrap helper's shape table matches the generators (Env/S3 bare classed EXTPTRSXP, R6 `.__enclos_env__` -> `private` -> `.ptr`, S4 `ptr` slot behind `isS4()`, S7 `.ptr` attribute), protection reasoning is sound (inner SEXP stays reachable from the protected argument), and the negative tests pin both the downcast-mismatch and the new not-a-handle hint. No changes requested.

One sequencing note for the merger: the regenerated `condition_class_system_tests.Rd` picks up `\alias`/`\usage` entries for the three new `noexport` fixtures. That's the known A10 alias-leak bug (noexport currently doesn't suppress Rd aliases), not a defect in this PR — the in-flight A10 branch (`fix/macros-internal-vs-noexport`) gates `ClassDocBuilder` so these lines disappear on its regen. Merge order A9 → #1186 → A10 keeps everything consistent.

Issue #1180, opened 2026-07-06
vctrs class generator never gates docs on @noRd/internal/noexport (self-coercion + instance/static methods)


CGMossa:
While fixing audit finding #6 (`internal` vs `noexport` export-control modes were observationally identical — see `audit/2026-07-03-api-sense-conversions-dataframe-errors.md`), I found that `miniextendr-macros/src/miniextendr_impl/vctrs_class.rs` has a broader, pre-existing gap: **unlike the other 5 class-system generators (R6/Env/S3/S4/S7), it has no `class_has_no_rd` variable at all**, and never checks `@noRd`/`internal`/`noexport` before emitting roxygen for anything except the constructor (which routes through the shared `ClassDocBuilder` and is fixed by the A10 PR).

Concretely, in `generate_vctrs_r_wrapper`:
- The `vec_ptype_abbr.<class>` doc block (`~L126`) always pushes `#' @rdname {class_name}`.
- The `vec_ptype2.<class>.<class>` self-coercion doc block (`~L140`) always pushes `#' @rdname`, `#' @param x`, `#' @param y`, `#' @param ...`.
- The `vec_cast.<class>.<class>` doc block (`~L193`) always pushes `#' @rdname` + `#' @param` tags.
- Instance-method S3 generic docs (`~L228`) always push `#' @title`/`#' @description`/`#' @rdname`/`#' @name`.
- Instance/static method docs go through `MethodDocBuilder` (`~L250`, `~L303`, `~L313`) but are never called with `.with_class_no_rd(...)`, so `MethodDocBuilder`'s built-in `@noRd` gate (used correctly by R6/S3/S4) is never engaged for vctrs.

Only `should_export` (which does correctly fold `noexport || internal`) gates the `#' @export` line in each of these blocks — so a vctrs class marked `#[miniextendr(vctrs, noexport)]` (or even a plain user-written `@noRd` on the impl block) still gets a fully-aliased, fully-titled Rd page for every self-coercion and instance/static method, just without an `@export` line. This is the same class of bug as A10, but wasn't fixed there because it requires restructuring ~5 separate emission sites in `vctrs_class.rs` (compute a `class_has_no_rd` fold like the other 5 generators now have — `class_has_no_rd || (noexport && !internal)` — and thread it through every doc-emitting block, including `MethodDocBuilder::with_class_no_rd`), which is a larger, more invasive change than the standalone-fn + `ClassDocBuilder`/`with_export_control` fix that closed A10.

[...]

Issue #431, opened 2026-05-08
doc: undocumented s3_* condition helpers (raise_condition, raise_error, etc.)


CGMossa:
## Background

`R CMD check --as-cran` reports 8 undocumented user-level objects:

```
* checking for missing documentation entries ... WARNING
Undocumented code objects:
  's3_id' 's3_raise_condition' 's3_raise_condition_classed'
  's3_raise_error' 's3_raise_error_classed' 's3_raise_message'
  's3_raise_warning' 's3_raise_warning_classed'
```

These were added by PR #344 (condition system unification). They're R-side helpers that ride alongside the `error!` / `warning!` / `message!` / `condition!` macros and need either roxygen docs or `@keywords internal` + `@noRd` to suppress NAMESPACE export.

## Acceptance

- Add `@keywords internal` + `@noRd` if not user-facing, OR a roxygen block with `@param` + `@return` if they are.
- `R CMD check --as-cran` no longer flags these under "missing documentation entries".
- Pre-commit hook still passes (NAMESPACE in sync with wrappers).

## Related

- #344 — condition system PR that introduced these
- Discovered while running R CMD check for #427

## Why now

Surfaced cleanly during the #427 audit's empirical verification. Unrelated to that PR's change but blocks any future CRAN submission. Cheap fix.

PR #1203, opened 2026-07-07
fix(macros): make noexport suppress Rd aliases so internal/noexport actually differ (audit A10)

CGMossa:
`internal` and `noexport` were two names for the same observable behavior: both kept the function callable via `:::`, both dropped it from NAMESPACE `export()`, and both still landed an `\alias{}` + usage entry in the shared Rd page under one `\keyword{internal}` (audit finding #6, `audit/2026-07-03-api-sense-conversions-dataframe-errors.md`). The docs always promised more — `internal` = documented-but-hidden, `noexport` = no man entry at all — so this makes the codegen deliver that: `noexport` now emits `@noRd` and drops out of every grouped/shared Rd page (standalone fns, the 5 class generators, and trait impls), while `internal` keeps its alias under `\keyword{internal}`. For S3 impls, `noexport` now also drops the `S3method()` dispatch registration — zero observable trace — while a hand-written `@noRd` keeps dispatch as before.

## Summary

Fixes audit A10 (finding #6): `#[miniextendr(internal)]` and `#[miniextendr(noexport)]` are now observably different at the documentation level.

**Before:** both modes only suppressed the `@export` line. A `noexport` fn still contributed `\alias{export_control_noexport}` + usage + description to `man/export_control_tests.Rd`.

**After:**
- `internal` → unexported, documented: alias + usage in the (shared) Rd under `\keyword{internal}`.
- `noexport` → unexported, `@noRd`: no alias, no usage, no Rd contribution anywhere.

### Paths fixed

| Path | File | Fix |
|---|---|---|
| Standalone fns | `miniextendr-macros/src/lib.rs` | emit `#' @noRd` for `noexport`; `collect_r_wrappers()`'s per-source-file `@rdname` grouping already skips `@noRd` blocks, so the fn drops off the grouped page |
| Class-doc header (all 6 systems) | `r_class_formatter.rs` (`ClassDocBuilder::build`) | fold plain `noexport` into the same suppression gate as user `@noRd` (emits `@noRd`, skips `@title`/`@name`/`@rdname`/`@source`/imports); `internal` wins when both flags meet |
| env / r6 / s3 / s4 / s7 generators | `miniextendr_impl/*.rs` | fold `noexport && !internal` into `class_has_no_rd` so method/static/generic doc blocks are suppressed consistently; S3 additionally drops `S3method()` registration for `noexport` (mirrors `should_register_s3method`, #431) |
| Trait impls | `miniextendr_impl_trait/r_wrappers.rs` | `noexport` routes through the `@noRd` strip path (previously it only deleted `@export` lines, leaving alias contributions) |
| S4 blind pop | `miniextendr_impl/s4_class.rs` | pre-existing bug: the generator popped the last class-doc line assuming it's `@export`; with `internal`/`noexport` that deleted `@keywords internal` / `@noRd`. Pop is now conditional — `#[miniextendr(s4, internal)]` class pages now correctly gain `\keyword{internal}` |

### Observable diff in this PR

- `rpkg/man/export_control_tests.Rd`: the `export_control_noexport` alias, usage entry, and description are gone.
- `rpkg/man/ExportControlTraitPoint.Rd`: the `RDisplay` (`noexport`) trait-method alias is gone; `RDebug` (`internal`) stays.
- `man/*.Rd` for `#[miniextendr(s4, internal)]` fixtures gain `\keyword{internal}` (blind-pop fix).
- `docs/CLASS_SYSTEMS.md`: states the alias-level distinction + the S3-dispatch rule crisply.

### Sequencing with #1186

PR #1186 (vctrs doc gating, fixing #1180) landed on `main` while this branch was in flight and anticipates it: the vctrs generator previously had no `class_has_no_rd` gate at all (except on the constructor via the shared `ClassDocBuilder`), so #1186 closed that gap for self-coercion blocks, instance-method S3 generic docs, and MethodDocBuilder call sites — but its own test fixture (`VctrsNoexportGated`) still left behind a constructor-only Rd page emitted by the `ClassDocBuilder` path, with a comment noting this branch would remove it once it gates that block. Rebasing onto `main` and re-running `force-document` did exactly that: `rpkg/man/VctrsNoexportGated.Rd` is now deleted, confirmed by `test-vctrs-export-control.R`'s Rd_db assertions passing against the freshly regenerated docs. #1180 is closed; no further deferral needed there.

## Test plan

- [x] `cargo test -p miniextendr-macros` — 432 unit tests pass, including new coverage: `ClassDocBuilder` noexport/internal/default triples, trait-wrapper noexport-strips-all/internal-keeps-rdname (S3 + S4), the 5-generator sweep (`class_generators_noexport_suppresses_all_rd`, `class_generators_internal_stays_documented`), and the S4 blind-pop regression guard.
- [x] `just check` clean; `just test` — only the pre-existing 7-of-138 `derive_dataframe_*` trybuild mismatches (local rustc 1.95 vs CI's pinned stable toolchain, documented drift, unrelated to this change); `cargo fmt --all` clean, no diff.
- [x] rpkg loop (`just configure && just rcmdinstall && just force-document`): regenerated Rd shows the `noexport` aliases vanish (plus the `VctrsNoexportGated.Rd` deletion above); NAMESPACE unchanged (both modes were already unexported).
- [x] `rpkg/tests/testthat/test-export-control.R`: Rd_db-level assertions (skip-guarded per `test-r6-noexport-field.R`) — `export_control_internal` has an alias, `export_control_noexport` has no alias on any Rd page; same pair for the `ExportControlTraitPoint` trait impls. These skip under `devtools::test()`'s `load_all()` (no installed Rd database in that mode) — verified separately by running the test file directly against the freshly-installed package (`library(miniextendr); testthat::test_file(...)`): `test-export-control.R` (10/10) and `test-vctrs-export-control.R` (21/21) both pass in full, with zero skips.
- [x] Full `rpkg` testthat sweep against the fresh install, chunked alphabetically: a-h (`[ FAIL 0 | PASS 4951 ]`), i-l (`[ FAIL 0 | PASS 244 ]`), m-r (`[ FAIL 0 | PASS 713 ]`), s-z (`[ FAIL 0 | PASS 1058 ]`) — 6966 passes, zero failures. All skips are pre-existing and expected (feature-gated tests, `test-thread-broken.R`'s intentional skip, and the Rd_db tests noted above, which correctly skip under `load_all`).

</details>


CGMossa:
Reviewed the mechanism — approach is correct. The single `suppress_rd = has_no_rd || (attr_noexport && !effective_internal)` predicate gating every page-generating tag (`@title`/`@name`/`@rdname`/source/imports/`@export`) in `ClassDocBuilder`, plus the parallel `no_rd_comment` injection for standalone fns in `lib.rs`, is the right shape: `internal` stays documented under `\keyword{internal}`, `noexport` now contributes nothing to Rd, and `internal` wins when both are set (matches the standalone-fn precedence). Applying it to the trait-method path (`miniextendr_impl_trait/r_wrappers.rs`) too closes the gap for all generators. The direct-Rd_db verification (test-export-control 10/10, test-vctrs-export-control 21/21, Rd_db excludes `VctrsNoexportGated.Rd`) confirms the end-to-end behavior. No changes requested.

Sequencing/merge notes for whoever merges:
- **#1186 precondition is satisfied** — it's already merged, so the `VctrsNoexportGated.Rd` deletion here is the anticipated follow-through.
- **This branch predates #1189 (A7), which is now on main.** A7 marked ~276 `unsafe_C_`/`gc_stress_`/`test_` standalone fns `noexport`. Since this PR is what makes `noexport` inject `@noRd` on standalone fns, the rebase-immediately-before-merge pass must regenerate wrappers.R + man/ against post-#1189 main and re-run `just wrappers-sync-check` — the regenerated man/ diff may be **larger** than what's shown here (it should additionally suppress any A7 fixture Rd contributions that main still carries). That's expected and correct, not drift to fight.

Issue #1206, opened 2026-07-08
impl method-tag nudge is a silently-dead #[deprecated] const


CGMossa:
## What

`strip_method_tags` / `strip_method_tags_r6` (`miniextendr-macros/src/roxygen.rs`) emit, for each method-only roxygen tag placed on a `#[miniextendr]` impl block, a nudge const:

```rust
#[deprecated(note = "@param on impl block `Foo` has no effect — move it to the method")]
#[doc(hidden)]
#[allow(dead_code)]
const _MINIEXTENDR_IMPL_METHOD_TAG_WARN_Foo_..: () = ();
```

This **never surfaces a warning**. A `#[deprecated]` item only warns when it is *used*; declaring an unused deprecated const emits nothing (confirmed with a standalone `rustc` repro). So the intended nudge ("move this method tag off the impl block") is dead code today.

## Making it visible

Referencing the const in a sibling const initializer triggers the `deprecated` lint at the impl-block span, e.g.:

```rust
#[deprecated(note = "...")] #[doc(hidden)] #[allow(non_upper_case_globals)]
const _WARN_...: () = ();
#[doc(hidden)] #[allow(dead_code, non_upper_case_globals)]
const _USE_...: () = _WARN_...;   // <- use of deprecated const => warning fires here
```

## Why it wasn't activated in #1118

PR for #1118 only made the const *names* unique (they were colliding with E0428). Activating the nudge is a **separate, larger** change because CI's `cargo clippy --workspace --all-targets -- -D warnings` would then fail on the ~40 pre-existing impl-*level* method-only tags across `rpkg/src/rust/*` and `tests/cross-package/producer.pkg` — e.g. `class_system_matrix.rs` (`@param x` on an S4 trait impl), `producer.pkg/src/rust/lib.rs` (`@examples` on impl blocks), `s3_tests.rs`, `pipe_builder_tests.rs`. These are silently stripped today, so they compile.

Note: R6 keeps class-level `@param` intentionally (roxygen2 8.0.0 inherits them), so only the non-R6 `@param` + all-system `@return`/`@returns`/`@examples`/`@export` occurrences are affected.

## Plan when picked up

1. Activate the nudge (use-the-const trick above).
2. Sweep the workspace + rpkg + cross-package and delete/move every impl-level method-only tag. Deleting is safe and needs no doc regeneration — these tags are already stripped, so generated `wrappers.R`/`NAMESPACE`/`man` are unchanged.
[...]

PR #1207, opened 2026-07-08
fix(api,macros): rust_* macro aliases, RngGuard unsafe contract, impl-tag const collision (#1105 #1096 #1118)

### LINKED ISSUE #1096: Tighten RngGuard unsafe contract: drop Default, make new() unsafe
## Summary

`RngGuard` (`miniextendr-api/src/rng.rs`) should drop its `Default` impl and make `RngGuard::new()` an `unsafe fn`, to stop hiding R-main-thread-only unsafe behaviour behind a safe API. This is the deferred half of the "tighten unsafe contracts" work in #1092.

## Background

`RngGuard::new()` calls `unsafe { GetRNGstate() }` — an R C-API call that is only valid on the R main thread — but is exposed as a **safe** `fn`. It also has a `Default` impl whose `default()` silently invokes that same unsafe path (the foot-gun: `mem::take`, `Option::unwrap_or_default`, struct-update syntax, generic `T: Default` bounds can all call it without the caller realising they've entered unsafe territory).

PR #1092 already:
- removed `Default for ProtectScope` (its `new()` is `unsafe`, so removing `Default` is clippy-clean), and
- marked `RngGuard` `!Send + !Sync`.

It deliberately **left `RngGuard`'s `Default` + safe `new()` alone**, because they are interlocking:

- Removing `Default for RngGuard` **alone** trips `clippy::new_without_default` (CI runs `-D warnings`), since `RngGuard::new()` is currently a *safe* no-arg `new`.
- The clean resolution is to make `new()` `unsafe` (which both silences `new_without_default` *and* reflects the real main-thread contract) **and** drop `Default` together, as one contract change.

## Proposed change

1. `pub unsafe fn new() -> Self` (reflects: must be on the R main thread; `GetRNGstate` precondition).
2. Remove `impl Default for RngGuard`.
3. Update `with_rng` to call `unsafe { RngGuard::new() }` inside its existing safe wrapper (it already documents the main-thread requirement) — decide whether `with_rng` itself should become `unsafe` or assert the contract. **This is the maintainer judgement call**: `#[miniextendr(rng)]` codegen does *not* use `RngGuard` (it emits explicit `GetRNGstate`/`PutRNGstate`), so the blast radius is just `with_rng` + any manual callers.

## Source

Audit: `audit/rng.md` issues #1–#3. Verified against `miniextendr-api/src/rng.rs:141,154,160`.

<sub>Drafted by Claude (Opus 4.8) from the automated `audit/` review; filed at the maintainer's request.</sub>

### LINKED ISSUE #1105: Add collision-free macro aliases (rust_error!/rust_condition!) for the error!/condition! module clash
## Summary

The `error!` and `condition!` macros collide with the same-named crate-root modules (`pub mod error`, `pub mod condition`), so `use miniextendr_api::{error, condition};` then `error!(...)` fails with a confusing "cannot find macro" error. Track adding collision-free aliases.

## Background

This is a documented gotcha (root + `miniextendr-api/CLAUDE.md`): under `use miniextendr_api::error`, the name `error` resolves to the *module*, shadowing the `error!` macro. Today users must fully-qualify (`miniextendr_api::error!(...)`) or alias the crate. `warning!` / `message!` have no module conflict, so the asymmetry is itself a paper-cut for new users; rustc's diagnostic ("cannot find macro `error` in this scope") doesn't hint at the cause.

## Options (audit's recommendation)

1. Add collision-free macro aliases `rust_error!` / `rust_condition!` (mirroring the `rust_*` R class prefix already used in the condition transport) alongside the existing macros, and steer docs toward them.
2. Leave as-is (documented) — the fully-qualified path works.

Either way, keep `warning!`/`message!` as the precedent for naming.

## Source

Audit: `audit/condition.md` issue #2. The collision is documented in `miniextendr-api/CLAUDE.md` ("#[macro_export] collides with same-named modules at crate root").

<sub>Drafted by Claude (Opus 4.8) from the automated `audit/` review; filed at the maintainer's request.</sub>

### LINKED ISSUE #1118: macros: impl-block method-tag warning consts collide (E0428) across multiple #[miniextendr] impl blocks of one type

CGMossa:
Three small, independent fixes.

- **#1105 — collision-free macro aliases.** Added `rust_error!` / `rust_condition!` beside `error!` / `condition!` (identical expansions and grammar). The bare names are shadowed by the crate-root `pub mod error` / `pub mod condition`, so `use miniextendr_api::*;` resolves to the module, not the macro; the `rust_*` aliases have no such clash. Steered the name-collision doc notes toward the aliases, added a light note to `docs/CONDITIONS.md`, and added a compile-pass integration test (`miniextendr-api/tests/macro_aliases.rs`) proving the module imports and alias macros coexist. `warning!` / `message!` left as-is.

- **#1096 — RngGuard unsafe contract.** `RngGuard::new()` is now `unsafe fn` with a `# Safety` note (its `GetRNGstate()` call is only valid on R's main thread), and the `Default` impl is removed. `with_rng` stays a safe fn and upholds the invariant with a `debug_assert!(is_r_main_thread())` before the `unsafe { RngGuard::new() }` call. The `#[miniextendr(rng)]` codegen is unaffected (it calls GetRNGstate/PutRNGstate directly, never RngGuard). Fixed the sole other call site (`rpkg` `rng_guard_test`) and the doc examples.

- **#1118 — impl method-tag const collision.** The `#[deprecated]` nudge const emitted per stripped method-only impl-tag was named `_MINIEXTENDR_IMPL_METHOD_TAG_WARN_<Type>_<warning_id>`, restarting `warning_id` at 0 per block, so two `#[miniextendr] impl` blocks on one type (e.g. inherent + trait, both with an impl-level method tag) collided with E0428. Added a per-block disambiguator from a monotonic `next_impl_tag_block_id()` counter. Regression: a compile-pass trybuild test (inherent + trait impl on one type) plus a unit test asserting distinct block ids yield distinct const names.

**Related finding (not changed here):** the #1118 nudge const is a silently-dead `#[deprecated]` (unused deprecated items never warn); activating it would trip CI `-D warnings` on ~40 pre-existing impl-level method tags across rpkg/cross-package. Left as-is and tracked in #1206.

**Verification (pure cargo; no R needed):**
- `cargo test -p miniextendr-macros`: 438 unit tests pass (incl. the new disambiguation test); the new trybuild pass test passes. The 7 `derive_dataframe_*` trybuild mismatches are the known local-rustc-vs-CI-stable drift (MEMORY.md) — unrelated to this diff; CI is authoritative.
- `cargo test -p miniextendr-api --test macro_aliases`: passes.
- All three CI clippy legs clean (`-D warnings`): `clippy_default`, `clippy_all` (`full-codegen`), `clippy_all_s7` (`full-codegen-s7`).
- `cargo fmt --all` clean.
- The `rpkg/src/rust/rng_tests.rs` one-line unsafe-wrap is in rpkg's excluded workspace (not reached by `cargo --workspace`); it's a mechanical unsafe-block wrap, verified by inspection and exercised by the R-based rpkg CI jobs.

Closes #1105
Closes #1096
Closes #1118


CGMossa:

<details>

## Summary
- **#1105 (rust_* aliases): sound, simplified.** Solves the real, documented collision where crate-root `pub mod error` / `pub mod condition` shadow the `error!` / `condition!` macros under globs or direct imports (audit's recommended option 1). I rewrote the aliases to forward raw token trees to `$crate::error!` / `$crate::condition!` instead of duplicating the four-arm expansions, which removes the drift hazard if the originals ever change. Verified the path-qualified delegation resolves past the module shadow with the PR's own `macro_aliases` test (imports the modules and the aliases together). Also steered the `miniextendr-api/CLAUDE.md` collision gotcha at the aliases, which the #1105 plan asked for and the PR had applied only to `docs/CONDITIONS.md`.
- **#1096 (RngGuard unsafe contract): sound.** `new()` is now `unsafe` with a real `# Safety` note, `Default` is gone, and `with_rng` stays safe with a `debug_assert!(is_r_main_thread())` before the `unsafe { RngGuard::new() }`. Confirmed `worker::is_r_main_thread()` is not feature-gated, so `with_rng` compiles under every feature set, and the `#[miniextendr(rng)]` codegen is correctly untouched (it emits raw Get/PutRNGstate).
- **#1118 (impl method-tag const collision): sound.** The monotonic `AtomicUsize` block id is order-independent (`fetch_add`), so the emitted const names are unique within a crate compilation regardless of expansion order; the consts are dead `#[doc(hidden)]` internal items, so name non-determinism is harmless. The issue endorsed either a counter or a hash, and the counter is simpler than the plan's hash. All four call sites take the block id; the new compile-pass trybuild (inherent + trait impl on one type) compiles, and the unit test asserts distinct block ids yield distinct names. The dead-nudge follow-up is correctly tracked in #1206.

## Verification
- `cargo test -p miniextendr-api -p miniextendr-macros`: green except the known local-rustc trybuild snapshot mismatches (see Notes). New #1118 pass-test compiles, #1118 unit test passes, `macro_aliases` passes, `refcount_protect` passes 16/16.
- `cargo clippy -p miniextendr-api -p miniextendr-macros --all-targets -- -D warnings`: clean (after the fix below).
- `cargo fmt --all --check`: clean.

## Pushed (on top of the PR's three commits)
- `2e3be2f7` delegate `rust_error!` / `rust_condition!` to the originals + point the CLAUDE.md gotcha at the aliases (#1105).
- `fc80d2f1` drop `RefCountedArena`'s `Default` impl (sibling of #1096, see below).
- `b3143fc1` rewrite the ALTREP iter cache-fill with `?` (pre-existing `clippy::question_mark` surfaced by local rustc 1.97 in a file this PR does not otherwise touch; blocked `-D warnings`; semantically identical).

## Sibling findings
- **#1096:** swept every `impl Default` in `miniextendr-api`. `RefCountedArena` (`refcount_protect.rs`) had the identical foot-gun: `default()` wrapped `unsafe { Self::new() }` behind a safe API, and `new()` allocates an R VECSXP (`Rf_allocVector` / `R_PreserveObject`), main-thread-only. `default()` is unused workspace-wide (all sites use `new()` / `with_capacity()`), and removal is clippy-clean because `new()` is `unsafe`. Fixed in `fc80d2f1`. `WorkerUnprotectGuard::new(n)` is correctly safe (no R contact until `Drop` routes through `with_r_thread`); the other `Default` impls all wrap a safe `new()`.
- **#1105:** the crate root has no `pub mod warning` / `pub mod message`, so `warning!` / `message!` genuinely have no clash and were correctly left without aliases.
- **#1118:** swept the codegen'd per-type statics. `__mx_class_name_entry_<type>` (`miniextendr_impl.rs:3313`) is keyed only on the type and ignores the impl `label`, unlike `r_wrappers_const` (`R_WRAPPERS_IMPL_<TYPE>_<LABEL>`) and `c_wrapper_ident`. So two labeled `#[miniextendr]` impl blocks on one type (a documented feature, `CLASS_SYSTEMS.md`; labels are required by MXL009) collide on E0428, a distinct symbol from the tag-const this PR fixes. The PR's own trybuild comment documents it but it was untracked, so I filed **#1242** (needs a dedup-vs-disambiguate design call, out of scope here).

## Notes
- Local shows 14 trybuild `[should fail] ... mismatch` snapshots (`derive_dataframe_*` plus `fn_implicit_dots_name_conflict`), up from the 7 in prior notes. All are local-rustc-1.97-vs-CI-stable note-format drift in code neither this PR nor my changes touch; CI is authoritative, and I did not `TRYBUILD=overwrite` them.
- Deferred and tracked, not fixed here: dead-nudge activation (#1206) and the label-blind class-name-entry collision (#1242).

---

</details>

Issue #1113, opened 2026-07-01
feature-legs: broaden r6-default / s7-default legs from fixture-only to the full testthat suite


CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h3>AI-written details</h3></summary>

## Summary
- The scheduled `feature-legs` CI job (plan `plans/2026-07-01-denylist-feature-runtime-ci.md`, audit A5) runs the `r6-default` and `s7-default` legs with `testthat::test_local(filter = "feature-defaults")` instead of the full suite.
- The plan originally said to run the full suite per leg and triage failures as bugs. That is not workable for these legs: rpkg has ~52 bare `#[miniextendr] impl` blocks whose class system flips wholesale under `r6-default`/`s7-default` (Env constructor shapes disappear), and `coerce-default` changes the accepted input types of every `bool`/`f32`/multi-source parameter (#1112). The resulting failures are by-design semantic differences, not triage-able bugs, so a full-suite run would be permanently red.
- Current coverage on those legs: package build, wrapper regeneration, package load under the flipped default, plus the divergence assertions in `rpkg/tests/testthat/test-feature-defaults.R` (class probe, strict pair, coerce pair, worker pair).

## What broadening would take
- Either class-system-aware guards in every test that touches a bare impl block (large, noisy), or
- Making the existing class fixtures explicit (`#[miniextendr(env)]` etc.) so the suite is default-agnostic, leaving only deliberate probes sensitive to the flip. This is the cleaner path and also makes rpkg document which fixtures are class-system-relevant.
- The `extras` and `worker-default` legs already run the full suite, so this issue only covers the two class-flipping legs (plus `strict-default`/`coerce-default`, which ride on the s7 leg).

---
_Drafted by Claude (claude-fable-5). Reviewed by the author._

</details>

CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h4>AI-written details</h4></summary>

## Summary
- Heads-up on two upstream plans that chip at this issue's precondition (making rpkg fixtures class-explicit so the full suite is default-agnostic):
  - The #1114 fix reverts the s7 env-pins (`NdVec` and friends) and adds explicit s7 fixtures for the collision path.
  - The #1112 fix (coerce widening: bare acceptance first, native+`TryCoerce` fallback) removes the "coerce changes accepted input types" failure class entirely; the `test-feature-defaults.R` pins flip to superset assertions.
[...]

Issue #669, opened 2026-05-20
docs: surface the `fast` / `no_preconditions` / `no_call_attribution` knobs in CLAUDE.md and user docs


CGMossa:
## Context

Commit e3279011 added three new \`#[miniextendr]\` knobs but they're currently undocumented in user-facing docs. See \`analysis/scaffolding-perf-roadmap.md\` (P11–P12).

## What

Add documentation entries in:

1. \`CLAUDE.md\` — note the new knobs alongside existing fn-level options, with the headline measurement (8.25× speedup for 1-arg trivial body).
2. \`miniextendr-macros/CLAUDE.md\` — note that \`no_preconditions\` / \`no_call_attribution\` thread through \`MethodContext\` for class-system codegen.
3. The mdbook docs under \`docs/\` (if applicable) — a dedicated short section on the speed/UX trade-off and when to reach for \`fast\`.
4. \`miniextendr-fn.rs\` proc-macro docstring already lists the new flags (in commit e3279011) but doesn't cite the analysis docs. Could link to them.

## Why deprioritised

The fixture file \`rpkg/src/rust/fast_fixtures.rs\` already documents the knobs by example, and the proc-macro emits good error messages when users mis-spell flag names (commit e3279011 updated \`FN_BOOL_FLAGS_HELP\`). The fact-finding/analysis docs in \`analysis/\` are the system of record. So discoverability isn't broken — just less polished than it could be.

## Effort estimate

~30 min.

## Related

- Roadmap: \`analysis/scaffolding-perf-roadmap.md\` (P11/P12)
- Parent commit: e3279011
- Analysis chain: \`analysis/scaffolding-*\`

CGMossa:
> 🤖 **Automated triage pass** (Claude Code loop). Investigated; can't be done as specified — details below, AI-generated.

This can't be completed as a docs task because the `fast` / `no_preconditions` / `no_call_attribution` knobs **were never merged into `main`** — so there's nothing to document yet. The introducing commits (`e3279011`, `ea4d22b4`) are dangling objects, not ancestors of any ref; the working tree has zero occurrences of the flags; and `FN_BOOL_FLAGS_HELP` would reject all three with a compile error (accepted flags today: `invisible, visible, check_interrupt, worker, no_worker, coerce, no_coerce, rng, unwrap_in_r, strict, no_strict, internal, noexport, export`).

[...]

PR #1210, opened 2026-07-08
feat(macros): fast knob codegen base + fast-default cargo feature (#666)

### LINKED ISSUE #666: perf: `default-fast` cargo feature for project-wide fast-path opt-in
## Context

Follow-up from commit e3279011 which added per-fn / per-impl \`#[miniextendr(fast)]\` knobs. See \`analysis/scaffolding-perf-roadmap.md\` (P8).

## What

A \`default-fast\` cargo feature in \`miniextendr-macros\` (and a corresponding pass-through in \`miniextendr-api\`), parallel to the existing \`default-strict\` / \`default-coerce\` features. When enabled, every \`#[miniextendr]\` fn and impl block treats \`fast\` as the default — i.e. \`no_preconditions: true\` and \`no_call_attribution: true\` unless explicitly overridden per fn.

## Why this is small

The plumbing is already there. Look at how \`default-strict\` flows in \`miniextendr_fn.rs\`:

\`\`\`rust
strict: strict.unwrap_or(cfg!(feature = "default-strict")),
\`\`\`

Doing the same for \`no_preconditions\` and \`no_call_attribution\` is a 3-line change in two places (\`miniextendr_fn.rs\` and \`miniextendr_impl.rs\`), plus the feature declaration in two Cargo.toml files.

## When to do it

When someone wants project-wide opt-in without annotating every fn — e.g. a benchmark-heavy crate, or an internal framework crate where users never see error UX.

## Effort estimate

~1 hour including tests.

## Related

- Roadmap: \`analysis/scaffolding-perf-roadmap.md\` (P8)
- Parent commit: e3279011
- Sibling features: \`default-strict\`, \`default-coerce\`, \`default-r6\`, \`default-s7\`, \`default-worker\`

CGMossa:
## Summary

Re-lands the `fast` / `no_preconditions` / `no_call_attribution` / `no_fast`
knobs on `#[miniextendr]` fns and impl blocks, plus the `fast-default` cargo
feature — this time rebased onto current `main` with the `error_direct`
dependency it was previously stacked on **surgically removed**.

## Background — why this is a "re-land"

This feature has been attempted twice and shipped zero times:

1. **2026-05-20** (`e3279011`/`ea4d22b4`): written directly, never opened as a
   PR — dangling commits, unreachable from any ref today.
2. **PR #950 → #1018** (this branch's origin, `feat/g6-fast-default-666`):
   #1018 fully re-implemented the fast knobs against current macro internals.
   It was stacked on top of #950 (`error_direct`, issue #665) purely for
   review-sequencing convenience. On 2026-07-01 `error_direct` was declined
   ("would need a bunch of analysis and audits to ensure that this strategy
   is as good as the current condition-object signals") and **both PRs were
   closed unmerged** — #1018's fast-knob content was never itself objected
   to, it died as collateral from the stack.

Full investigation, timeline, and root-cause analysis in
`reviews/2026-07-08-fast-knob-vaporware-and-stacked-pr-collateral.md`
(included in this PR).

## What this PR does

- Rebases `origin/feat/g6-fast-default-666` (PR #1018's head) onto current
  `main` (86 commits behind → caught up).
- Reverts the two `error_direct` commits it was stacked on, and removes the
  `fast_and_error_direct` fixture/test/docs section the fast-knob commit had
  added on top of `error_direct` — this re-land is fast-knob **only**.
  `error_direct`/#665 is not re-attempted; the maintainer's objection stands.
- Corrects `analysis/scaffolding-perf-roadmap.md`'s 2026-06-13 status note,
  which incorrectly asserted both PRs had already landed.

## What ships

Four new `#[miniextendr]` options (fn- and impl-level):

- `no_preconditions` — drop the generated `stopifnot(...)` block. `TryFromSexp`
  still rejects bad input; the error message comes from Rust instead of R.
- `no_call_attribution` — emit `.call = NULL` instead of `.call = match.call()`.
  R's `stop()` fills in `sys.call()` on the error path, so `conditionCall(e)`
  stays non-NULL.
- `fast` — bundle alias for both (7.78x/8.54x measured on 1-/3-arg standalone
  fns, 2.19x/3.36x on R6 method dispatch — see
  `analysis/scaffolding-deep-findings-2026-05-20.md`).
- `no_fast` — explicit opt-out, for use alongside the crate-wide
  `fast-default` cargo feature.

New `fast-default` cargo feature in `miniextendr-macros` (forwarded by
`miniextendr-api`), added to both `full-codegen` and `full-codegen-s7` so
CI's `clippy_all`/`clippy_all_s7` exercise it automatically.

Fixture: `rpkg/src/rust/fast_fixtures.rs` (6 standalone fns + `FastCounter`/
`FastCounterFast` R6 classes). Tests:
`rpkg/tests/testthat/test-fast-fixtures.R` (13 `test_that` blocks).

## Verification (local)

- `just check` — full workspace + rpkg/src/rust, green
- `just devtools-test` — `[ FAIL 0 | WARN 0 | SKIP 27 | PASS 7037 ]`
- `cargo fmt --all -- --check` — clean
- `clippy_default`, `clippy_all` (`full-codegen` — includes `fast-default`
  crate-wide), `clippy_all_s7` (`full-codegen-s7`) — all `-D warnings` clean
- UI tests: 7/138 pre-existing local-rustc-drift failures (unrelated
  `derive_dataframe_enum_*` snapshots — confirmed identical failure set on
  unmodified `main`; not caused by this change)

Closes #666.

---


CGMossa:
Re-verified this one end to end, with extra skepticism because #1212 was flagged as a possible "this is vaporware" audit. It isn't in conflict: #1212's audit is about the *historical* never-landed state of these knobs and it explicitly names this PR as the re-land, so the two are complementary. The knob here is real, reachable, and tested. I traced the whole path, ran both feature positions, and pushed one pre-existing clippy fix that was blocking the macros lint check. One genuine CI gap (no runtime leg for the crate-wide default) is filed as #1244.

<details>

## Summary

- **Verdict: sound / real, not vaporware.** PR #1212's "vaporware audit" describes the two *prior* abandoned landings of these knobs (dangling commits, then #1018 dying as collateral under the declined `error_direct` stack); its own body says "the actual re-land is #1210", and the two PRs even share the same review file. I did not take that on faith and traced reachability myself: the `fast-default` cargo feature is declared in `miniextendr-macros`, forwarded by `miniextendr-api/fast-default`, and both `no_preconditions` / `no_call_attribution` resolve via `unwrap_or(cfg!(feature = "fast-default"))` in `MiniextendrFnAttrs` and `ImplAttrs` (exactly parallel to the existing `coerce-default`). Those two booleans then change real emitted output: `no_preconditions` returns an empty precondition prelude (drops the `stopifnot(...)` block) and `no_call_attribution` emits `.call = NULL` instead of `.call = match.call()`. The impl path routes through `MethodContext::with_fast_flags(...)` across all six method-context iterators plus `static_call` / `instance_call`, so every class system inherits it. A user can turn a knob (`#[miniextendr(fast)]`, `no_preconditions`, `no_call_attribution`, `no_fast`) or flip it crate-wide with the feature. This is a knob that changes emitted code, not scaffolding.

- **Both knob positions verified.** `cargo test -p miniextendr-macros --lib` = 444 passed; with `--features fast-default` = 445 passed. That second run executes the two `#[cfg(feature = "fast-default")]` resolution tests (`fast_default_fn_attrs_resolve_both_true`, `fast_default_no_fast_opt_out_restores_false`) that **no CI job runs today** (see gap below), and they pass. The R fixtures cover both positions with a genuinely differential assertion: `fast_i32_no_preconditions("not an int")` asserts a `rust_error` with the Rust conversion message, which only holds if the `stopifnot` block is actually gone; `fast_i32_default` asserts the R `stopifnot` text. Same default-vs-fast split for the `FastCounter` / `FastCounterFast` R6 pair.

- **Feature matrix + clippy.** `cargo clippy -p miniextendr-macros --all-targets -- -D warnings` is green both with and without `fast-default`. `fast-default` is wired into `full-codegen` and `full-codegen-s7`, so `clippy_all` / `clippy_all_s7` compile it (with `--all-targets`, that includes the cfg-gated test targets) and it will not rot as dead code. It is orthogonal to the `r6-default` / `s7-default` mutex, correctly living in both aggregates. `cargo fmt --all --check` clean.

- **UI snapshots checked, not rubber-stamped.** The 14 `.stderr` changes are all mechanical and correct: 13 `derive_dataframe_*` files gain `fast-default` in rustc's "expected values for `feature`" note (required once the feature is declared), and `unknown_miniextendr_option.stderr` gains the new option names. I confirmed no other line moved, so the known local-rustc E0080 drift was not baked into these snapshots. They will match on CI.

- **Pushed (`b13d0f34`):** `fix(api): rewrite exhausted-iterator guard with ?` in `miniextendr-api/src/altrep_data/iter/state.rs`. This is a pre-existing `clippy::question_mark` (last touched by #1191, unchanged vs main) that surfaces because `--all-targets` pulls `miniextendr-api` in as a dev-dependency of the macros crate; under `-D warnings` it aborted the build before the macros crate itself was linted. Fixed per the repo's "fix warnings you see, even if pre-existing" principle; also unblocked the macros clippy verification.

## Notes

- **Filed #1244** for the one genuine CI gap: `fast-default` is compiled but its crate-wide-default *runtime* effect is never executed. Unlike `strict-default` / `coerce-default`, it has no leg in the weekly `feature-legs` matrix, and `cargo test --workspace` runs with the feature off. The explicit-knob path *is* runtime-tested per-PR via `test-fast-fixtures.R`, so this only concerns the default flip. Same family as #1113.
- Minor, not blocking: the PR body says the review file is "included in this PR", but it is actually in sibling #1212 (which lists it under "no code changes here"). The `no_call_attribution` R test is not differential on its own axis (both positions yield a non-NULL `conditionCall` on the error path), so that knob's distinct behavior is proven by code inspection plus the macro-level parse tests rather than a runtime assertion; the `DotCallBuilder` null-attribution path it delegates to is separately unit-tested.
- Siblings: #666 is closed by this PR. #667 (should `internal` imply `fast`?) was blocked only on these knobs existing and is now answerable; its own thread argues against auto-coupling, so no code is owed here. #669 already became post-mortem PR #947. Blocked #667 stays open as a design question, not a defect.

---

</details>

Issue #668, opened 2026-05-20
perf-research: S7 method dispatch is ~4× R6 — attribute the cost via Rprof


CGMossa:
## Context

Follow-up from commit e3279011 and the bench work in \`analysis/scaffolding-bench-2026-05-20.md\`. See \`analysis/scaffolding-perf-roadmap.md\` (P10).

## Observation

Class-system method dispatch bench (1-arg \`&self -> i32\`, no real work):

| Class system | min (ns) |
|---|---:|
| bare fn (\`conv_i32_ret()\`) | 1435 |
| R6 (\`obj\$value()\`) | 2132 |
| S4 (\`s4_value(s4_obj)\`) | 2337 |
| Env (\`obj\$get_value()\`) | 2788 |
| **S7 (\`s7_value(s7_obj)\`)** | **9061** |

S7 is **~4× slower than R6** for the same trivial method dispatch. Adding \`fast\` to the impl block (commit e3279011) drops the wrapper layer but doesn't touch the S7 dispatch itself.

## What's needed

\`Rprof()\` drill-down over a tight loop of \`s7_value(obj)\` calls to identify whether the cost driver is:

1. Argument matching in S7's generic dispatch
2. Validator running on the property accessor
3. Class ancestry walk
4. Method-table lookup
5. The \`function(x, ...) S7::S7_dispatch()\` wrapping itself
6. Property getter invocation

Then either:
- Open an upstream PR to the S7 package if it's an S7-side fix
- Document the cost so users can route around it (e.g. cache the resolved method as a local closure)
- Investigate whether miniextendr could emit a hand-cached dispatch closure that bypasses S7's generic resolution

## Why deprioritised

S7 is a third-party package; miniextendr can't directly optimise it. This is mostly an upstream / documentation issue. But it's worth a thorough investigation so we have a clear answer for users asking "why is S7 slow?".

## Effort estimate

~1-2 hours of investigation + a documentation update.

## Related

- Roadmap: \`analysis/scaffolding-perf-roadmap.md\` (P10 / M5)
- First-pass measurement: \`analysis/scaffolding-bench-2026-05-20.md\` § 6
- Background: \`background/S7-main/\` (reference S7 source)

CGMossa:
> 🤖 **Automated triage pass** (Claude Code loop). Investigated, attributed the cost, opened a draft analysis PR; summary below is AI-generated.

[...]

PR #1212, opened 2026-07-08
fix(bench): repair broken R-side benchmark suite + fast-knob vaporware audit

CGMossa:
## Summary

- Fixes a promise-memoization bug shared by 4 hand-rolled `bench_one()`
  helpers (`bench-class-dispatch.R`, `bench-dots.R`,
  `bench-altrep-vs-sexp.R`, `bench-vctrs-protocol.R`): referencing the
  `expr` argument directly forces the R promise once and caches the
  value, so every "iteration" after warmup just re-read the cached
  result instead of re-invoking the call. All four scripts had been
  reporting near-zero, undifferentiated timings for their entire
  history. Fixed via `substitute()`/`eval()`.
- `bench-class-dispatch.R`: fixes unqualified calls to
  `#[miniextendr(internal)]` (noexport) `S4Counter`/`S7Counter` symbols
  (needed `miniextendr:::`), adds an explicit "R-side dispatch tax"
  table (get()/mutate() minus the plain-`.Call` baseline) to surface
  the R-vs-Rust split the script already implied.
- `bench-vctrs-protocol.R`: fixes calls to a nonexistent
  `DerivedPercent$new()`-style API — the real constructors are plain
  functions (`new_derived_percent()`, etc.) — plus an int/double arg
  mismatch.
- `bench-altrep-visual.R`: `ggplot2::autoplot()`'s default beeswarm
  plot type needed the (previously uninstalled) `ggbeeswarm` package,
  which halted `just bench-r` before it ever reached the later scripts
  alphabetically. Added via `rv add` (`rproject.toml`/`rv.lock`).
- `rproject.toml`: primary repository now points at the internal
  `stratus` snapshot mirror, CRAN kept as a fallback for packages
  stratus doesn't carry (`ggbeeswarm`+deps). `roxygen2` now resolves to
  `8.0.0` directly from stratus instead of a hardcoded CRAN tarball URL.
- `reviews/2026-07-08-fast-knob-vaporware-and-stacked-pr-collateral.md`:
  investigation into why `analysis/scaffolding-fast-bench.R` referenced
  nonexistent functions — traces two abandoned landings of the
  `fast`/`no_preconditions`/`no_call_attribution` knobs and finds the
  second died as collateral from being stacked under a separately-declined
  `error_direct` PR, not from any objection to its own content. Investigation
  only, no code changes here — the actual re-land is #1210.

`just bench-r` now runs to completion (exit 0) for the first time.

## Test plan

- [x] `Rscript rpkg/tests/testthat/bench-class-dispatch.R` runs clean,
      shows real per-system numbers (S7 ~4x R6/S3/S4, matching #668)
- [x] `Rscript rpkg/tests/testthat/bench-dots.R` / `bench-vctrs-protocol.R` /
      `bench-altrep-vs-sexp.R` run clean with plausible, differentiated timings
- [x] `just bench-r` (full suite) exits 0
- [x] `just configure && just rcmdinstall && just force-document` — no
      NAMESPACE/man drift

---

### DIFF
--- a/reviews/2026-07-08-fast-knob-vaporware-and-stacked-pr-collateral.md
+++ b/reviews/2026-07-08-fast-knob-vaporware-and-stacked-pr-collateral.md
@@ -23,6 +23,21 @@ since the second abandonment three weeks ago. The code for attempt #2 is not
 lost — it is sitting on `origin/feat/g6-fast-default-666`, 86 commits behind
 `main`.
 
+## Update (2026-07-09): attempt #3 is now in flight as #1210
+
+Since this investigation, the feature has a **third** landing: PR #1210
+(`feat(macros): fast knob codegen base + fast-default cargo feature (#666)`,
+opened 2026-07-08) does exactly the first "Fix" option below — it rebases the
+`00c8ee01` / `3985865b` commits onto `main` standalone, off the declined
+`error_direct` stack. An independent review traced it end to end and judged it
+real and sound: `fast-default` reachably flips `no_preconditions` /
+`no_call_attribution` through `MethodContext::with_fast_flags` across all six
+class systems, with differential R fixtures for both knob positions. So the
+"silently blocked / shipped zero times" framing below is the **pre-#1210**
+state, not a standing TODO — the re-land is in review (still open, not yet
+merged). The loose end #1210 does not itself close is issue #669, still closed
+against its own recommendation.
+
 ## Timeline
 
 1. **2026-05-20** — `Mossa` (the maintainer) writes `e3279011` (fn/impl-level
@@ -122,7 +137,8 @@ clean cherry-pick.
 ## Fix
 
 Not applied in this pass — this review documents the investigation only.
-Options for whoever picks this up (flat, no priority implied):
+**The first option below is now in flight as #1210** (see the 2026-07-09 update
+above). Remaining options, flat, no priority implied:
 
 - Rebase `00c8ee01`/`3985865b` onto `main` standalone (drop the `error_direct`
   dependency entirely), reopen #666/#1017 against it, close #669 for real
--- a/reviews/2026-07-08-fast-knob-vaporware-and-stacked-pr-collateral.md
+++ b/reviews/2026-07-08-fast-knob-vaporware-and-stacked-pr-collateral.md
@@ -1,0 +1,148 @@
+# The `fast`/`no_preconditions`/`no_call_attribution` knobs — two abandoned
+# landings, an issue closed against its own recommendation, and stale docs
+# that still describe them as shipped
+
+**Date:** 2026-07-08
+**Area:** `miniextendr-macros` fn/impl attribute parsing, issue tracker hygiene,
+`analysis/scaffolding-*` evidence chain
+**Trigger:** benchmark-audit investigation (asked to go deeper on the finding
+that `analysis/scaffolding-fast-bench.R` references nonexistent functions)
+
+## TL;DR
+
+The `fast` knob (`no_preconditions` / `no_call_attribution` / `fast` /
+`no_fast` on `#[miniextendr]` fns and impl blocks) has been implemented
+**twice** and is on `main` **neither** time. It is not dead because the idea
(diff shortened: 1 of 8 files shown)


CGMossa:

<details>

## Summary

- **Bench repair: sound, verified end to end.** The core fix targets a real R promise bug: the four hand-rolled `bench_one(label, expr)` helpers referenced `expr` directly, which forces the promise once and caches its value, so every post-warmup "iteration" re-read the cached result instead of re-invoking the call. Timings had been near-zero and undifferentiated for the whole history of these scripts. The `substitute(expr)` + `eval(call, parent.frame())` rewrite is the correct fix and is applied identically across all four affected files (`bench-class-dispatch.R`, `bench-dots.R`, `bench-altrep-vs-sexp.R`, `bench-vctrs-protocol.R`).
- **Smoke-run (R 4.6.0, installed package).** `Rscript bench-class-dispatch.R` and `Rscript bench-vctrs-protocol.R` both exit 0 with real, differentiated numbers. Class dispatch shows S7 `get()` at 11.8 us/call versus 2.7 to 3.5 for R6/S3/S4/Env, roughly the 4x gap #668 predicted, and the new "R-side dispatch tax" table computes correctly (S7 +7.3/+8.7 us, the others near zero, exactly what the arg-shape caveat comment says to expect). vctrs shows `new_derived_percent` 3.5 us versus raw 0.5, `vec_c` 63 us versus `c()` 0.5, all plausible.
- **vctrs API fix confirmed against source.** `new_derived_percent` / `new_derived_point` / `new_derived_temp` are real `#[miniextendr]` functions in `rpkg/src/rust/vctrs_derive_example.rs`; `DerivedPercent::new` is only a Rust-side method with no R `$new`, so the old `DerivedPercent$new(...)` calls were genuinely broken. The `as.double(seq_len(100))` change matches `new_derived_point`'s `(Vec<f64>, Vec<f64>)` signature.
- **Audit doc: accurate historically, corrected for today.** The two abandoned landings, the #669 close-against-its-own-recommendation, and the docs-first stack-order inversion all check out (verified: #669 CLOSED, #666/#667/#1017 OPEN, `origin/feat/g6-fast-default-666` still present). But the doc left the re-land as an unclaimed "options for whoever picks this up" TODO, which now reads as "still vaporware." #1210 (opened 2026-07-08, independently reviewed real and sound) is exactly that re-land. I pushed `99e5617d`, adding a dated 2026-07-09 update note plus a Fix-section flag so the post-mortem points at #1210 instead of implying nothing has happened. The historical timeline is left intact.
- **Sibling sweep: clean.** The other five bench files (`bench-altrep-{comprehensive,performance,simple,zerocopy,visual}.R`) do not share the promise bug. They use `bench::mark` / `bench::press` or inline `for` loops with literal bodies, neither of which memoizes. `ggbeeswarm` (the dependency whose absence halted `just bench-r` alphabetically at `bench-altrep-visual.R`) is installed and loads. No other bench entry point needed fixing.

## Notes

- The `build(rv)` commit points the primary rv repository at the internal `prism.dev.a2-ai.cloud/stratus` snapshot (2026-06-18) and moves `roxygen2` off its hardcoded CRAN tarball URL onto that mirror, keeping CRAN as fallback. This is mild scope creep for a bench-repair PR, but `rproject.toml` is dev-only (not shipped to CRAN or end users), the pinned CRAN tarball URL it replaces is genuinely fragile (it 404s once `roxygen2` 8.0.0 moves to Archive), and the PR's green CI ran with it. Worth confirming that a contributor without stratus access still resolves `roxygen2` through the CRAN fallback rather than failing hard on the pinned-source lock entry.
- I ran only the two most-changed scripts plus a `ggbeeswarm` load check, not the full `just bench-r` suite, per the "quick iteration, not full statistical runs" scope. `bench-altrep-visual.R` writes PNGs and `Rplots.pdf` into the cwd via `ggsave`, so a full run would leave untracked artifacts.
- CI on this PR is green. It touches only R scripts, `rproject.toml` / `rv.lock`, and a review markdown, so the Rust legs are path-skipped. The fleet-wide red Rust-tests leg (trybuild 1.97 snapshot skew, tracked separately) does not apply here.

---

</details>

Issue #1097, opened 2026-06-28
Collect all errors in vectorized storage-directed coercion (IntoRAs)


CGMossa:
## Summary

Vectorized coercion to R storage types currently reports only the **first** failing element. The project principle is "collect all errors in vectorized ops — give users one batched diagnostic." Making `IntoRAs` (and the underlying `TryCoerce` vector path) collect *all* element errors would align it with that principle, but it's an error-type change worth designing deliberately.

## Background

`IntoRAs` (`miniextendr-api/src/into_r_as.rs`) is the storage-directed outbound path (caller picks the target R storage type; out-of-range/precision/NA values error instead of silently widening). For vector targets it stops at the first failing element and returns a single `StorageCoerceError` annotated with that element's index (`at_index`). See `audit/into_r_as.md` issue #1.

A user converting `Vec<i64> -> INTSXP` where elements 3, 17, and 42 are out of range sees only "index 3", fixes it, re-runs, sees "index 17", etc. — instead of one diagnostic listing all three.

## Why it's not trivial

- It changes the error contract: a batched API needs `Vec<StorageCoerceError>` (or a new aggregate error type), not a single `StorageCoerceError`.
- The companion `TryCoerce<Vec<R>> for Vec<T>` blanket is **coherence-infeasible** on stable Rust (overlaps the blanket `impl<T: Coerce<R>> TryCoerce<R> for T` via the container `Coerce<Vec<R>> for Vec<T>`, E0119) — so the batched path likely has to live as explicit methods on `IntoRAs`/free functions, not a blanket.

## Options to weigh

1. Add `into_r_as_collecting::<Target>() -> Result<SEXP, Vec<StorageCoerceError>>` alongside the existing first-error method (non-breaking; opt-in).
2. Change the existing method's error type to an aggregate (breaking, but project is pre-release / "no backwards compat").
3. Leave first-error-with-index as-is (it's a documented pragmatic compromise) and close.

## Source

Audit: `audit/into_r_as.md` issue #1; related coherence note in `audit/coerce.md` issue #4 (verified E0119 against `coerce.rs:112` + `coerce.rs:982`).

<sub>Drafted by Claude (Opus 4.8) from the automated `audit/` review; filed at the maintainer's request.</sub>

CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h4>AI-written details</h4></summary>

## Summary
[...]

PR #1216, opened 2026-07-08
feat(api): batch numeric-coercion vector conversion errors with per-element index (#1192)

### LINKED ISSUE #1192: Batched diagnostics for numeric-coercion vector conversions (from_numeric_vec_with / coerce_slice_to_vec / strict checked_vec_*)
## Context

#1143 batched the string-parse vector conversions (uuid/url/regex/num-bigint): the `Vec<T>` / `Vec<Option<T>>` arms of `try_from_sexp_via_str_parse!` now walk the whole vector, accumulate per-element failures, and return one `SexpError::InvalidValue` via the new `batch_conversion_errors()` helper (`miniextendr-api/src/from_r.rs`), capped at 10 entries + "and N more". That was a one-site change because the macro had already centralized the eight impls and its per-element errors already carried indices.

The numeric-coercion vector conversions were evaluated for the same treatment in that PR and deliberately deferred: they are **not** the same one-site accumulate-and-report shape.

## The gap

The coerced numeric vector paths still bail on the first failing element, and their per-element errors carry **no index at all** — just debug formatting of the coercion error:

- `from_numeric_vec_with()` (`miniextendr-api/src/from_r.rs`) — the shared INTSXP/REALSXP/RAWSXP/LGLSXP dispatch shell; each branch does `slice.iter().copied().map(map_x).collect()`, short-circuiting on the first `Err`.
- The per-element closures (`coerce_value` in `miniextendr-api/src/from_r/coerced_scalars.rs`, and the NA-aware closures in `miniextendr-api/src/from_r/na_vectors.rs` `try_from_sexp_numeric_option_vec`) produce `SexpError::InvalidValue(format!("{e:?}"))` — no element index, `Debug` output.
- `coerce_slice_to_vec()` (`miniextendr-api/src/from_r.rs`) — same short-circuit for the `Vec<bool>` LGLSXP leg.
- `HashSet<T>` / `BTreeSet<T>` impls route through `try_from_sexp_numeric_vec` and inherit the behavior.

So a `Vec<u32>` fed `c(-1, 5, -3)` reports only the `-1` failure, without saying which element it was.

## Sketch

1. Enumerate inside `from_numeric_vec_with` (and `coerce_slice_to_vec`): wrap each element failure as `"invalid value at index <i>: <err>"`, accumulate instead of `collect()`-short-circuiting.
2. Reuse `batch_conversion_errors(container, errors)` from #1143 for the final message (same "; "-joined, capped-at-10 shape).
3. Decide `Display` vs the current `Debug` formatting for `TryCoerce` errors while touching the message (the `{e:?}` output is developer-facing today).
4. Container label needs threading (the shell is generic over the target type; callers know `Vec<u32>` vs `HashSet<i64>` etc.), or settle for a generic label.
5. Coordinate with rpkg testthat greps: `rpkg/tests/testthat/test-scalar-conversions.R` matches the substring `"strict conversion failed"` for the strict path — that path (`miniextendr-api/src/strict.rs`, `checked_vec_*` with indexed per-element messages) is a **sibling site** that short-circuits too and could adopt the same batching in the same pass.

Multi-site, message-shape-changing — hence a follow-up rather than part of #1143.

CGMossa:

<details>

## Summary

Closes #1192. Extends the #1143 string-parse batching to the numeric-coercion vector conversion paths, which previously bailed on the first failing element and carried no element index.

- `from_numeric_vec_with` (via a new `collect_coerced` helper) and `coerce_slice_to_vec` now walk the whole vector, accumulate every per-element failure as `"invalid value at index <i>: <err>"`, and return one batched `SexpError::InvalidValue` through the shared `batch_conversion_errors` helper (same `"; "` joiner and `"; and N more"` cap at 10 as #1143).
- Container labels (`"Vec<u32>"`, `"HashSet<i64>"`, `"Vec<Option<u32>>"`, `"Vec<bool>"`, ...) are threaded from the impl macros via `stringify!`.
- Per-element coercion errors now format with `Display` (`{e}`) instead of `Debug` (`{e:?}`). The `TryCoerce::Error` bound is widened `Debug` to `Display` across the coerced scalar/vector/set paths; both `CoerceError` and `LogicalCoerceError` already implement `Display`, so nothing is truncated.
- Only vector legs batch. Scalars keep their single-error shape (now Display-formatted). `Vec<Option<T>>` keeps NA to `None` passthrough while real coercion failures batch with their original index. `HashSet<T>` / `BTreeSet<T>` inherit via `try_from_sexp_numeric_vec`.
- New unit tests in `miniextendr-api/tests/numeric_vec_batching.rs` (auto-discovered integration test): multi-index batching, the `"; and N more"` cap, Option NA passthrough vs real-failure batching, `HashSet`/`BTreeSet` inheritance, and `Vec<bool>` via `coerce_slice_to_vec`.

### Before / after

`Vec<u32>` fed `c(-1L, 5L, -3L)`:

```
# before (first failure only, Debug-formatted, no index)
invalid value: Overflow

# after (every failing index, Display-formatted, one batched diagnostic)
invalid value: Vec<u32> conversion failed: invalid value at index 0: value out of range; invalid value at index 2: value out of range
```

## Test plan

- [x] `cargo check -p miniextendr-api --all-targets --locked` clean
- [x] `cargo test -p miniextendr-api` — 53/53 suites pass, including the 8 new `numeric_vec_batching` tests
- [x] `cargo fmt --all --check` clean
- [x] clippy_default: `cargo clippy --workspace --all-targets --locked -- -D warnings`
- [x] clippy_all (`full-codegen` + full integration feature list from `.github/workflows/ci.yml`) `-D warnings`
- [x] clippy_all_s7 (`full-codegen-s7` + same list) `-D warnings`
- [x] `cargo test --workspace` green except the pre-existing local `test ui` trybuild drift (see Notes)
- [x] rpkg testthat over the affected surface (installed package): 13 files, 889 pass, 0 fail / 0 warn / 0 skip (coerce, conversions, scalar-conversions, convert-pref, r-coerce, slice-conversions, columnar-option-none, dots, match-arg, option-self, trait-method-options)
- [x] `just rcmdinstall` DONE, no NAMESPACE / man drift (runtime-only change, no codegen touched)

## Notes

- No `#[miniextendr]` codegen changed, so `NAMESPACE` / `man/*.Rd` / `wrappers.R` are unchanged.
- The `test ui` failure under `cargo test --workspace` is the documented local-toolchain trybuild drift (rustc 1.95.0 vs CI stable) on the `derive_dataframe_*` snapshots. This diff touches no macro or dataframe code and the snapshots were not modified; CI (stable toolchain) is authoritative.
- Scope was deliberately kept to the runtime `TryFromSexp` numeric-coercion vector paths named in the issue. Three sibling short-circuit sites remain out of scope and are tracked in a follow-up issue: the macro-generated argument coerce path (`rust_conversion_builder.rs` `CoercionMapping::Vec`, which is the path a bare `Vec<u32>` function argument actually takes), the strict `checked_vec_*` path, and the `tinyvec` `coerce_slice_to_vec`. That issue also captures the deferred design question of whether the non-Option coerced `Vec<T>` path should reject NA (today it round-trips the sentinel per its documented contract, and the R-level precondition is intentionally NA-safe, so changing it is a user-visible behaviour decision rather than a diagnostics change).
- Linkage: this PR's message grammar and the `batch_conversion_errors` helper are the template for #1097 (`IntoRAs` outbound batching); #1097 is intentionally not touched here.

---

</details>


CGMossa:
Rebased onto current main and re-reviewed. The implementation is sound and faithful to #1192; I pushed one hardening commit (153de174) that bounds the batched-error memory so a large all-failing vector can no longer build a giant `Vec<String>` just to show 10 entries. Full findings below.

## Summary

Verdict: faithful, well-tested implementation of #1192. It extends the #1143 batching grammar to the numeric-coercion vector shells (`from_numeric_vec_with` via the new `collect_coerced`, and `coerce_slice_to_vec`), threads a concrete container label from each impl via `stringify!`, and switches per-element formatting from `Debug` to `Display`.

- Happy-path cost: clean. The result `Vec` is a single `with_capacity`, same as the old `.collect()`. The error accumulator allocates nothing until the first failure, so the hot no-error conversion loop is unchanged.
- Error-list explosion: found one real defect. The display was capped at 10, but the callers first built the full `Vec<String>` (one heap string per failing element), so an all-failing N-element vector materialised N strings only to discard all but 10. This was shared with the pre-existing #1143 str-parse arms, which used the same helper. Commit 153de174 replaces the free `batch_conversion_errors(container, Vec<String>)` with a `BatchedErrors` accumulator used by all four batching sites (both `try_from_sexp_via_str_parse!` arms and both numeric shells). It retains at most `BATCHED_ERROR_CAP` messages and counts the rest; beyond the cap the per-element `format!` closure is never invoked. The emitted message is byte-identical, pinned by the existing uuid and numeric cap tests (which fail if `push` stops capping, since `into_error` joins whatever it retained).
- Indexing: 0-based, matching the already-merged #1143 subsystem. I did not flip it to 1-based because that would desync from #1143 and its snapshot tests. Flagging it as a possible R-user papercut for a future holistic decision, not a bug in this PR.
- strict vs coerce parity: correct. Only the coerce path batches, as intended. The strict path (`strict.rs` `checked_vec_*`) still short-circuits, but it is panic-based (returns `T`, not `Result`) and `Debug`-formatted, so batching it is not mechanical; it is legitimately deferred.
- NA vs failure: `Vec<Option<T>>` passes NA through as `None` while real failures batch with their original index; the non-Option `Vec<T>` path treats the NA sentinel as a coercion failure, per its documented contract.
- Debug to Display switch: verified safe. No unit or testthat test greps the `Debug` spellings (tests compare `CoerceError::Overflow` as an enum value, never as a message substring), and both `CoerceError` and `LogicalCoerceError` implement `Display`.

Reach caveat worth surfacing: under a package with no `coerce-default` (rpkg's default, and non-`#[miniextendr(coerce)]` methods), a bare `Vec<u32>` argument DOES take the batched path, and `Vec<Option<T>>` / `HashSet` / `BTreeSet` arguments always do. But when coercion is active, a `Vec<numeric>` / `Vec<bool>` / `Vec<f32>` argument takes the inline `rust_conversion_builder.rs` `CoercionMapping::Vec` codegen path, which still short-circuits on the first failure with no index. This is honestly disclosed in the PR Notes.

Verification (local): `cargo fmt --all` clean; `cargo test -p miniextendr-api` green (298 unit tests + every integration suite, including the 8 `numeric_vec_batching` and, under `uuid,url,regex,num-bigint`, the 12 `str_parse` cases); `cargo clippy -p miniextendr-api -p miniextendr-macros --all-targets -D warnings` clean on both default and the str-parse featured build (needed to type-check the exported-macro arms I touched).

## Notes

- Sibling sweep: #1217 already tracks the three remaining short-circuit sites (macro `CoercionMapping::Vec`, strict `checked_vec_*`, tinyvec `coerce_slice_to_vec`) plus the non-Option NA-policy decision, so no new issue was filed.
- Runtime-only change (`from_r.rs`): no `#[miniextendr]` codegen touched, so `NAMESPACE` / `man/*.Rd` / `wrappers.R` are unchanged, matching the base PR.
- The fleet-wide Rust-tests trybuild red is unrelated to this PR and heals when #1239 lands.

</details>

Issue #1218, opened 2026-07-08
Cross-package producer.pkg/consumer.pkg wrappers.R are stale (@title -> @description drift)


CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h4>AI-written details</h4></summary>

## Summary
- `tests/cross-package/producer.pkg/R/producer.pkg-wrappers.R` and `consumer.pkg/R/consumer.pkg-wrappers.R` are **tracked** generated files that have drifted from current `#[miniextendr]` codegen.
- Regenerating them via `just cross-install` produces ~165 lines of `#' @title` -> `#' @description` swaps in the S3Point/S4Point **inherent-impl** wrappers (plus surrounding context). This is unrelated to trait codegen; it reflects a prior inherent-impl/roxygen macro change that never had the committed cross-package wrappers regenerated.
- Found incidentally while working on #1141/#1115 (which touches only trait-impl codegen). Those wrapper changes were reverted there to keep that PR scoped.

## Notes
- Low severity: the installed producer/consumer packages regenerate their wrappers.R at install time, so cross-package tests pass regardless of the committed snapshot.
- Fix is a one-shot regenerate-and-commit of the two cross-package wrappers.R. Consider whether CI should gain a cross-package wrappers-sync-check analogous to rpkg's `just wrappers-sync-check` to prevent recurrence.

---
_Drafted by Claude (claude-opus-4-8). Reviewed by the author._

</details>

Issue #1115, opened 2026-07-01
r6 trait-impl wrapper names are not class-qualified — two r6 impls of one trait collide (r6_trait_<Trait>_<method>)


CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h3>AI-written details</h3></summary>

## Summary
- R6-style trait impls generate wrapper functions named `r6_trait_<Trait>_<method>` with no class component. Two `#[miniextendr(r6)] impl <Trait> for <Type>` blocks for the *same trait* on *different types* therefore emit identical wrapper names.
- The duplicate-definition guard in `miniextendr-api/src/registry.rs:1381` catches it at wrapper-generation time and aborts the install, but its message only mentions the S7 shortcut case, which is misleading here.
- Found by the feature-legs runtime CI (audit A5): under `r6-default`, rpkg's two bare (env) `Counter` trait impls (`SimpleCounter`, `PanickyCounter`) flipped to r6 and collided with `R6TraitCounter`'s explicit r6 trait impl. But the bug does not require feature flips: two explicit `#[miniextendr(r6)]` impls of one trait reproduce it directly.
- Interim mitigation in rpkg: the two bare `Counter` trait impls are pinned to `#[miniextendr(env)]`.

## Options
- Qualify the r6 trait wrapper names with the implementing class (`r6_trait_<Class>_<Trait>_<method>`), mirroring how S7 shortcuts are class-qualified, or
- If the wrapper is intentionally shared per-trait, dedupe identical definitions at generation time instead of aborting.
- Either way, extend the registry.rs duplicate-name panic message to mention the r6 trait-impl case, not just S7 shortcuts.

---
_Drafted by Claude (claude-fable-5). Reviewed by the author._

</details>

CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h4>AI-written details</h4></summary>

## Summary
- This will be fixed as part of #1141 (trait-method namespace unification), not as a standalone rename. The unification moves R6 instance trait methods from standalone `r6_trait_<Trait>_<method>()` functions to the class-scoped `Type$Trait$method` shape, which makes this collision impossible by construction.
- Of the two options listed here, that is effectively option 1 taken to its conclusion: instead of qualifying the flat name with the class, the flat name disappears entirely.
[...]

PR #1219, opened 2026-07-08
feat(macros): unify trait-method namespace shapes + Self-return re-wrapping (closes #1141, #1115)

CGMossa:

<details>

## Summary
- Routes all 5 trait-impl R-wrapper generators (env/s3/s4/s7/r6) in `miniextendr_impl_trait/r_wrappers.rs` through one namespace-shape owner, `trait_namespace_target` (exposed as `TraitMethodContext::namespace_target`), and deletes the 5 inline hand-rolled namespace decisions. Root cause from `audit/2026-07-03-dogfooding-macros-codegen.md` finding #1.
- **#1115**: R6 instance trait methods move from the unqualified standalone `r6_trait_<Trait>_<method>(x)` into the class-scoped `Type$Trait$method(x)` namespace (the Env/S3 shape). Class-scoped names are collision-free by construction, so two R6 impls of one trait on different types no longer emit identical wrapper names and abort wrapper-gen via the duplicate-definition guard. Breaking rename, no backcompat (unreleased project).
- **#1141 resp-4**: `-> Self` / `Result<Self, E>` / `Option<Self>` trait methods (instance and static) now re-wrap the bare `ExternalPtr` into a classed object through the shared `MethodReturnBuilder` (`TraitMethodContext::method_body_lines`), exactly as inherent constructors do, instead of leaking an unclassed pointer.
- Generalized the registry duplicate-wrapper panic message (`miniextendr-api/src/registry.rs`) to name trait impls, not just the S7 shortcut case.
- Reverted #1115's interim env-pins to bare: SimpleCounter/PanickyCounter (`Counter`) in `trait_abi_tests.rs`; SharedSimpleCounter/AtomicCounter (`SharedCounter`, inherent+trait pair) in `shared_trait_test.rs`. Default-build codegen for these is unchanged (bare defaults to env); the pins only masked the feature-leg flip that #1115 now makes safe.
- Added fixture `rpkg/src/rust/trait_r6_collision.rs` (two explicit `#[miniextendr(r6)]` impls of one trait + `-> Self` factory methods across r6 and env) and `test-trait-r6-collision.R`.
- Regenerated `NAMESPACE` + `man/` (8 `r6_trait_*` exports removed; `DoublerA/B/Env` added; R6 trait man pages switched to namespace-form aliases).

## Namespace policy (before / after)

| Class system | Trait member placement | Before | After |
|---|---|---|---|
| Env / S3 (static + namespace instance) | `Type$Trait$method` | `Type$Trait$method` | `Type$Trait$method` (now via shared helper) |
| R6 static | `Type$Trait$method` | `Type$Trait$method` | `Type$Trait$method` |
| **R6 instance** | class-scoped namespace | **`r6_trait_<Trait>_<method>(x)`** | **`Type$Trait$method(x)`** |
| S4 (static/const) | flat, class-qualified (`$<-` interception) | `Type_Trait_method()` | `Type_Trait_method()` |
| S7 (static/const) | `.Type__Trait` env attached via `attr()` (`$<-` interception) | `attr(Type,"Trait")$method()` | `attr(Type,"Trait")$method()` |

S4 instance methods (`setMethod`) and S7 instance methods (S7 generic + fast-path shortcut) keep their own dispatch mechanisms; only their `-> Self` bodies gained re-wrapping. The only user-visible shape change is the R6 instance row.

## Test plan
- [x] `cargo test -p miniextendr-macros --lib`: 441 pass, incl. 28 trait tests (4 new: R6 instance uses class-qualified namespace; two r6 impls of one trait are distinct; `-> Self` re-wrapped per class system; static `-> Self` re-wrapped). No insta snapshot drift (inherent generators untouched).
- [x] All 3 CI clippy legs `-D warnings` clean: `clippy_default`, `clippy_all` (`full-codegen`), `clippy_all_s7` (`full-codegen-s7`).
- [x] `cargo fmt --all --check` clean (root + rpkg workspaces).
- [x] rpkg installs cleanly via the double regen loop (`rcmdinstall && force-document && rcmdinstall`) — proves two r6 impls of one trait no longer collide. Generated wrappers.R confirmed: R6 `Type$Trait$method` + `.ptr` extraction; Self-return re-wrap (`Class$new(.ptr=.val)` / `class(.val)<-`); S4 flat static names; S7 attr-env attach; BUG1 (`x_factor`) intact.
- [x] Runtime against the installed package: R6 rename, two-r6-collision dispatch, and `-> Self` re-wrap (r6 instance/static, env instance/static) all pass.
- [x] Cross-package trait ABI: `just cross-install && just cross-test` -> producer.pkg `PASS 52`, consumer.pkg `PASS 335`, 0 failures.
- [x] `just wrappers-sync-check` diff is exactly the intended NAMESPACE/man changes (no unrelated drift).

## Notes
- The full `just devtools-test` was killed (SIGKILL/OOM) in this sandbox from testthat's per-test callr fan-out (~48 R processes), not a test failure. The affected files were verified in-namespace and against the installed package instead; CI's r-tests job runs the full suite with adequate resources.
- S3/S4/S7 trait codegen is string-identical after the refactor (verified by the macro string-assertion tests and by inspecting the generated wrappers.R); the only behavioral changes are the R6 instance rename and `-> Self` re-wrapping.
- `just cross-install` regenerated a pre-existing, unrelated `@title` -> `@description` drift in the tracked cross-package wrappers.R; reverted here to keep scope and tracked separately in #1218.

---

</details>


CGMossa:
Rebased onto current main (clean, force-pushed) and re-ran the full loop end to end: this holds up. The namespace unification is applied to every generator, the `-> Self` re-wrapping matches the inherent-impl idioms one-for-one, and the R6 rename that #1115 needs falls out of the same change. Approving on the merits; one narrow edge case is worth a follow-up (filed #1246), and it collides with #1232 on regenerated files so whichever lands second has to regen.

## Summary

Verdict: correct and complete for both #1141 and #1115. Rebased `worktree-agent-ab2318fe3844ca4e4` onto `origin/main` (0fb04fb2), no conflicts, force-pushed.

Six-generator parity (the classic failure mode here) is fully covered, not 4-of-6:

| Class system | Namespace shape | `-> Self` re-wrap idiom | Inherent uses same idiom? |
|---|---|---|---|
| Env | `Type$Trait$m` | `class(.val) <- "Type"` | yes (`env_class.rs` `.build()`) |
| S3 | `Type$Trait$m` | `structure(.val, class = "Type")` | yes (`build_s3_body`) |
| Vctrs | `Type$Trait$m` (via S3 generator) | `structure(.val, class = "Type")` | yes (`vctrs_class.rs` `build_s3_body`) |
| S4 | `Type_Trait_m` (flat, `$<-` interception) | `methods::new("Type", ptr = .val)` | yes (`build_s4_body`) |
| S7 | `.Type__Trait$m` assigned, documented `Type$Trait$m` | `Type(.ptr = .val)` | yes (`build_s7_body`) |
| R6 | `Type$Trait$m` (was `r6_trait_<Trait>_<m>`) | `Type$new(.ptr = .val)` | yes (`build_r6_body`) |

There are five physical trait generators; Vctrs dispatches through the S3 one (`r_wrappers.rs:58`), and every path routes through the single `trait_namespace_target` / `TraitMethodContext::method_body_lines` owner. The only user-visible shape change is the R6 instance row.

Self-return edge cases checked: constructor-vs-method distinction (static `make() -> Self` re-wraps too); `Result<Self,E>` / `Option<Self>` handled via the same `ReturnSelf` tail because the C wrapper already raised on `Err`/`None`, so `.val` is a bare handle; no double-wrap (trait C wrappers hand back a bare `ExternalPtr`, and #1232's inherent-only wrapping does not touch trait generators); void `invisible(x)` preserved (void and `-> Self` are mutually exclusive, so the appended `invisible(x)` never races the re-wrap).

Nothing rewritten in the diff. The code is correct as authored; my only change is the rebase.

Verification (main repo, R 4.6.0):
- `cargo fmt --all --check` clean.
- `cargo test -p miniextendr-macros --lib`: 442 pass, including the 4 new tests; no insta drift.
- `cargo clippy -p miniextendr-macros -p miniextendr-api --all-targets --locked -- -D warnings` clean.
- `just rcmdinstall` + `just force-document`: zero tracked drift, so the committed NAMESPACE/man reproduce exactly. Regenerated wrappers.R confirmed to carry `R6TraitCounter$Counter$value <- function(x)`, `DoublerA$new(.ptr = .val)`, `class(.val) <- "DoublerEnv"`, `DoublerB$Doubler$doubled <- function(x)`, and zero `r6_trait_`.
- Runtime: `test-trait-r6-collision.R` passes (two r6 impls of one trait dispatch, r6 + env `-> Self` re-wrap). `test-trait-abi.R` / `test-class-system-matrix.R` / `test-trait-method-emitter.R` all green under `pkgload::load_all` (a first pass via bare `test_file` against the installed package showed S3-generic / S4-S7-constructor "could not find function" errors, which are a harness artifact of not using pkgload, not regressions).
- Cross-package trait ABI: `just cross-install && just cross-test` -> producer.pkg PASS 52, consumer.pkg PASS 335, 0 failures.

## Notes

Breaking shape: R6 instance trait methods move from `r6_trait_<Trait>_<method>(x)` to `Type$Trait$method(x)`; 8 `r6_trait_*` exports drop from NAMESPACE. Fine per no-backcompat, and it is the only R-visible rename.

Follow-up #1246 (not blocking): the R6/S7 `-> Self` re-wrap emits `Type$new(.ptr = .val)` / `Type(.ptr = .val)`, but whether the inherent constructor accepts `.ptr` is gated on `has_self_returning_methods`, computed from inherent methods only. A class with a trait `-> Self` method but no inherent `-> Self` (no constructor, no inherent factory) would get a constructor without `.ptr` and fail loudly. Narrow, non-silent, env/s3/s4 immune, and shared with #1232's cross-class wrapping which needs the same `.ptr` acceptance. The fixtures mask it because DoublerA/DoublerB both have `new(v) -> Self`.

#1232 interaction: both PRs touch `registry.rs` (different regions: this one edits the duplicate-wrapper panic in `detect_duplicate_wrapper_defs`, #1232 adds a `.__MX_WRAP_RETURN_` resolver in `write_r_wrappers_to_file`), `method_return_builder.rs` (#1232 only; this PR only consumes the existing API), and regenerated `NAMESPACE`/`man/`. No semantic conflict and no double-wrap (disjoint inherent vs trait generators). Whichever merges second must rebase and regenerate NAMESPACE/man.

#1228 contradiction: the docs/skill in #1228 states "R6 trait methods appear as flat exported functions; for R6, the shape is `r6_trait_<Trait>_<method>(obj)`" and "only Env gets dual calling." This PR makes both claims false for R6 (now class-scoped `Type$Trait$method`, same namespace shape as Env). Whichever merges second, #1228's skill text needs updating to the new R6 shape (do not edit #1228's branch from here).

Fleet CI: the Rust-tests leg is red fleet-wide on the trybuild-vs-rustc-1.97 mismatch and heals via #1239; unrelated to this PR (I ran `--lib`, not the UI leg).

</details>

Issue #509, opened 2026-05-11
feat: scope-as-allocator (protected-by-construction)


CGMossa:
# Plan: Scope-as-Allocator (Protected-by-Construction)

Builds on `unified-gc-protection.md`. That plan adds `Protector` trait + pool. This plan
changes the **API surface**: instead of allocate-then-protect, the scope IS the allocator.

## Problem

Current pattern has an unprotected gap:

```rust
let sexp = Rf_allocVector(VECSXP, 10);  // unprotected SEXP exists here
let root = scope.protect(sexp);          // now safe
```

Between lines 1 and 2, any R allocation (including `Rf_mkChar` for names) could trigger
GC and collect `sexp`. In practice R doesn't GC between two C-level ops without its
own allocation, but:
- The pattern is error-prone (easy to forget `protect`)
- It's impossible to enforce at compile time
- Every new contributor has to learn the rule

## Insight

R function arguments are auto-protected by `.Call()`. `VECTOR_ELT`, `STRING_ELT`,
`CAR`/`CDR` return references into already-protected parents. The **only** SEXPs that
need protection are ones we **allocate**.

So if the scope provides the allocation functions, `scope.protect(raw_sexp)` becomes
a rare escape hatch, not the primary API. Protection is guaranteed by construction.

## Proposed API

```rust
// Today:
unsafe {
    let scope = ProtectScope::new();
    let vec = scope.protect(Rf_allocVector(INTSXP, 10));
    let names = scope.protect(Rf_allocVector(STRSXP, 10));
    // fill vec and names...
    Rf_setAttrib(*vec, R_NamesSymbol, *names);
    vec.into_raw()
}

// Proposed:
unsafe {
    let scope = ProtectScope::new();
    let vec = scope.alloc_integer(10);
    let names = scope.alloc_character(10);
    // fill vec and names...
    Rf_setAttrib(*vec, R_NamesSymbol, *names);
    vec.into_raw()
}
```

### Core allocation methods on `ProtectScope`

```rust
impl ProtectScope {
    // Vectors
    pub unsafe fn alloc_integer(&self, n: usize) -> Root<'_>;
    pub unsafe fn alloc_real(&self, n: usize) -> Root<'_>;
    pub unsafe fn alloc_logical(&self, n: usize) -> Root<'_>;
    pub unsafe fn alloc_raw(&self, n: usize) -> Root<'_>;
    pub unsafe fn alloc_complex(&self, n: usize) -> Root<'_>;
[...]

PR #1220, opened 2026-07-08
fix(api): root Rust-built data.frames via owned BuiltDataFrame handle (#1128)

### LINKED ISSUE #1128: `DataFrame` is an unrooted `Copy` SEXP wrapper — every hold-across-allocation site is a latent use-after-free

<details>

## Summary
- `DataFrame` (miniextendr-api/src/dataframe.rs) is `#[derive(Clone, Copy)]` over a bare SEXP with no GC root. `SerdeRowBuilder::finish()` / `into_dataframe()` assemble the frame inside a `ProtectScope`, drop the scope, and hand back the unrooted wrapper. Any R allocation while Rust still holds the wrapper lets the GC reclaim the frame. Because a freed node's payload stays intact until slot reuse, most such sites *pass* tests silently (freed-but-intact reads).
- The gc_stress dynamic sweep (plan 06, audit A8) caught two manifestations on its first full pass: `dispatch_to_dataframes` held `ok_df` unrooted across `err_builder.finish()` (segfault under `gctorture(TRUE)`), and the derive reader blanket (`FromDataFrame for Vec<T>`) read Rust-built frames with no root (intermittent "DataFrame always carries a names attribute" panic).
- Both got targeted `OwnedProtect` guards (see `reviews/2026-07-03-gc-sweep-unrooted-dataframe.md`), mirroring the guard serde's `dataframe_to_vec` has carried since `reviews/2026-05-29-serde-deserialize-fixture-gctorture-input-protect.md`. That makes three point-wise fixes for the same bug class.
- Design question: should holding a Rust-built `DataFrame` across allocations be safe by construction?
  1. **Rooted RAII handle**: make `DataFrame` non-`Copy`, root on construction (`R_PreserveObject` or a protect-index slot), release on drop. Safe by construction; touches every `DataFrame` call site and the `Copy`-dependent APIs.
  2. **Split types**: keep the cheap `Copy` view for R-supplied frames (rooted by the `.Call` frame) and introduce an owned `BuiltDataFrame` RAII type as the return of `finish()` / `into_dataframe()`, `Deref`-ing to the view. Smaller blast radius; the unrooted view can still be smuggled across allocations.
  3. **Status quo + contract**: document "root before any allocation" on `DataFrame` and keep adding guards at framework composition points as the sweep finds them. Zero churn; users composing builder output with their own R calls stay exposed.

## Notes
- The sweep is the regression harness for whichever direction: any new hold-across-allocation site inside a `gc_stress_*` fixture's production path now fails PR CI.

---

</details>

CGMossa:
Closes #1128.

Fixes the `DataFrame` unrooted-`Copy` use-after-free at its source by splitting the type: `DataFrame` stays the cheap `Copy` **view** for R-supplied frames (rooted by the `.Call` frame / a `ProtectScope`), and a new owned, GC-rooted `BuiltDataFrame` becomes the return type of every Rust-side constructor.

> [!NOTE]
> **Design decision — needs sign-off.** This implements **option 2** from the issue (split types), not option 1 (make `DataFrame` always-rooted). Rationale: bounded blast radius, keeps the hot R-supplied-frame read path allocation- and root-free, and the compiler drives the call-site sweep. Flagging for review before merge.

<details>
<summary>🤖 Implementation & verification detail (AI-generated)</summary>

### What shipped
- **`BuiltDataFrame`** (`miniextendr-api/src/dataframe.rs`): owned RAII handle, `!Send`/`!Sync` (`PhantomData<*mut ()>`), roots via `R_PreserveObject` in `adopt_sexp`, releases via `R_ReleaseObject` in `Drop`. `Deref`s to `DataFrame` so all read/edit methods keep working unchanged. Hands off to R through `IntoR` using `release_into_sexp` + `mem::forget` (no double-release; no allocation between release and return).
- **Producers re-typed** to return `BuiltDataFrame`: `IntoDataFrame::into_dataframe`/`_par`, `SerdeRowBuilder::finish`, `DataFrame::builder().build()`, the serde `*_to_dataframe` verbs, `NamedList::as_data_frame`. Call sites that only read keep compiling via `Deref`; the few that stored/copied the old view surfaced as compile errors — each was exactly a previously-unrooted site.
- **`!Send`** makes returning a built frame from `with_r_thread`/`run_on_worker` unrepresentable (build + drop stay on the R thread).
- **Point-guards consolidated** where the owned handle supersedes them; serde `dataframe_to_vec`'s input guard kept (different hazard — roots an R-*supplied* frame during deserialization).
- **Docs**: `dataframe.rs` module rustdoc documents the view-vs-owned contract and the residual footgun (editing methods still return unrooted views); CLAUDE.md/MEMORY gotcha line updated.

### Regression harness (the point of the issue)
`gc_stress_built_dataframe()` (no-arg, `noexport`, `rpkg/src/rust/gc_stress_fixtures.rs`): builds a frame via `SerdeRowBuilder::finish`, forces an R-allocation burst **while holding the handle**, then reads every column back. Fails on pre-fix code under `gctorture(TRUE)` (frame reclaimed → panic/garbage); passes now.

### Verification
- **Rust:** `just test` green except the 7 pre-existing `derive_dataframe_*` UI-snapshot tests that skew on the local toolchain (CI stable is authoritative; unrelated to this change — no macro code touched).
- **Clippy:** all 4 CI legs clean `-D warnings` (default, `full-codegen`, `full-codegen-s7`, revendor; feature lists read from `.github/workflows/ci.yml`).
- **fmt:** clean.
- **gctorture:** `gc_stress_built_dataframe()` passes under `gctorture(TRUE)` (per `docs/GCTORTURE_TESTING.md`: load package first, then torture, call via `getNamespace`).

</details>


CGMossa:

<details>

## Summary
- **Verdict: approve.** This is a correct, well-scoped fix for the constructor-return leg of #1128, and the memory-safety-critical primitive is sound. Rebased cleanly onto current `origin/main` (disjoint file sets, no conflicts), re-verified, and pushed. Final stack: `85658795` (fix) + `400f88d8` (fmt) + `28f5bf4b` (my doc addition, below).
- **Root lifecycle audit (clean).** `BuiltDataFrame` does exactly one `R_PreserveObject` in `adopt_sexp` and exactly one `R_ReleaseObject` in `Drop`. It is neither `Copy` nor `Clone`, and `!Send`/`!Sync` via `PhantomData<*mut ()>`, so a double-release is unrepresentable. The R hand-off (`release_into_sexp`) does `R_ReleaseObject` then `mem::forget(self)`, so `Drop` cannot release a second time; the release-then-return window is allocation-free, matching the existing `DataFrame::into_sexp` / `List::build().into_sexp()` contract. On panic between build and return the `Drop` guard releases the root, no leak.
- **Constructor coverage map (all paths root now).** `IntoDataFrame::into_dataframe`/`_par` (blanket + `SerdeRows`), `SerdeRowBuilder::finish`, `RDataFrameBuilder::build`, `NamedList::as_data_frame`, `List::as_data_frame`, and the serde verbs `vec_/iter_/par_iter_/par_iter_growing_/map_/hashmap_/flatten_enums_to_dataframe` all return `BuiltDataFrame`. The empty-frame short-circuits go through `BuiltDataFrame::adopt_sexp(empty_dataframe())`; `assemble_with_scope` adopts before dropping the per-column `ProtectScope`, so the frame is never momentarily unrooted. Grepped for remaining bare-view producers, none left.
- **Internal serde composition is strictly safer than before.** In `dispatch_to_dataframes` and `build_split`, the owned handles stay rooted across the second builder's allocation (the old code needed explicit `OwnedProtect` point-guards there, and `build_split`'s `error_df` was actually reaped-able pre-fix). The `*handle` derefs feed the cheap view into `NamedDataFrameListBuilder`, which re-roots it into its own scope before any allocation; each handle lives to the end of its statement.
- **`!Send` enforcement is witnessed by the bench.** `benches/rayon.rs` had to move the handle's use inside `with_r_thread` because it can no longer cross back out; no worker path returns the handle, and no cross-package trait ABI signature carries `DataFrame`.
- **gctorture: PASS.** `just configure && just rcmdinstall` (R 4.6, exit 0), then a `gctorture(TRUE)` per-function loop: `gc_stress_built_dataframe` survived 60 iterations plus the four `create_*_df` producers at 20 each, all green. The no-arg `noexport` fixture is auto-enumerated by the dynamic sweep in `test-gc-stress-fixtures.R` (it enumerates the whole namespace, not just exports), so it gates PR CI without a hand-written test block. `cargo test -p miniextendr-api --features serde,rayon`: all suites pass (353 + others, 0 failed). `cargo clippy ... -D warnings` on api: clean. `cargo fmt --all --check`: clean.
- **What I added (`28f5bf4b`).** Per-method `# Rooting` caveats on the five frame-building editing methods (`drop`, `select`, `select_rows`, `prepend_column`, `with_column`-append), pointing at the `ProtectScope` / `BuiltDataFrame::adopt` escape hatch. The residual was documented at the type level but not on the methods a caller actually lands on in rustdoc.

## Notes
- **Residual is real but bounded and now tracked (#1247).** The frame-building editing methods still return unrooted views; `create_df().drop("x")` held across an allocation is a live (opt-in) footgun, since the source handle's root does not transfer to the freshly-built frame. Internal callers are safe (they `protect_raw` immediately). #1247 covers finishing the class (either return `BuiltDataFrame` from these methods, or fold into the protected-by-construction design in #509/#510) and notes the sibling `List`/`NamedList` unrooted-`Copy` wrappers (`from_raw_pairs` etc.), which are the same hazard.
- **Fleet-wide Rust-tests red is unrelated.** The trybuild UI snapshots skew against the 1.97 toolchain fleet-wide (per the review-chain context); not touched here (no macro code) and not caused by this PR.
- One observation, not blocking and not from this PR: `RUSTDOCFLAGS="-D warnings" cargo doc -p miniextendr-api` errors on 11 pre-existing `private_intra_doc_links` to `crate::altrep_data::iter`. My added links resolve cleanly; this is outside the rooting change and left alone to keep the PR focused.

---

</details>

Issue #1158, opened 2026-07-05
load_all() fails on shared S4 generic ordering — `setMethod("s4_get_value")` before `setGeneric` (valgrind 'R tests / Linux' job)


CGMossa:
## Symptom

The `R tests / Linux` CI job (valgrind memcheck over ALTREP materialization, which runs `library(miniextendr); testthat::test_local(filter = "altrep-materialization")`) fails during `pkgload::load_all()`:

```
Error in `load_all()`:
! Failed to load 'R/miniextendr-wrappers.R'
Caused by error in `methods::setMethod()`:
! no existing definition for function 's4_get_value'
```

Observed on PR #1044's run (job 85141690915), but #1044 does not touch any S4 codegen — see below.

## Why this is a latent `main` bug, not #1044's

- `s4_get_value` is a **shared** S4 generic: multiple `#[miniextendr(s4)]` classes each expose an inherent `get_value()` — `CounterTraitS4`, `S4Counter`, `S4TraitCounter` (see `rpkg/src/rust/class_system_matrix.rs` and `man/s4_get_value.Rd`, `man/S4Counter.Rd`, `man/S4TraitCounter.Rd`). The generic `setGeneric("s4_get_value", ...)` and the per-class `setMethod("s4_get_value", "<Class>", ...)` are emitted as **separate** entries in the generated `R/miniextendr-wrappers.R`.
- `installed` load works (namespace/lazy-load semantics), but `load_all()` `source()`s the file top-to-bottom. If any `setMethod("s4_get_value", ...)` entry is ordered **before** the `setGeneric` guard, `setMethod` errors with "no existing definition".
- `git diff origin/main...` for PR #1044 touches only `condition.rs` + condition demo/test files — nothing in `class_system_matrix.rs`, the S4 generators, `r_class_formatter.rs`, or `collect_r_wrappers()`. The `s4_get_value` emission is byte-identical to `main`.
- `main`'s own `R tests / Linux` job has been **cancelled** (superseded by rapid pushes — the job takes ~54 min) on every run since the last success at `61d4854c` (#1116). So the regression is masked on `main`; a PR whose run happens to *complete* the job (like #1044) is where it surfaces.

## Root-cause candidates

[...]

PR #1223, opened 2026-07-09
fix(macros): S7 local generics classify existing binding, shadow non-generics (#1114)

### LINKED ISSUE #1114: s7 class system: method registration fails at package load when the generic name resolves to a non-generic base function

<details>

## Summary
- The S7 class generator emits, per method name:
  ```r
  if (!exists("var", mode = "function")) {
    var <- S7::new_generic("var", "x", function(x, ...) S7::S7_dispatch())
  }
  S7::method(var, ClassName) <- function(...) ...
  ```
- When the name already resolves to a **non-generic** base/stats function (`var`, `get`, `col`, `row`, `diag`, `reshape`, ...), the `new_generic` branch is skipped and `S7::method(<plain closure>, ...) <-` fails at package load with `` `generic` is a function, but not an S3 generic function ``. The package then fails `R CMD INSTALL` (lazy-loading error).
- Primitives and S3 generics are fine (`sum`, `mean`, `min`, `max` all register). Only plain-closure collisions break.
- Found by the feature-legs runtime CI (audit A5): flipping `s7-default` on turned rpkg's bare ndarray fixtures (`fn var`, `fn get`, ...) into S7 classes that cannot load. The same failure hits an explicit `#[miniextendr(s7)]` impl with such a method name, so this is not specific to `s7-default`.
- Interim mitigation in rpkg: the colliding fixture impls (`NdVec`, `NdMatrix`, `NdArrayDyn`, `NdIntVec`, `ErrorInRCounter`) are pinned to `#[miniextendr(env)]`.

## Options
- Shadow with a package-local S7 generic when the existing function is not a generic (S7 supports package-local generics that mask base; needs care with NAMESPACE exports and user expectations).
- Or detect the collision in the generated R code and produce a clear load-time (or better, macro-time) diagnostic naming the method and suggesting a rename / `env`. A macro-time denylist of known base R non-generics would catch the common cases (`var`, `get`, `row`, `col`, `diag`, `reshape`).

## Notes
- Repro: any `#[miniextendr(s7)]` impl with `fn var(&self)`, then `R CMD INSTALL`.

---

</details>

CGMossa:
Makes `#[miniextendr(s7)]` impls install even when a method name collides with a plain (non-generic) base/stats closure like `var`, `get`, `row`, `col`. Previously the S7 codegen guarded `new_generic` with a bare `exists(name)` check, so a colliding name skipped generic creation and `S7::method(<plain closure>, ...) <-` errored at package load ("generic is a function, but not an S3 generic function").

Closes #1114.

<details>
<summary>🤖 Implementation & verification detail (AI-generated)</summary>

### The fix (`miniextendr-macros/src/miniextendr_impl/s7_class.rs`)
Replaced the bare existence check with a **usability classifier**. The existing binding is reused only when it is something `S7::method<-` actually accepts as a generic — an S7 generic, a primitive, an S3 standard generic (`utils::isS3stdGeneric`), or an S4 generic (`methods::isGeneric`). Otherwise the codegen defines a **package-local S7 generic that shadows** the plain closure and registers a `class_any` fallback delegating to the masked function, so ordinary calls keep working while S7 objects dispatch.

- Lookups are `base::exists` / `base::get`-qualified so a shadow generic named e.g. `get` can't re-route a later generic's classifier through itself.
- The S7 **trait-impl** path is unaffected — its generics are namespace-prefixed (`s7_trait_<Trait>_<method>`) and never collide with base names.
- **roxygen2 gotcha handled:** the classifier's `.mx_gen` binding lives inside the braced `else if` condition, not as a leading top-level `.mx_gen <- …` — the latter would make roxygen2 emit a bogus `\alias{.mx_gen}`/`\usage` on every S7 man page.

### Decisions (flagged)
- **Macro-time denylist: skipped (YAGNI).** The issue floated a compile-time warning for known base names; the runtime classifier makes install *work*, so a warning would just add noise for a now-working case. Revisit only if the masking UX confuses users.
- **Known behavioral divergence (documented, not a bug):** because masking re-roots the shadow generic's default environment, a bare `get("x")` for a *caller-frame local* differs from `base::get` — pass `envir=` for locals. All other `get` forms (search-path names, explicit `envir=`/`mode=`) match. Pinned in the testthat fixture.

### Fixtures / test coverage
- Reverted the interim `#[miniextendr(env)]` pins on `NdVec`, `NdMatrix`, `NdArrayDyn`, `NdIntVec`, `ErrorInRCounter` — they're live s7 fixtures again.
- Added an explicit `#[miniextendr(s7)]` `S7NonGenericCollision` fixture (`fn var`, `fn get`) so the collision path is covered on **default-feature** CI, not just the `s7-default` leg. `rpkg/tests/testthat/test-s7-nongeneric-collision.R` pins: install+construct succeed, S7 dispatch (`var(g)==12` ≠ `stats::var`), fast-path shortcuts, fallback (`var(1:10)==stats::var(1:10)`, `get(..., envir=)`), and loud failure on no-applicable-method.
- Notes the interaction with #1113 (make fixtures explicit) without doing that work here.

### Verification
- **Clippy:** all 4 CI legs clean `-D warnings` (default, `full-codegen`, `full-codegen-s7` — the compile gate for this change — revendor; feature lists from `.github/workflows/ci.yml`).
- **Snapshot:** `miniextendr_impl` s7 snapshot rebaselined; diff reviewed.
- **`just test`:** green (bar the 7 pre-existing `derive_dataframe_*` UI-snapshot tests that skew on the local toolchain — CI stable is authoritative; not touched here).
- **R:** targeted `test-s7-nongeneric-collision.R` against the installed package — all pass (install proves the collision path works; dispatch + fallback + loud-failure pins green).

</details>


CGMossa:

<details>

## Summary
- Verdict: sound, resolves #1114. Verified end to end on a fresh rebase onto origin/main (post-#1239): `just configure && rcmdinstall && force-document` installs rpkg with the new `S7NonGenericCollision` fixture (`fn var`, `fn get` as S7 generics), which by itself proves the load-time failure is gone; all 10 assertions in `test-s7-nongeneric-collision.R` pass under `pkgload::load_all`; the regen loop produces zero NAMESPACE / man / wrappers drift; 438 macros lib+snapshot tests pass; `clippy_default -D warnings` and `cargo fmt --all --check` clean.
- Classification matrix (reviewed, all correct): no binding creates a plain generic; a plain closure (`var`/`get`/`row`/`col`) shadows with a `class_any` fallback delegating to the masked function; a primitive, an S3 standard generic, an S4 generic, or an existing S7 generic all reuse as-is. The common base names land right: `print`/`format`/`mean` (S3 std generics) and `length`/`sum` (primitives) reuse rather than shadow. Both misjudgment directions are safe: treating a plain closure as usable fails loudly at load (the original #1114 error), and treating a usable generic as non-usable just adds a redundant but working shadow. No silent mis-dispatch path.
- Load-order: the classifier runs once at lazy-load build time; the resulting generic object and its eagerly-captured `.mx_masked` are serialized, so runtime dispatch is stable regardless of what attaches later. One build-time dependency: resolving `var` to `stats::var` relies on the default packages being attached during install (they are), which is why rpkg does not need to `import(stats)`.
- Added one commit (9f5671a7): `minirextendr::use_s7()` now also imports `methods` + `utils`. The emitted classifier calls `utils::isS3stdGeneric()` / `methods::isGeneric()`, so a scaffolded S7 package would otherwise draw an undeclared-`::`-import NOTE from `R CMD check`. rpkg already declared both (the PR added `utils` to its DESCRIPTION); only the scaffold path was missing.
- Sibling sweep: the S3/vctrs generator (`emit_s3_generic_guard`, `r_class_formatter.rs:57`) still uses the identical bare `exists()` guard, so an `#[miniextendr(s3)]` method named `var`/`get` silently fails to dispatch (worse than S7 since there is no load error). Filed #1248 with a fix sketch. S4 is immune (namespace-local `exists(where=topenv, inherits=FALSE)` + `setGeneric` masking, per #1158). The S7 trait path is safe (`s7_trait_<Trait>_<method>` names never collide with base). Env/R6 are immune (methods bind into the object, no namespace generic).

## Notes
- The trybuild `ui` suite shows 5 mismatches locally, all `derive_dataframe_enum_*` tests untouched by this PR. The only diff is the stdlib panic-span rendering (`$crate::panicking::panic_fmt(...)` vs the bare `= note: the failure occurred here`), i.e. the rust-src-component skew (present on the dev toolchain, absent on CI's minimal profile). Pre-existing and local-only; green on CI. Not blessed.
- `get()` has one documented divergence pinned in the fixture: a bare `get("x")` for a caller-frame local resolves against the fallback closure's environment, not the caller's, so callers pass `envir=` for locals. All search-path and explicit-`envir=`/`mode=` forms match `base::get`.
- #1219 interplay: #1219 (approved, unmerged) rewrites trait-method namespace shapes across all six generators including S7, but it touches the trait path (`s7_trait_` generics), not the inherent-impl classifier changed here. No semantic conflict; whichever merges second rebases and regenerates artifacts.

---

</details>

Issue #1213, opened 2026-07-08
design: marker types to control R wrapper generation (Invisible<T>, WrapAs<T>, and friends)


CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h3>AI-written details</h3></summary>

## Summary

Brainstormed during the 2026-07-08 golife dogfooding session, while investigating why `&mut self -> ()` R6 methods unconditionally emit `invisible(self)` (see `journal/2026-07-08-golife-dogfood-hiccups.md`'s design suggestion, and `plans/2026-07-08-golife-hiccup9-cross-class-return-wrapping.md` for cross-class return wrapping). Both of those problems share a root shape: **the R-side wrapper's return-handling behavior is currently inferred from the Rust return type's syntactic shape** (`()` -> invisible-chainable; `Self`/`Result<Self,E>`/etc. -> constructor-wrap), with no way for a method to state its intent explicitly.

Revised after discussion (see thread): this is no longer just a menu of options, it's a policy plus two concrete mechanisms.

## Implementation-mechanism caveat (read this first)

"Marker trait" is the natural name to reach for, but proc-macros expand **before** trait resolution -- `#[miniextendr]` cannot ask "does this type implement `Chainable`?" at macro-expansion time, because that answer doesn't exist yet. Two mechanisms actually work today, and the framework already uses both elsewhere:

1. **Syntactic marker newtypes**, matched by literal type-path name in the AST -- the same mechanism `ParsedMethod::returns_self()` (`miniextendr_impl.rs:2052-2132`) already uses to special-case a literal `Self` return. This is the only mechanism that can change the **generated R source text** and it needs no lookup beyond the syntax already in front of the macro.
2. **Real marker traits**, resolved by Rust's own trait system *after* macro expansion -- viable only for changing what a blanket `IntoR`/`TryFromSexp` impl *does* in Rust, not what R-side wrapper code gets emitted. `AsSerialize<T>` (`miniextendr-api/src/serde/traits.rs`) is this pattern today. Doesn't apply here.

So the type-based half of every design below is a **newtype wrapper type**, matched syntactically -- "marker trait" in spirit, not in Rust's technical sense.

## Cross-cutting policy: every mechanism ships as a type marker *and* an attribute, doing the identical codegen

[...]

Issue #860, opened 2026-06-04
miniextendr_build() installs before document() → first build exports nothing (NAMESPACE lags one build)


CGMossa:
## Summary

`miniextendr_build()` installs the package (`devtools::install()`, Step 3)
**before** regenerating the NAMESPACE (`devtools::document()`, Step 4). The R
wrapper files (`R/<pkg>-wrappers.R`) are generated *during* the install/`make`
step, and `document()` reads their roxygen `@export` tags to write NAMESPACE.
So on a **first build** — or any build that adds/renames an exported function
— the install collates the *old* NAMESPACE (missing the new `export()`
directives), and the installed image exports nothing new. `document()` then
fixes the on-disk NAMESPACE, but the already-installed package is stale.

Net effect: a fresh `miniextendr_build()` must be run **twice** before
`library(pkg)` exposes the exported wrappers.

This is **distinct** from the #757/#775 tarball-mode wrapper-gen-skip bug: there
the wrapper *file* was missing; here the wrapper file **and** the namespace
objects are generated correctly — only the `export()`s lag by one build.

## Reproduction

Standalone package scaffolded with `create_miniextendr_package()`, with an
`#[miniextendr]` fn whose name differs from the scaffold's `add`/`hello`
(triggers a new `@export`). One `miniextendr_build(path = ".")`:

```r
suppressMessages(library(pkg))
getNamespaceExports("pkg")           # character(0)  (or stale add/hello)
ls(getNamespace("pkg"))              # the new fns ARE here, just not exported
pkg::new_fn()                        # Error: could not find function "new_fn"
```

NAMESPACE on disk is *correct* at this point (document() wrote it). A second
`miniextendr_build(path = ".")` installs against that NAMESPACE → exports
appear, functions resolve.

Hit while wiring a `diffsol`-based engine crate into an R package; confirmed on
`minirextendr` @ current `main`, R 4.6, macOS arm64.

## Root cause

`minirextendr/R/workflow.R`, `miniextendr_build()`:

- Step 3 (~line 109): `devtools::install()`
- Step 4 (~line 144): `devtools::document()`

Chicken-and-egg: `document()` needs the wrapper files (generated during
install) to exist, but `install()` needs the documented NAMESPACE to export.
Running them install→document means the export set is always one build behind.

## Proposed fix

Split wrapper-generation out of the install step so the order becomes
**wrapper-gen → document → install**:

1. Build the cdylib and write `R/<pkg>-wrappers.R` (the existing
[...]

PR #1225, opened 2026-07-09
docs: capture golife dogfood session — hiccups journal + skill touch-ups

CGMossa:
Captures the 2026-07-08 golife (Conway's Game of Life) dogfooding session: writing a real consumer package end-to-end via `minirextendr::use_miniextendr()` to surface framework friction.

Docs-only. No code changes.

- `journal/2026-07-08-golife-dogfood-hiccups.md` — the narrative: builder-pattern idiom, the two-pass build gotcha, and the 9 friction points ("hiccups") found. Hiccup #6 (a suspected `miniextendr_build` two-pass bug) was investigated and **retracted as a false alarm** in a follow-up commit.
- Skill touch-ups prompted by the session: `miniextendr-getting-started` (builder idiom + desired call syntax), `miniextendr-class-systems`, and an `miniextendr-altrep` example fix (the prelude doesn't cover per-family derives).
- Links the `Invisible<T>` return-wrapping design idea to the filed issue **#1213**.

The other 8 hiccups are captured as individual plan files (`plans/2026-07-08-golife-hiccup*.md`) and will land as separate PRs. The golife package itself stays out of this repo — it's a standalone consumer package.

### DIFF
--- a/.claude/skills/miniextendr-altrep/SKILL.md
+++ b/.claude/skills/miniextendr-altrep/SKILL.md
@@ -122,6 +122,8 @@ Annotate your struct:
 
 ```rust
 use miniextendr_api::prelude::*;
+use miniextendr_api::altrep_data::{AltIntegerData, AltrepLen};
+use miniextendr_api::AltrepInteger;
 
 #[derive(AltrepInteger)]
 #[altrep(len = "length", elt = "value", class = "ConstantInt")]
@@ -130,6 +132,13 @@ pub struct ConstantIntVector {
     pub value: i32,
 }
 ```
+
+The prelude only re-exports the generic `Altrep` derive and `IntoRAltrep` —
+not the seven per-family derives (`AltrepInteger`/`Real`/`Logical`/`Raw`/
+`String`/`Complex`/`List`) or the `AltrepLen`/`Alt*Data` traits, which live
+at the crate root and in `altrep_data` respectively. Always add the explicit
+imports above alongside the prelude glob; a bare `use
+miniextendr_api::prelude::*;` is not enough for ALTREP.
 
 The derive emits:
 - `impl AltrepLen for ConstantIntVector` reading `self.length`.
--- a/.claude/skills/miniextendr-altrep/SKILL.md
+++ b/.claude/skills/miniextendr-altrep/SKILL.md
@@ -122,8 +122,6 @@ Annotate your struct:
 
 ```rust
 use miniextendr_api::prelude::*;
-use miniextendr_api::altrep_data::{AltIntegerData, AltrepLen};
-use miniextendr_api::AltrepInteger;
 
 #[derive(AltrepInteger)]
 #[altrep(len = "length", elt = "value", class = "ConstantInt")]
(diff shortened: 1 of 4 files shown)


CGMossa:

## Summary

Verdict: approve after the fix pushed below. Rebased onto current main (`0a58ce5a`, 12 commits ahead including the merged #1219) with no conflicts. Head is now `bbe30340`.

**Journal half (historical session record): accurate, left intact.** I did not rewrite the narrative. I added one `**Tracked:**` pointer line per hiccup mapping each to the PR that now fixes it (#1233 h1, #1235 h2, #1230 h3, #1234 h4, #1227 h5, #1226 h8, #1232 h9), plus a note on the retracted hiccup #6.

**Skill half (living docs): one real defect fixed and pushed.** The getting-started iteration loop's step 3 told users to run `miniextendr_build()` a *second time* to make a newly added export callable. That is wrong: `miniextendr_build()` already collapses the reinstall into a single call (it snapshots `NAMESPACE` before/after `document()` and reinstalls once if the export set changed, `minirextendr/R/workflow.R`, issue #860 closed 2026-06-09). Running it twice is a no-op fixpoint. It also contradicted this PR's own hiccup #6 retraction, which established the single-call reinstall is correct. I split step 3 so the single-call behavior is stated for `miniextendr_build()` and the manual second install is scoped to the monorepo raw recipes only (`just rcmdinstall && just force-document && just rcmdinstall`, per CLAUDE.md).

**Post-#1219 staleness check on the skill touch-ups: clean.** The builder-idiom section this PR adds to the class-systems skill is correct on main. #1219's Self-return re-wrapping applies to owned `-> Self` trait returns, not the `&mut self -> &mut Self` borrow the builder idiom uses. The `pipe_builder` fixture (`rpkg/tests/testthat/test-pipe-builder.R`) asserts R6 chains preserve object identity and passes on main, so "codegen must return the receiver directly, not re-wrap it" still holds. The altrep import note is also correct on main (the per-family derives and data traits are not in the prelude yet). No stale `r6_trait_*` naming appears in any skill file.

**Freshness audit:** `bash scripts/skill-freshness-audit.sh` exits 0 (0 BLOCKING). The 2 WARN on getting-started are the documented benign false positives (generated `.cargo/config.toml`, scaffolded `src/rust/src/lib.rs` layout path).

## Notes

- **Overlap with #1228 (r6-trait-dispatch docfix):** this PR does NOT touch the "Trait methods across class systems" section (class-systems SKILL lines 287-346), which is #1228's territory. I left it untouched to avoid duplicating or conflicting with #1228. Heads-up for the #1228 reviewer: that section teaches universal dual calling, and #1219 has now moved R6 instance trait methods into the `Type$Trait$method(x)` namespace (the Env/S3 shape). So #1228's title framing ("R6/S3/S4/S7 emit flat functions, scope dual calling to Env") is now superseded for R6 by the merged #1219 and needs reconciling. Flagged in the journal hiccup #7 cross-reference as well.
- **Overlap with #1230 (prelude ALTREP exports):** this PR's altrep skill note says the prelude does not cover the per-family derives, so add explicit imports. #1230 re-exports them from the prelude. Correct on main today; once #1230 merges, this note becomes obsolete and whoever merges #1230 should trim it. Flagged in the journal hiccup #3 cross-reference.
- Nothing in this PR's own hunks was left stale on purpose.
- SHAs: base `0a58ce5a`, head `bbe30340`.

</details>

PR #1205, opened 2026-07-08
fix(rpkg): Makevars pkg-name, configure.ac dnl backticks, rustdoc Rd links (#1123 #1076 #1188)

CGMossa:
Three small config/docs fixes, batched into one PR.

- **`#1123`** — `rpkg/src/Makevars.in` wrapper-gen echo messages interpolated the undefined make variable `$(PACKAGE_NAME)` (printing `R/-wrappers.R`). Replaced with the autoconf-substituted `@PACKAGE_NAME@` token that the `WRAPPERS_R` variable in the same file already uses. Ported identically to both `minirextendr/inst/templates/{rpkg,monorepo/rpkg}/Makevars.in`. Closes #1123
- **`#1076`** — `checkbashisms` (no m4 awareness) misreads backtick-quoted prose in `dnl` comments as command substitution, emitting spurious `R CMD check --as-cran` NOTEs. Reworded every `dnl` comment line in `rpkg/configure.ac` to use single quotes instead of backtick pairs; `dnl` lines are stripped by autoconf, so no `configure` regeneration and no behavior change. Ported to both template `configure.ac` copies so scaffolded packages get the fix too. Closes #1076
- **`#1188`** — rustdoc intra-doc links in `///` doc comments leaked into `man/*.Rd` as invalid URLs: `[RValue](miniextendr_api::RValue)` became `\href{miniextendr_api::RValue}{...}` (conversions.Rd) and the angle-bracketed `Vec<civil::Date>` was parsed as a commonmark autolink (the `civil:` scheme) into `\url{civil::Date}` (jiff_adapter_tests.Rd). Reworded to plain backticked / `\code` / `\verb` forms in `conversions.rs`, `gc_stress_fixtures.rs`, and `jiff_adapter_tests.rs`, and regenerated `NAMESPACE` + `man/`. Closes #1188

### Verification
- `#1123`: `grep` confirms the three echo messages now use `@PACKAGE_NAME@`; template copies match.
- `#1076`: `grep` confirms zero `dnl` lines contain a backtick pair (rpkg + both templates); real shell-code backticks left untouched. Full `R CMD check --as-cran` not run locally (heavy) — CI will confirm the NOTE is gone.
[...]

PR #1226, opened 2026-07-09
fix(scaffolding): gitignore ignores src/rust/.cargo/config.toml at its real path

CGMossa:
## Summary
- Fixes golife hiccup #8 from `journal/2026-07-08-golife-dogfood-hiccups.md` on `docs/golife-dogfood-2026-07-08`.
- Changes the package and scaffold gitignore pattern from `.cargo/config.toml` to `src/rust/.cargo/config.toml`.
- Adds `rpkg/gitignore` and `monorepo/rpkg/gitignore` to `templates-sources` and re-locks `patches/templates.patch`.

## Notes
- The original hiccup also mentioned `R/*-wrappers.R` and `src/rust/wasm_registry.rs`, but the plan identified those as stale/already fixed. This PR leaves them untouched.
- No deferred scope was cut.

## Verification
- `rig default 4.6 && R --version | head -1` -> `R version 4.6.0 (2026-04-24) -- "Because it was There"`.
- `just worktree-sync` passed.
- `just configure` passed.
- Baseline `just templates-check` passed before edits.
- `just templates-approve` regenerated `patches/templates.patch`.
- Post-approve `just templates-check` passed.
- In `rpkg/`, `git check-ignore -v src/rust/.cargo/config.toml` exited 0.
- Fresh standalone scaffold from this worktree: after `bash ./configure`, `git check-ignore -v src/rust/.cargo/config.toml` exited 0 and matched `.gitignore:5:src/rust/.cargo/config.toml`.
- Rust fmt/clippy/test/install gates were skipped per the task-specific instructions because there is no Rust delta.


CGMossa:

<details>

## Summary
- Verdict: solid, correct fix. Rebased onto current `main` (10 merges since branch point) and hardened with a regression test plus an upgrade-path cleanup.
- Path semantics verified: the old pattern `.cargo/config.toml` contains a mid-string slash, so git anchors it to the `.gitignore`'s own directory and it only ever matched `<pkg>/.cargo/config.toml`, never the real nested `src/rust/.cargo/config.toml`. Confirmed in a throwaway repo with `git check-ignore -v`: old pattern exits 1 (not ignored, the bug), new pattern exits 0 and matches `.gitignore:5`. The monorepo exemplar was masked only by the repo-root backstop `rpkg/src/rust/.cargo/`.
- Template sync verified: the fix lands in all three files (`rpkg/.gitignore` master plus both `minirextendr/inst/templates/{rpkg,monorepo/rpkg}/gitignore`), and the two gitignores were correctly added to the `templates-sources` manifest so this class of drift is now caught. All other generated-file patterns in the gitignore (`src/Makevars`, `R/*-wrappers.R`, `src/rust/wasm_registry.rs`, `vendor/`, target dirs) are already correctly anchored, so `config.toml` was the sole broken line.
- Rebase hazard handled: `main` #1205 rewrote the same `patches/templates.patch` (configure.ac dnl backticks to single quotes). The branch's committed patch conflicted and was stale (backticks). I regenerated it with `just templates-approve` so it preserves all of #1205's content and adds only the two gitignore hunks (verified: the sole semantic delta from `origin/main`'s patch is those two hunks). `just templates-check` passes.
- Retroactivity: a `.gitignore` only affects untracked files, so packages scaffolded before this fix that already committed `config.toml` (the golife dogfood package) get no benefit. I added the superseded bare `.cargo/config.toml` to `upgrade_gitignore()`'s obsolete list so `upgrade_miniextendr_package()` prunes the dead line instead of leaving it beside the corrected one. The full remediation (a `minirextendr_doctor()` warning when the generated file is git-tracked, advising `git rm --cached`) is filed as #1250.
- Added a behavioral `git check-ignore` regression test in `test-ignore-scaffolding.R`. The existing tests only assert the template text, so the broken pattern passed them while never matching the real path. The new test drives the paths configure and the host cdylib pass actually generate.
- SHAs: fix commit `c493b311` (rebased and patch-corrected), review-hardening `4ca2200a`. minirextendr suite: `[ FAIL 0 | WARN 0 | SKIP 3 | PASS 701 ]`.

## Notes
- The bare `.cargo/config.toml` pattern never matched anything in a scaffolded package, so nothing legitimately relied on it and treating it as obsolete on upgrade is safe.
- The `test-templates.R` "templates patch is in sync with rpkg sources" test is the CI-side guard for the regenerated patch and it passes against the corrected file.
- Retroactivity follow-up: #1250.
- CI was still in progress at the time of writing; the Generated Files Check and Version Check had already gone green.

---

</details>

PR #1227, opened 2026-07-09
fix(api): let TypedList::get/get_opt retrieve vector fields, not just scalars

CGMossa:
Context:
- Fixes golife dogfood hiccup #5 from `docs/golife-dogfood-2026-07-08:journal/2026-07-08-golife-dogfood-hiccups.md`.
- `TypedList::get` / `get_opt` now accept any `TryFromSexp` target whose error implements `Display`, so vector targets like `Vec<f64>`, `Vec<i32>`, and `Vec<String>` compile instead of being limited to scalar `SexpError` conversions.
- Adds the `typed_list_get_vectors` rpkg fixture, testthat coverage, docs/skill text, and regenerated `rpkg/NAMESPACE` / `rpkg/man/dots_tests.Rd`.

Verification:
- `rig default 4.6 && R --version | head -1` -> `R version 4.6.0 (2026-04-24) -- "Because it was There"`
- `just fmt`
- `cargo clippy --workspace --all-targets --locked --features full-codegen,rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,jiff,blake3,md5,globset,zstd -- -D warnings`
- `cargo clippy --workspace --all-targets --locked --features full-codegen-s7,rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,jiff,blake3,md5,globset,zstd -- -D warnings`
- `just check`
- `just test` ran; the only nonzero result was the known local rustc vs CI-stable trybuild skew: 7 `derive_dataframe_*` UI mismatches. Snapshots were not overwritten.
- `just rcmdinstall && just force-document && just rcmdinstall`
- `R -e 'library(miniextendr); testthat::test_file("rpkg/tests/testthat/test-dots-more.R")'` -> `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 31 ]`

No scope was cut or deferred.


CGMossa:
Rebased onto current main and re-reviewed: this is the correct fix, and I pushed one sibling extension (`DataFrame::column` carried the identical scalar-only bound). Verified green end to end. Details below.

## Summary

Verdict: approve. Root cause and fix are both right, and the retrieval surface is now consistent.

Root cause is a trait-bound over-constraint, not a runtime check. `get`/`get_opt` pinned `T: TryFromSexp<Error = SexpError>`. Scalar impls use `SexpError`; every vector/collection impl uses the narrower `SexpTypeError` (`from_r/collections.rs:185`), so `.get::<Vec<_>>()` never satisfied the bound and failed to *compile*. The PR widens to `T: TryFromSexp` with `T::Error: Display`, which is exactly what the body needs (it only ever does `format!("{}", e)`). This is a genuine widening, not a parallel code path: `From<SexpTypeError> for SexpError` already exists (`from_r.rs:303`), confirming the two error types were always a precision split rather than an intended incompatibility. Both `get` and `get_opt` were widened, so it is not a half fix; `get_raw` returns a raw SEXP and needs no bound.

NA and coercion parity holds by construction. Retrieval routes through the same `T::try_from_sexp` impls used for function-argument conversion, so `get::<Vec<i32>>()` behaves identically to a `Vec<i32>` argument: the native slice fast path exposes NA as the sentinel, and NA-safe retrieval via `get::<Vec<Option<i32>>>()` is now reachable too (its `SexpTypeError` also implements `Display`). No new i32::MIN hazard beyond the framework's existing native-vector contract. `WrongType` errors still name the field, the spec-derived expected type, and the actual conversion error; scalar messages are unchanged.

Retrieval-surface consistency check: `List::get_index`/`get_named` and `NamedList::get`/`get_index` already accepted any `T: TryFromSexp` (they discard the error). The one sibling still carrying the scalar-only bound was `DataFrame::column`/`column_index`, which redeclared `T: TryFromSexp<Error = SexpError>` even though its delegate `NamedList::get` does not need it, and a column is inherently a vector. I widened both (commit f411717a) so `df.column::<Vec<f64>>("x")` compiles; it is a pure widening with no existing callers of the read method.

Verification: `cargo fmt --all` clean; `clippy_default` and `clippy_all` at `-D warnings` green; `cargo test -p miniextendr-api` all pass, macros non-UI green. The 5 `derive_dataframe_enum_*` UI mismatches are the known local rust-src vs CI-minimal stdlib-span skew (panic wording byte-identical, only the `$RUST/core/src/panic.rs` span note differs); not overwritten. R loop reinstalled: `test-dots.R` PASS 4, `test-dots-more.R` PASS 31 (includes the new `typed_list_get_vectors` fixture), `test-typed-dataframe.R` PASS 25.

## Notes

`TryFromList` for `HashMap`/`BTreeMap`/`Vec` (`list.rs`) still constrains its element type to `Error = SexpError`, but that is a list-to-collection conversion, not a `::<T>()` retrieval surface a user calls, so I left it untouched. No items deferred; no issues filed.

</details>

PR #1228, opened 2026-07-09
docs(skill): scope trait "dual calling" to Env; R6/S3/S4/S7 emit flat functions

CGMossa:


CGMossa:

<details>

## Summary

When codex opened this PR on the morning of 2026-07-09, the skill genuinely overclaimed: trait "dual calling" was described as universal, and the fix scoped it to Env while stating that R6/S3/S4/S7 emit flat free functions, with R6 exported as `r6_trait_<Trait>_<method>(obj)`. Later that day #1219 (merged, closes #1141 and #1115) rewrote the trait-method emitter: the per-class R-visible name is now owned by `trait_namespace_target` in `miniextendr-macros/src/miniextendr_impl_trait/method_context.rs`, and the unqualified `r6_trait_*` name is gone. That made codex's replacement text false on main (the R6 shape it named no longer exists, and "flat free functions" was never accurate for S3/S4/S7 either).

This rewrite (verified against the current generators in `miniextendr_impl_trait/r_wrappers.rs` and `method_context.rs`, plus the real call sites in `rpkg/tests/testthat/test-trait-abi.R`, `test-class-system-matrix.R`, and `test-trait-r6-collision.R`) teaches the post-#1219 reality. Ground truth, one line per system:

- Env: dual-form. Callable both as `Type$Trait$method(obj)` and as `obj$Trait$method()` (the `.__mx_instance__` attribute drives `$.Type` dispatch to prepend `self`). The only system with dual calling.
- R6: single-form `Type$Trait$method(obj)` in a class-scoped trait namespace on the generator object. Not dual (a monolithic `R6Class()` cannot host `obj$method()`). The old `r6_trait_*` name is gone and asserted absent by `test-trait-r6-collision.R`.
- S3 / Vctrs: idiomatic S3 dispatch, an S3 generic plus `method.Type`, called `method(obj)`. Static methods and consts live in `Type$Trait$name`.
- S4: an S4 generic `s4_trait_<Trait>_<method>(obj)` via `setGeneric`/`setMethod`. Static methods and consts are flat, class-qualified `Type_Trait_method()` functions (S4 intercepts `$<-`).
- S7: an S7 generic `s7_trait_<Trait>_<method>(obj)` plus an optional per-method fast-path `Type_method(obj)`. Static methods and consts are reachable as `Type$Trait$name` through an `attr()`-attached env.
- Self-returning methods (`-> Self` / `Result<Self, E>` / `Option<Self>`) now re-wrap into a classed object on every system.

## Notes

- What of codex's fix survived: the central correction is correct and preserved. Dual calling really is Env-only, and the Env `function(x = self, ...)` re-parenting mechanics are unchanged. What changed is the characterization of the other five systems (they use their own dispatch conventions, not "flat functions"), and the now-dead R6 `r6_trait_*` naming.
- #1115 and #1141 are both CLOSED by the merged #1219, so the skill no longer frames the R6 gap as an open tracked issue.
- Rebased onto current `origin/main` and rewritten into a single clean commit; force-pushed with lease.
- Swept the other skills and `docs/` for the same stale claims (`r6_trait_`, "dual calling", "flat functions"); none carry the doc-bug, so no other files needed changes. The only additional edit is the file-map bullet in this same skill, which now points at `method_context.rs` as the namespace owner.
- `bash scripts/skill-freshness-audit.sh` is clean for this skill (0 BLOCKING; the remaining WARNs are pre-existing false positives on generated/gitignored and external paths).

---

</details>

PR #1232, opened 2026-07-09
fix(macros): auto-wrap instance-method returns that name another registered class

### LINKED ISSUE #1231: golife hiccup #9: auto-wrap cross-class ExternalPtr method returns
From golife dogfood hiccup #9 in `journal/2026-07-08-golife-dogfood-hiccups.md` on branch `docs/golife-dogfood-2026-07-08`.

A class-system instance method can return a different `#[miniextendr]` ExternalPtr-backed class, e.g. `PatternBuilder::build(&self, ...) -> Board`. The C wrapper returns the target pointer, but the generated R wrapper currently falls through to a direct `.val` tail, so users receive a bare `externalptr` instead of `Board$new(.ptr = .val)` / the target system equivalent.

Expected behavior: when the return type names another registered class, generated wrappers should construct the returned class object using the target class system constructor. Unregistered capitalized return types should remain direct values.

Related context: #1141 §4 discusses factoring pointer-wrapping helpers for trait methods returning `Self`; this hiccup is the inherent-method cross-class case and does not overlap directly, though a shared helper could be reused later.

CGMossa:
## Summary

- Fixes golife dogfood hiccup #9 from `journal/2026-07-08-golife-dogfood-hiccups.md` on branch `docs/golife-dogfood-2026-07-08`.
- Adds a write-time `.__MX_WRAP_RETURN_<RustType>__(.val)` resolver that uses `MX_CLASS_NAMES` to wrap returned ExternalPtrs with the returned class system constructor.
- Adds macro detection for bare capitalized non-container return types and threads the return type through Env/R6/S3/S4/S7/vctrs method wrapper generation.
- Broadens R6/S7 `.ptr` constructors so returned R6/S7 targets can always be constructed from an existing pointer.
- Adds same-crate R6 and S7 cross-class builder fixtures plus targeted testthat coverage.

Closes #1231.
Related: #1141 §4 is adjacent helper/refactoring context but does not cover this inherent cross-class return case.

## Verification

- `rig default 4.6 && R --version | head -1` -> `R version 4.6.0 (2026-04-24)`
- `just fmt`
- `cargo clippy --workspace --all-targets --locked --features full-codegen,rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,jiff,blake3,md5,globset,zstd -- -D warnings`
- `cargo clippy --workspace --all-targets --locked --features full-codegen-s7,rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,jiff,blake3,md5,globset,zstd -- -D warnings`
- `just check` passed
- `just test` reached the known local trybuild stderr skew: exactly 7 `derive_dataframe_*` UI mismatches; snapshots were not overwritten
- `just configure`
- `just rcmdinstall`
- `just force-document`
- `just rcmdinstall`
- `R -e 'library(miniextendr); testthat::test_file("rpkg/tests/testthat/test-pipe-builder.R")'` -> `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 39 ]`


CGMossa:
Rebased onto current main (which now carries #1219) and re-verified end to end. This is correct and, pleasingly, resolves the #1246 follow-up as a side effect. I added a mixed class-system fixture (R6->S7 and S7->R6) to prove the resolver keys off the returned class and not the receiver's system; both directions pass at runtime. Approving on the merits, with one honest scope gap filed as #1251.

## Summary

Verdict: correct and complete for the stated scope (#1231, golife hiccup #9). Rebased `fix/macros-cross-class-return-wrapping` from base `e3826c55` onto `origin/main`; the rebase was clean with no conflicts. The team-lead brief anticipated heavy conflicts in the macros crate, but #1219 only touched `registry.rs` (13 lines, in `detect_duplicate_wrapper_defs`) among the overlapping files. Its trait rewrites live in `miniextendr_impl_trait/`, whereas this PR edits `miniextendr_impl/` plus the resolver in `write_r_wrappers_to_file`, so the two are disjoint. Author's rebased commit `6a5a8b7b`; my hardening commit `7563d60f`.

**Mechanism.** A syntactic macro-time heuristic (`ParsedMethod::returns_other_class`) tags any method whose return is a bare capitalized `Type::Path` (not `Self`, not a listed primitive, not a listed container) with `ReturnStrategy::ReturnOtherClass`. All six generators then emit a `.__MX_WRAP_RETURN_<RustName>__(.val)` placeholder. At write time, `write_r_wrappers_to_file` looks `<RustName>` up in `MX_CLASS_NAMES` (populated by link time) and rewrites to the target's own constructor: r6 `X$new(.ptr=)`, s7 `X(.ptr=)`, s4 `methods::new("X", ptr=)`, s3/env/vctrs `structure(., class="X")`. A registry miss is a quiet no-op (emits the bare value), which is why the deliberately loose macro-time heuristic is safe.

**Type-matching semantics, mapped honestly.**
- Fully-qualified paths work: matching uses the last path segment, so `crate::Cell` becomes marker `Cell` and resolves.
- Registered class returned by value is always correct: the derive's `IntoR` is the only `IntoR` for that type, so the pointer handed back is that class's handle, and wrapping into its constructor cannot be wrong.
- False positives are harmless: a capitalized `IntoR` type that is not a registered class (`Robj`, a newtype, a jiff/time type) misses the registry and falls back to the bare value, which is today's behavior.
- False negatives (real, filed as #1251): `Option<Class>`, `Result<Class, E>`, `Vec<Class>` are excluded by the container list, so they return bare handles. This is an asymmetry with #1219, whose `-> Self` path does re-wrap `Option<Self>` / `Result<Self, E>`.
- Type aliases (`type Board = Cell; -> Board`) miss the registry (alias name is not the registered rust_type) and fall back to bare. Non-silent to the extent the user notices an unclassed handle; documented behavior of the miss path.

**Cross-class-system matrix.** Resolution is keyed entirely on the returned class's `MX_CLASS_NAMES` entry and never consults the receiver's system, so all combinations are handled by construction. The original fixtures only covered R6->R6 and S7->S7. I added the two mixed directions to `pipe_builder_tests.rs` and asserted them at runtime in `test-pipe-builder.R`: `R6CrossPlan::build_s7 -> S7CrossBoard` generates `S7CrossBoard(.ptr = .val)`, and `S7CrossPlan::s7_build_r6 -> R6CrossBoard` generates `R6CrossBoard$new(.ptr = .val)` in both the generic method and the fast-path shortcut. The five per-system resolver unit tests in `registry.rs` cover r6/s7/s4/s3/env independently.

**#1246 findings.** This PR resolves #1246 rather than amplifying it. It drops the `has_self_returning_methods` gate on the `.ptr` parameter for both R6 and S7 inherent constructors, so every ExternalPtr-backed R6/S7 class now unconditionally accepts `.ptr` (a minimal `.ptr`-only constructor is synthesized where no `new()` exists). That is exactly the fix suggested in #1246, and it also removes the "constructor without `.ptr`" failure mode for #1219's trait `-> Self` re-wrap. The gap is regression-covered at the macro level by the two new tests `r6_class_without_constructor_still_has_ptr_initialize` and `s7_class_without_constructor_still_has_ptr_constructor`. Commented on #1246 recommending it close on merge. Env/S4/S3/vctrs targets need no constructor change since they wrap by attribute or slot, and the resolver's `structure()` / `methods::new()` fallbacks match #1219's proven inherent-Self idioms one-for-one.

**Double-wrap.** No risk: `for_method` selects `ReturnSelf` and `ChainableMutation` before `ReturnOtherClass`, and the C wrapper for a cross-class return hands back a bare (unclassed) handle, so exactly one wrap is applied. `-> impl IntoR` is `Type::ImplTrait`, not a `Type::Path`, so it never matches.

**Verification (main repo, R 4.6.0, everything green).**
- `cargo fmt --all --check` clean; rpkg crate `cargo fmt --check` clean.
- `cargo test -p miniextendr-macros --lib`: 453 pass (incl. the 4 new detection/marker/hatch tests). `cargo test -p miniextendr-api --lib`: 303 pass (incl. the 5 resolver tests).
- All three CI clippy legs pass `-D warnings`: `clippy_default`, `clippy_all` (full-codegen), `clippy_all_s7` (full-codegen-s7).
- `just configure && just rcmdinstall && just force-document && just rcmdinstall`: generated `wrappers.R` has zero unresolved markers; committed NAMESPACE/man reproduce with no unintended drift; no roxygen warnings.
- Runtime via `pkgload::load_all`: `test-pipe-builder.R` PASS with 0 fail/warn/skip (same-system plus both mixed cases). Regression sweep of the class-system and trait suites (`test-class-system-matrix`, `test-class-systems`, `test-trait-r6-collision`, `test-trait-abi`, `test-match-arg-impl`, `test-r6-noexport-field`, `test-s4-helpers`): 285 pass, 0 fail, 0 warn. The `test-trait-r6-collision` pass confirms the unconditional `.ptr` broadening does not disturb #1219's `-> Self` re-wrapping.

## Notes

- Scope gap filed as #1251: container-wrapped cross-class returns (`Option<Class>` / `Result<Class, E>` / `Vec<Class>`) are not re-wrapped. The scalar `Option`/`Result` forms are a small extension of `returns_other_class` since the C wrapper already unwraps them; `Vec` needs an `lapply` over the constructor and can stay separate.
- No per-method opt-out (`no_wrap`) exists for a user who wants the raw handle back. Consistent with the no-back-compat philosophy (returning a registered class type almost always means you want the classed object), but flagging it as an API-design point rather than a defect. Not filed.
- Cross-package trait ABI was not run: this PR touches only R-side wrappers and the write-time resolver, not the trait ABI or C wrapper signatures, and the resolver is a proven no-op on inputs without markers (zero leftover markers on rpkg's ~32K-line wrappers.R).
- Two pre-existing pipe-builder tests were requalified to `miniextendr:::` because `S4PipeBuilder` / `S7PipeBuilder` are `#[miniextendr(.., internal)]`. Harmless: it makes `test_file` work against the installed package and is a no-op under `test_check`.

</details>

PR #1233, opened 2026-07-09
fix(scaffold): surface the .git-ancestor trigger when configure auto-vendors into tarball mode

CGMossa:
## Summary
- Adds an auto-vendor notice explaining that no .git ancestor caused configure to assume a CRAN-style/offline build context.
- Ports the notice to the standalone scaffold template and updates the approved template patch.
- Reframes miniextendr_doctor no-git tarball guidance as an accidental-latch warning, and updates the getting-started skill to recommend git init before configure.
- Strengthens the standalone scaffold smoke test to assert the no-git notice is present.

References golife hiccup #1 in journal/2026-07-08-golife-dogfood-hiccups.md on docs/golife-dogfood-2026-07-08.

## Verification
- R version check: R version 4.6.0.
- rpkg autoconf regenerated configure cleanly.
- rpkg bash ./configure in the git worktree stayed in source mode and did not print the no-git notice.
- Temporary no-git rpkg copy ran bash ./configure, printed the no-git auto-vendor notice, and completed in tarball mode.
- just templates-check failed before approval as expected, then just templates-approve and just templates-check passed.
- just fmt passed.
- just minirextendr-test 2>&1 > /tmp/minirextendr-test.log passed: [ FAIL 0 | WARN 0 | SKIP 3 | PASS 697 ].

Skipped Rust clippy legs and rpkg double-install per task instructions because this change does not modify Rust or add a new miniextendr export.

No deferred scope.

### DIFF
--- a/.claude/skills/miniextendr-getting-started/SKILL.md
+++ b/.claude/skills/miniextendr-getting-started/SKILL.md
@@ -87,6 +87,10 @@ Run this from your new package's root directory (which should already have a
 `DESCRIPTION` file — create it first with `usethis::create_package("mypkg")`
 if needed).
 
+Before running `./configure`, initialize git with `usethis::use_git()` or
+`git init`. Without a `.git` ancestor, configure assumes a CRAN-style/offline
+build context, auto-vendors, and latches into tarball mode.
+
 `use_miniextendr()` writes the following into your package:
 
 - `src/Makevars.in` — build configuration template. `configure` writes
--- a/minirextendr/R/doctor.R
+++ b/minirextendr/R/doctor.R
@@ -252,9 +252,11 @@ tarball-mode cleanup in Makevars. Fix: \\
   #     behind it makes the post-build Makevars cleanup delete
   #     src/rust/.cargo/ from the source tree on the next
   #     `miniextendr_build()`. Fail loudly.
-  #   - no `.git` ancestor: most likely intentional CRAN-prep staging
+  #   - no `.git` ancestor: may be intentional CRAN-prep staging
   #     (bootstrap.R runs ./configure in a build-staging dir with no .git) or
-  #     a legitimate offline install. Warn instead of failing.
+  #     a legitimate offline install, but can also mean configure was run
+  #     before `git init` and accidentally latched into tarball mode. Warn
+  #     instead of failing.
   cli::cli_h2("Vendor tarball")
 
   is_dev_source_tree <- !is.null(find_root_with_file(".git", usethis::proj_get()))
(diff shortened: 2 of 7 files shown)


CGMossa:

<details>

## Summary
- Verdict: LGTM. Rebased onto current origin/main (was `3c222dd0`, now `bf985b0e`) and force-pushed. The messaging fix is correct and behaviorally verified.
- Message accuracy: the new `AC_MSG_NOTICE` in `rpkg/configure.ac:156` names the real trigger (no `.git` ancestor) and the consequence (assuming a CRAN-style/offline build context). It prints only when the auto-vendor branch actually fires, which is exactly the silent path from golife hiccup #1.
- Both-paths coverage: the fix targets the auto-vendor self-repair branch (the genuinely silent one). The pre-existing-tarball path at `configure.ac:183` already prints an explicit `install mode = tarball install (offline, vendored)` label, and its confusing git-tracked subcase is hard-errored by the leaked-tarball guard at `configure.ac:131`, so no additional message is warranted there.
- autoconf / template sync: `rpkg/configure` is in sync with `configure.ac` (re-running `autoconf` produced zero drift). The rebase conflicted on the generated `patches/templates.patch`; I regenerated it via `just templates-approve` (never hand-merged), amended it into the commit, and `just templates-check` is green on the committed bytes.
- Behavioral test: ran `bash ./configure` on a `.git`-free copy of `rpkg` with a stub `cargo-revendor` on PATH. The notice printed to both stdout and `config.log` (line 81), which validates the smoke test's `config.log` assertion mechanism. Re-ran after `git init` in the same copy: notice absent, stayed in source mode. The em-dash rendered correctly in both.
- `doctor.R`, the getting-started skill, and the smoke test were reviewed and parse-clean; the doctor warning correctly reframes the no-`.git` tarball case as a likely accidental latch while preserving the legitimate-CRAN clause and the `git init` + `miniextendr_clean_vendor_leak()` fix.

## Notes
- The em-dash in the notice is consistent with existing convention: `configure.ac` and the generated `configure` already carry em-dashes in many `AC_MSG_NOTICE` / `echo` messages that pass CI, so this is not a portability or checkbashisms regression. The notice writes to stdout/`config.log`, not stderr.
- The strengthened smoke-test assertion lives in a `skip_on_ci()` test (it needs cargo + autoconf), so CI will not exercise it; it runs under local `just minirextendr-test`. That is a pre-existing property of the test, not introduced here.
- Sibling candidates considered but not extended (neither is a defect): rpkg's silent hardcoded feature-fallback when `tools/detect-features.R` is absent or errors (the scaffold template variant already prints a notice for that case), and the tarball-present install-mode path, which already labels itself.
- CI was still in progress at push time; Generated Files Check and Version Check were already green.

---

</details>

PR #1234, opened 2026-07-09
fix(macros): rewrite ... variadic to &Dots in impl-block methods (golife hiccup #4)

CGMossa:
## Summary
- Rewrites raw `...` variadics in class-system impl methods into trailing `&::miniextendr_api::dots::Dots`, including the re-emitted impl block rustc sees.
- Auto-detects trailing `&Dots` in R wrapper argument generation so impl constructors/methods emit `...` and forward `list(...)`.
- Suppresses duplicate dispatch `...` for S3/S4/S7/vctrs method formals when user dots already exist.
- Adds R6 and S3 impl-block dots fixtures covering constructors, instance methods, explicit `&Dots`, and a no-arg GC stress export.

References golife hiccup #4 from `docs/golife-dogfood-2026-07-08:journal/2026-07-08-golife-dogfood-hiccups.md`. This unblocks the separate hiccup #2 plan for impl-method `dots = typed_list!(...)` attribute sugar; that sugar is intentionally not included here.

## Verification
- `rig default 4.6 && R --version | head -1` -> `R version 4.6.0 (2026-04-24)`.
- `just fmt` passed.
- `cargo clippy --workspace --all-targets --locked -- -D warnings` passed.
- `cargo clippy --workspace --all-targets --locked --features full-codegen,rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,jiff,blake3,md5,globset,zstd -- -D warnings` passed.
- `cargo clippy --workspace --all-targets --locked --features full-codegen-s7,rayon,rand,rand_distr,either,ndarray,nalgebra,serde,serde_json,num-bigint,rust_decimal,ordered-float,uuid,regex,indexmap,time,num-traits,bytes,num-complex,url,sha2,bitflags,bitvec,aho-corasick,toml,tabled,raw_conversions,vctrs,tinyvec,borsh,jiff,blake3,md5,globset,zstd -- -D warnings` passed.
- `just check` passed.
- `just configure && just rcmdinstall && just force-document && just rcmdinstall` passed; regenerated `NAMESPACE` and `man/*.Rd` committed.
- `R -e 'library(miniextendr); testthat::test_file("rpkg/tests/testthat/test-impl-dots.R")'` passed: `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 14 ]`.
- `just test` was run and failed only on the documented local-rustc trybuild skew: 7 `derive_dataframe_*` UI mismatches. Snapshots were not overwritten; CI stable is authoritative per repo instructions.

No scope was cut or deferred.


CGMossa:
Rebased onto current `main` (picks up #1219's trait-namespace work with no conflict) and re-verified end to end: the shared-helper approach is the right call, all six class systems emit dots correctly, and I added unit coverage for the four systems the PR only proved by shared-code reasoning. Verdict below is a green light, no blocking changes.

## Summary

**Verdict: approve.** The fix is sound and well-scoped. Rebased onto `main` at `0a58ce5a`; original commit is now `96222f7e`; I pushed one follow-up test commit `9efa1155`.

**Parity / sharing:** the impl path reuses the free-fn path rather than copying it. `rewrite_variadic_dots` and `trailing_dots_ident` were extracted into `miniextendr_fn.rs` and are called by both `MiniextendrFunctionParsed::parse` (free fns) and `ParsedMethod::from_impl_item` (impl methods). `RArgumentBuilder::new` now auto-detects a trailing `&Dots`, which makes `build_r_formals_from_sig` / `build_r_call_args_from_sig` emit `...` / `list(...)` for both paths idempotently. No drift risk from duplicated logic.

**Coverage map (self-kind x class system).** All emit a single `...` and forward `list(...)`:

| self-kind | R6 | S3 | S4 | S7 | Env | Vctrs |
|---|---|---|---|---|---|---|
| constructor (`new -> Self`) | ok | ok | ok | ok | ok | ok |
| instance method (`&self`/`&mut self`) | ok | ok | ok | ok | ok | n/a (vctrs is static-only, MXL120) |
| static method (no self) | ok | ok | ok | ok | ok | ok |
| explicit `&Dots` param | ok | ok | ok | ok | ok | ok |

R6 and S3 were covered by the PR's unit tests plus `test-impl-dots.R` (14 pass) at runtime. I added `s4/s7/env/vctrs_user_dots_*` unit tests (commit `9efa1155`) asserting constructor + method emit one `...` with no `function(x, ..., ...)` duplicate and forward `list(...)`. Full macro suite is 457 pass.

**Dispatch-dots suppression:** `instance_formals_with_dots` and `instance_formals_with_receiver` gate the appended dispatch `...` on `!self.method.has_dots`, so a user dots param is the only `...`. The dots-only method case is safe because `build_formals` never yields empty params when `has_dots` (it emits `...`). The vctrs protocol branch also picks up `!has_dots` and, as a bonus, fixes a latent leading-comma bug when params were empty.

**Position + `.call`:** dots are always the trailing input, so `...` lands last in both formals and call args, matching Rust order. `list(...)` is a normal positional arg after `.call = match.call()` (e.g. `.Call(C_R6DotsThing__collect, .call = match.call(), private$.ptr, list(...))`), no collision. Sidecar accessors (`externalptr_derive.rs`) are untouched, so the "sidecars must not get `.call`" invariant holds.

**Verification:** fmt clean; 457 macro tests pass; clippy default + `clippy_all` + `clippy_all_s7` all pass `-D warnings`; R loop (configure / rcmdinstall / force-document) clean with no NAMESPACE or man drift on the rebased tree; `test-impl-dots.R` 14 pass, `test-dots.R` 4, `test-dots-more.R` 30; `gc_stress_impl_dots_methods()` returns 42 under `gctorture(TRUE)`. UI tests: only the 5 `derive_dataframe_*` rust-src stdlib-span mismatches, the documented local-vs-CI toolchain skew, none dots-related and not overwritten.

## Notes

- **For #1235's reviewer (stacked on this):** #1235's branch `feat/macros-dots-sugar-impl-methods` was based on the pre-rebase head `90965413`, which no longer exists here. After this merges, #1235 must be rebased onto the new #1234 head (`9efa1155`). My only added change is unit tests in `miniextendr_impl/tests.rs`, so #1235's diff is otherwise unaffected.
- **Trait-impl methods are out of scope** (this PR handles inherent `impl` blocks across the six class systems). A `...` variadic in a `#[miniextendr_trait]` impl method still fails loudly at rustc (`... is not supported for non-extern functions`), not silently. Trait methods with an explicit `&Dots` would now pick up `...` / `list(...)` from the shared `build_r_formals_from_sig` auto-detection, but the trait dispatch wrapper shapes (env `function(x = self, ...)`, S7 generic) were not audited for duplicate-`...` and there is no existing trait method using `&Dots`, so no regression. If trait-method dots is wanted later it needs its own verification.
- **Runtime coverage of S4/S7/Env/Vctrs dots** is codegen-level (unit tests) rather than installed-package testthat, because those systems are feature-gated (`full-codegen-s7`, `vctrs`) and the PR ships no S4/S7/Env/Vctrs dots fixtures. The plan only required R6 + one dispatch system, which is met. Non-dots regression across all six is ruled out by the passing `s4_wrapper_full_snapshot` / `s7_wrapper_full_snapshot` / env / vctrs snapshot tests.

</details>

PR #1235, opened 2026-07-09
feat(macros): accept dots = typed_list!(...) on impl-block methods and constructors

CGMossa:
Wires the `#[miniextendr(dots = typed_list!(...))]` attribute sugar — previously supported only on standalone `#[miniextendr] pub fn` functions — into class-system `impl`-block methods and constructors. Before this, the sugar was rejected at compile time on any `#[miniextendr(r6/s3/s4/s7/env/vctrs)] impl` method with "unknown attribute", forcing a manual `dots.typed(typed_list!(...))` call in the body.

Closes golife dogfood **hiccup #2** (`journal/2026-07-08-golife-dogfood-hiccups.md`).

> **Stacked on #1234** (golife hiccup #4, raw `...`/`&Dots` support in impl methods). Until #1234 merges, this PR's diff includes #1234's commits — **merge #1234 first**, then this diff reduces to just the sugar change.

## The change (`miniextendr-macros`)
- `MethodAttrs` gains `dots_spec: Option<TokenStream>`, mirroring `MiniextendrFnAttrs.dots_spec`.
- `parse_method_attrs` gains a top-level `dots = typed_list!(...)` branch (with the same `typed_list!`-only validation and error wording as the fn path); the "unknown attribute" option list is extended to mention it.
- `from_impl_item` injects the `dots_typed` binding at the top of the method body via the shared `build_dots_validation_stmt` helper (made `pub(crate)`), using the dots-param ident hiccup #4 already resolves (`named_dots`). Injection is on the single `original_impl` node, so **all six class systems are covered by construction** — no per-generator duplication. A `dots = typed_list!(...)` on a method with no dots parameter is a clear compile error.

## Fixture / tests
- Extends `rpkg/src/rust/impl_dots_tests.rs` with `ImplDotsSugar` (R6): a constructor and a method both using `#[miniextendr(dots = typed_list!(... => numeric()))]` and reading validated fields via `dots_typed.get(...)`.
- Adds a no-arg `gc_stress_impl_dots_sugar()` fixture driving the sugar path.
- `rpkg/tests/testthat/test-impl-dots.R`: asserts the sugar validates on both constructor and method, errors on a missing required dot, and the expected formals shape.

## Verification (R 4.6.0, this branch)
- `just fmt` (no changes)
- clippy `full-codegen` leg (`-D warnings`) — feature list from `.github/workflows/ci.yml`
- clippy `full-codegen-s7` leg (`-D warnings`) — same list
- `just rcmdinstall && just force-document && just rcmdinstall` (double install for the new exports); regenerated `NAMESPACE` + `man/*.Rd` committed in sync
- `testthat::test_file("rpkg/tests/testthat/test-impl-dots.R")` → `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 20 ]`
- `just test`: only the known 7 `derive_dataframe_*` UI trybuild skews (local-rustc vs CI-stable `.stderr`); snapshots not overwritten, CI stable is authoritative.

No scope cut or deferred.


CGMossa:
Restacked onto the rebased #1234 (clean, no conflicts) and re-reviewed the sugar delta on top. This looks good to me: the injection-on-`original_impl` design really does cover all six class systems from a single site, and the parse/validation parity with the free-function path holds. I added three macro unit tests for the sugar (the PR only had the R6 end-to-end fixture) and, while doing so, confirmed the "requires a dots parameter" error is both correct and complete. Merge #1234 first, then this one.

<details>

## Summary
- **Verdict: approve, merge after #1234.** No behavior change requested. My only pushed change is additive test coverage.
- **Restack.** Old base `90965413` (pre-rebase #1234) no longer exists; the branch's own commit `2a9d3545` replayed cleanly onto the rebased #1234 head `9efa1155` via `git rebase --onto` (force-pushed as `bc19439c`). The sugar delta on top of #1234 is byte-identical to before the rebase, so no manual conflict resolution was needed. My test commit `dd2cb16b` sits on top; current head is `dd2cb16b`.
- **Attr plumbing across all six generators.** Verified the claim holds by construction: `ParsedMethod::from_impl_item` receives each method as a `&mut` into `original_impl.items`, and `original_impl` is the single node re-emitted as the Rust body (`miniextendr_impl.rs:2302-2306`, emitted at `:3428`). The six class generators emit only R/C wrapper glue, never the Rust method body, so the `dots_typed` injection reaches S4/S7/Env/Vctrs exactly as it reaches R6/S3. `MethodAttrs` derives `Default` and is not destructured exhaustively anywhere, so the new `dots_spec` field needs no generator updates (the "update all six generators" rule applies to `MiniextendrFnAttrs`/`ImplAttrs`/`ParsedImpl`, not `MethodAttrs`).
- **Validation parity.** The method-path reuses the shared `build_dots_validation_stmt` helper verbatim, so the `dots_typed` binding name, the `unwrap_or_else(panic!)` wording, and the audit-A8 `Display` (not `Debug`) formatting all match the free-fn path. The `typed_list!`-only check and its error wording ("dots expects `typed_list!(...)` macro") are identical.
- **Interaction with #1234.** A method with a `...`-variadic (rewritten to `&Dots` by #1234) plus the sugar is coherent: both resolve the same `named_dots` ident, which is exactly the fixture's shape. The sugar on an explicit `&Dots` parameter also resolves via `trailing_dots_ident`/`is_dots_type` (which recurses through the reference). Both paths are now covered by unit tests.
- **Tests added** (`miniextendr_impl/tests.rs`): sugar injection over a named `...`, sugar injection over an explicit `&Dots`, and the no-dots-parameter compile error. All pass against the PR's original macro code.

## Notes
- **Merge order: #1234 then #1235.** This PR is stacked; the GitHub diff (base `main`) shows #1234's commits too.
- **The "no dots parameter" error is correct and complete.** I initially thought a bare unnamed `...` plus sugar was a parity gap (the free-fn path falls back to the synthesized `__miniextendr_dots` when `named_dots` is None, whereas the method-path errors). It is not a real gap: current rustc denies a bare unnamed `...` at parse time (`varargs_without_pattern`, future-incompat deny, rust-lang/rust#145544), before the macro can rewrite it, so `named_dots` is None only for genuinely-no-dots methods. The free-fn path's fallback is vestigial dead code for real source; not worth touching here.
- **Minor, non-blocking parse divergence.** `dots = "string"` (a non-macro value) yields a generic syn parse error on the method-path versus the friendly "dots expects `typed_list!(...)`" message on the free-fn path. `dots = other!(...)` still gets the friendly message on both. Left as-is.

---

</details>

PR #1236, opened 2026-07-09
docs(agents): AGENTS.md in every CLAUDE.md dir + plan-file→PR workflow + gpt-5.5

CGMossa:
Gives Codex/Cursor the same per-directory guidance Claude Code already gets, and documents the plan-file→PR workflow used to clear the golife-hiccup backlog this session.

## What
- **AGENTS.md in every directory that has a CLAUDE.md** (12 of them). Each new file is a one-line `@CLAUDE.md` import plus a comment. That's Codex's own include mechanism — the root `AGENTS.md` already uses `@` (for `~/.codex/RTK.md`). So each directory's `CLAUDE.md` stays the **single source of truth**, and `AGENTS.md` inherits it with zero duplication and zero drift. (Symlinks were rejected — this repo targets Windows.)
- **Root `AGENTS.md`: new section "Working a plan file end-to-end (worktree → PR)"** — the reusable protocol (R 4.6 → read the plan → isolated worktree → implement → fmt+clippy+test → commit early → PR, don't merge) that the golife-hiccup codex wave ran on.
- **Bump the stale codex model** `gpt-5.3-codex` → `gpt-5.5` in `AGENTS.md` and `CLAUDE.md` (current frontier per the codex model list).

## Directories touched
`cargo-revendor`, `miniextendr-api`, `miniextendr-bench`, `miniextendr-engine`, `miniextendr-lint`, `miniextendr-macros`, `minirextendr`, `rpkg`, `rpkg/src/rust/satellite`, `rust-llm-docs`, `site`, `tests/cross-package`.

## Note / follow-up
If Codex resolves a relative `@CLAUDE.md` against the working directory rather than the file's own directory, the include would need an adjustment — but the human-readable comment still routes any reader to the sibling `CLAUDE.md`, so nothing is lost either way. Verifiable once codex credits are back.


CGMossa:
Re-reviewed after rebasing onto current main. The per-directory `@CLAUDE.md` import approach is the right call: it makes the subdir files drift-proof by construction, so the only genuine mirror left is the root pair, which the rebase reconciled with #1220 automatically. I pushed two small doc-consistency fixes on top and filed #1253 for an optional CI guard. LGTM to merge.

<details>

## Summary
- Verdict: ship-ready. Docs-only (14 `.md` files, no code touched). Rebased cleanly onto `origin/main` `0a58ce5a`; new head `db7ca4df` (original `f9508ad4` plus one hardening commit).
- Mirror completeness: 13 `CLAUDE.md` files, 13 `AGENTS.md` files, exact 1:1. No dir with a `CLAUDE.md` lacks an `AGENTS.md`; no orphan `AGENTS.md`. The two dirs beyond the CLAUDE.md orientation list (`rpkg/src/rust/satellite`, `rust-llm-docs`) each have a real `CLAUDE.md` sibling, so they are correct inclusions.
- Fidelity: the per-directory `AGENTS.md` files are one-line `@CLAUDE.md` imports, not content copies, so they inherit their sibling verbatim and cannot photocopy stale text. The only hand-kept mirror is the root `AGENTS.md`; #1220 (BuiltDataFrame) had already added the matching DataFrame line to both root files, and the rebase merged it into this branch automatically (different region from this PR's edits).
- `gpt-5.5` bump: consistent everywhere. The only codex model references in the repo are the four lines in the two root files (all `gpt-5.5`); no stale `gpt-5.3` remains, and there are no codex references in `.claude/skills` or scripts.
- New plan-file section: facts verified against the current tree (`just worktree-sync` at `justfile:673`, R pin `4.6` in `rproject.toml`, `clippy_default`/`clippy_all`/`clippy_all_s7` step names in `.github/workflows/ci.yml`).
- CI: Version Check and Generated Files Check pass; compile and test legs skip via the path filter, as expected for a docs-only change.

Hardening pushed in `db7ca4df`:
- The root `CLAUDE.md` orientation paragraph described only the root `AGENTS.md`. Updated it to state the invariant this PR creates (every `CLAUDE.md` dir has a sibling `AGENTS.md`, subdir ones being `@CLAUDE.md` imports, the root one a hand-kept mirror) plus the maintenance rule (a new `CLAUDE.md` gets a matching import beside it).
- The new "Working a plan file end-to-end" section sat 14 lines below the "forward-looking work lives in GitHub issues, not plan files" rule in `AGENTS.md`. Added a clause clarifying that a plan file is the execution spec for work being done now, while deferred or forward-looking work still goes to an issue.

## Notes
- Drift guard: I did not add a CI check to the PR itself. The import design already makes per-directory content drift impossible, and the update rule now lives in the root `CLAUDE.md` plus a comment in each `AGENTS.md`, so the residual risk is only a missing or orphaned file. Filed #1253 to track an optional one-line completeness grep (sibling to `templates-check` / `skill-freshness-audit`) if you want mechanical enforcement later; low priority.
- The new plan-file workflow section lives only in `AGENTS.md`, not `CLAUDE.md`. This reads as intentional: it consolidates for codex what `CLAUDE.md` already carries distributed across the Agent-worktrees section and memory. Flagging in case you would rather it also appear in `CLAUDE.md`.
- Pre-existing, out of scope: root `AGENTS.md` line 1 is `@/Users/elea/.codex/RTK.md`, a machine-specific absolute path that will not resolve for other contributors or CI. Untouched by this PR; noting only.
- The PR's own open question (does codex resolve a relative `@CLAUDE.md` against the file's own directory?) is still unverified pending codex credits. If it resolves against the working directory instead, the subdir imports are inert, but the human-readable comment still routes a reader to the sibling `CLAUDE.md`, so nothing is lost.

---

</details>

PR #1237, opened 2026-07-09
perf(minirextendr): jsonlite fast path for parse_cargo_metadata_json (#1064)

### LINKED ISSUE #1064: Simplify parse_cargo_metadata_json with a jsonlite-if-available fast path
Surfaced while fixing #1063 (the `detect-features` marker check). Not a bug — tracking a deferred simplification.

## What

`minirextendr/R/feature-detect-configure.R` hand-rolls a JSON parser for `cargo metadata` output: `parse_cargo_metadata_json()` uses regex for the `features` object and a char-by-char bracket-counting loop for the `dependencies` array (~65 lines, 4 dedicated tests). It feeds `list_cargo_features()` and its print method.

## Why it's there (and why it's defensible today)

This runs on the **maintainer's** machine, but minirextendr deliberately imports only `cli`/`fs`/`usethis` — no `jsonlite` — so the no-dep hand-roll avoids a new dependency. It's correct for current cargo output and well-tested. (Note: the `"features":\{[^}]*\}` regex is fine — feature specs can't contain `}`.)

This is **distinct from** `parse_cargo_features()` (the configure-time `[features]` line-scanner), which *must* stay base-R-only because it runs at install time with no package library. That one is necessary and should not change.

## Proposed

Use jsonlite when available, fall back to the current parser otherwise:

```r
parse_cargo_metadata_json <- function(json) {
  if (requireNamespace("jsonlite", quietly = TRUE)) {
    return(parse_cargo_metadata_jsonlite(json))  # ~10 lines: fromJSON, pluck features/deps
  }
  parse_cargo_metadata_regex(json)               # keep current impl as fallback
}
```

`jsonlite` is a transitive dep of usethis, so it's present in virtually every dev environment — the fast path runs almost always, and the hand-rolled fallback survives for the rare bare environment. Keep the existing tests pointed at the regex fallback and add a couple against the jsonlite path.

## Also consider

`list_cargo_features()` + `print.miniextendr_cargo_features` are diagnostic-only and depend on this parser. Their main output ("features without detection rules") is low-value given that rule-less features are enabled by default. Worth a look at whether they earn their keep — fold into this refactor or close as won't-do.

Low priority.

CGMossa:
Adds a `jsonlite` fast path to `parse_cargo_metadata_json` in minirextendr's maintainer-side feature detection, falling back to the existing hand-rolled regex parser when jsonlite isn't available.

Closes #1064.

- The dispatcher tries `jsonlite` via `requireNamespace("jsonlite", quietly = TRUE)` → `parse_cargo_metadata_jsonlite()` (~10 lines); otherwise the renamed `parse_cargo_metadata_regex()` (the original impl, behavior unchanged).
- **Maintainer-side only** — the configure-time base-R `parse_cargo_features` line-scanner is deliberately untouched.
- Tests: existing coverage kept on the regex path (called directly); jsonlite-path tests guarded by `skip_if_not_installed("jsonlite")`.
- The optional `list_cargo_features()` / `print.miniextendr_cargo_features` removal floated in the plan was left out of scope (no behavior change here); can be a follow-up if wanted.

## Verification (R 4.6)
Targeted `testthat::test_file("minirextendr/tests/testthat/test-feature-detect-configure.R")` → **FAIL 0 | SKIP 0 | PASS 85**, with `jsonlite` present so the fast path is actually exercised (not skipped).

### DIFF
--- a/minirextendr/DESCRIPTION
+++ b/minirextendr/DESCRIPTION
@@ -20,6 +20,7 @@ Imports:
     usethis (>= 2.0.0)
 Suggests:
     devtools,
+    jsonlite,
     knitr,
     pkgbuild,
     pkgload,
--- a/minirextendr/R/feature-detect-configure.R
+++ b/minirextendr/R/feature-detect-configure.R
@@ -349,13 +349,58 @@ print.miniextendr_cargo_features <- function(x, ...) {
 
 #' Parse cargo metadata JSON to extract features and optional deps
 #'
-#' Minimal JSON parser using regex — handles the specific structure of
-#' `cargo metadata --no-deps` output without requiring jsonlite.
+#' Dispatcher: uses [jsonlite::fromJSON()] when the package is available
+#' (the common case in a dev environment — jsonlite is transitively present
+#' almost everywhere), otherwise falls back to the dependency-free regex
+#' parser. Both paths return the same shape. This runs maintainer-side only
+#' (via [list_cargo_features()]); it is NOT part of the configure-time
+#' feature detection, which stays base-R.
 #'
 #' @param json Raw JSON string from cargo metadata
 #' @return List with `features` and `optional_deps`
 #' @noRd
 parse_cargo_metadata_json <- function(json) {
+  if (requireNamespace("jsonlite", quietly = TRUE)) {
+    parse_cargo_metadata_jsonlite(json)
+  } else {
+    parse_cargo_metadata_regex(json)
+  }
+}
+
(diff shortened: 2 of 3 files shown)


CGMossa:
Duplicate of #1238 (same commit 73b5d022). I opened this by salvaging the subagent's branch prematurely while the subagent was in fact resuming and self-completing as #1238, which has the cleaner branch name and fuller body. Closing this one; #1238 is the canonical PR for #1064.

PR #1238, opened 2026-07-09
perf(minirextendr): jsonlite fast path for parse_cargo_metadata_json (#1064)

### LINKED ISSUE #1064: Simplify parse_cargo_metadata_json with a jsonlite-if-available fast path

CGMossa:
Adds a jsonlite fast path to `parse_cargo_metadata_json()` in minirextendr, keeping the dependency-free regex parser as the fallback. Closes #1064.

<details>

## Summary
- Renamed the existing hand-rolled regex parser to `parse_cargo_metadata_regex()` (behavior unchanged).
- Added `parse_cargo_metadata_jsonlite()`, a ~10-line parser built on `jsonlite::fromJSON(..., simplifyVector = FALSE)`.
- Turned `parse_cargo_metadata_json()` into a dispatcher: it uses jsonlite when `requireNamespace("jsonlite", quietly = TRUE)` succeeds, otherwise the regex fallback. Both paths return byte-identical output (verified with `identical()`) across the four cargo-metadata shapes the code parses (features, single/multiple optional deps, empty).
- Maintainer-side only: this is reached via `list_cargo_features()`. The configure-time `parse_cargo_features()` line-scanner is untouched and stays base-R, so no new dependency leaks into `./configure`.
- Added `jsonlite` to `Suggests`.

## Tests
- Existing parser cases now call `parse_cargo_metadata_regex()` directly, so they run without jsonlite.
- A parallel block exercises the jsonlite path under `skip_if_not_installed("jsonlite")`, plus a test asserting both paths agree on identical input and a dispatcher shape test.
- `test-feature-detect-configure.R` (jsonlite present, so both paths executed, not skipped): `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 85 ]`
- `test-exports-exist.R` (guards against a rename regressing the `list_cargo_features` export): `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 1 ]`

## `list_cargo_features()` / `print.miniextendr_cargo_features` assessment
Per the issue, evaluated whether these earn their keep. Recommendation: keep, no removal commit. Both are exported, documented (`man/list_cargo_features.Rd`), covered by `test-exports-exist.R`, and provide real maintainer value (the `$without_rules` view surfaces which Cargo features still lack a detection rule). Not dead code.

---

</details>

### DIFF
--- a/minirextendr/DESCRIPTION
+++ b/minirextendr/DESCRIPTION
@@ -20,6 +20,7 @@ Imports:
     usethis (>= 2.0.0)
 Suggests:
     devtools,
+    jsonlite,
     knitr,
     pkgbuild,
     pkgload,
--- a/minirextendr/R/feature-detect-configure.R
+++ b/minirextendr/R/feature-detect-configure.R
@@ -324,7 +324,7 @@ print.miniextendr_cargo_features <- function(x, ...) {
       feat_str <- if (length(dep$features) > 0) {
         paste0(" [", paste(dep$features, collapse = ", "), "]")
       } else ""
-      cli::cli_li("{.val {name}} {dep$req}{feat_str}")
+      cli::cli_li("{.val {name}} {dep$version}{feat_str}")
     }
   }
 
--- a/minirextendr/R/feature-detect-configure.R
+++ b/minirextendr/R/feature-detect-configure.R
@@ -349,13 +349,58 @@ print.miniextendr_cargo_features <- function(x, ...) {
 
 #' Parse cargo metadata JSON to extract features and optional deps
 #'
-#' Minimal JSON parser using regex — handles the specific structure of
-#' `cargo metadata --no-deps` output without requiring jsonlite.
+#' Dispatcher: uses [jsonlite::fromJSON()] when the package is available
+#' (the common case in a dev environment — jsonlite is transitively present
+#' almost everywhere), otherwise falls back to the dependency-free regex
+#' parser. Both paths return the same shape. This runs maintainer-side only
+#' (via [list_cargo_features()]); it is NOT part of the configure-time
+#' feature detection, which stays base-R.
 #'
 #' @param json Raw JSON string from cargo metadata
 #' @return List with `features` and `optional_deps`
 #' @noRd
 parse_cargo_metadata_json <- function(json) {
+  if (requireNamespace("jsonlite", quietly = TRUE)) {
+    parse_cargo_metadata_jsonlite(json)
+  } else {
+    parse_cargo_metadata_regex(json)
+  }
+}
+
+#' Parse cargo metadata JSON via jsonlite (fast path)
+#'
+#' @param json Raw JSON string from cargo metadata
+#' @return List with `features` and `optional_deps`, matching
+#'   [parse_cargo_metadata_regex()]'s shape.
+#' @noRd
+parse_cargo_metadata_jsonlite <- function(json) {
+  pkg <- jsonlite::fromJSON(json, simplifyVector = FALSE)$packages[[1]]
+  features <- lapply(pkg$features, function(specs) as.character(unlist(specs)))
+  # An empty features object yields a *named* empty list from jsonlite; the
+  # regex fallback yields a plain one. Normalise so both paths are identical.
+  if (length(features) == 0L) features <- list()
+  optional_deps <- list()
+  for (dep in pkg$dependencies) {
+    if (isTRUE(dep$optional)) {
+      optional_deps[[dep$name]] <- list(
+        version = if (!is.null(dep$req)) dep$req else "*",
+        features = as.character(unlist(dep$features))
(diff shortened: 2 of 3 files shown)


CGMossa:

<details>

## Summary
- Verdict: this matches the design agreed in #1064 exactly (jsonlite fast path, regex kept as fallback, tests split so the fallback runs without jsonlite). Rebased onto current main (clean) and re-pushed. Recommend merge after the small follow-up below.
- Dependency posture is safe. `jsonlite` is added to `Suggests` only and gated behind `requireNamespace("jsonlite", quietly = TRUE)`. The sole caller is `list_cargo_features()` (maintainer-side), and the configure-time `parse_cargo_features()` line-scanner (the base-R `[features]` reader that runs at end-user install with no library) is untouched, so no new dependency reaches `./configure` or scaffolded packages. `jsonlite` is also a recursive dependency of `usethis` (a hard Import of minirextendr), so the fast path runs in virtually every dev environment.
- Parity is byte-identical. I sourced both parsers off the branch and ran `identical()` over the 4 test fixtures plus 3 cases the suite does not cover: a version req with an embedded space and comma (`">=1.0, <2.0"`), a dep with no `req` field (both default to `*`), and the real `cargo metadata --format-version=1 --no-deps` blob from `rpkg/src/rust` (51 features, 2 optional deps). All 7 identical.
- Measured perf (real 4.7 KB blob, 2000 iterations each, R 4.6): regex median 2942 us (IQR 2876 to 3032), jsonlite median 85 us (IQR 84 to 85), dispatcher-with-guard median 85 us. That is ~34.6x, and the `requireNamespace` guard costs ~1 us. Honest caveat: this parser is called once per `list_cargo_features()`, which itself shells out to `cargo metadata` (tens to hundreds of ms), so the absolute latency win is imperceptible. The real payoff is deleting a fragile hand-rolled regex/bracket-counting JSON parser, which is how #1064 framed it (a simplification, "low priority").
- Sibling sweep: no other hand-rolled cargo-metadata JSON parser exists. `cargo.R:1173` bracket-counts TOML (different format), and the configure-time line-scanner must stay base-R by design. Nothing else to migrate.

## Changes I pushed on top
- Rebased onto current main at `7306d26b` (clean, no conflicts).
- `17c74573`: fixed a latent bug in `print.miniextendr_cargo_features`. It read `dep$req`, but both parsers store the version under `dep$version`, so `req` was always NULL and glue collapsed the bullet, silently dropping the version from every optional-dep line (`serde [derive]` instead of `serde ^1.0 [derive]`). One-word fix plus a behavioral regression test that captures the print output via `cli::cli_fmt()` and asserts the version appears. This is the diagnostic surface #1064 explicitly asked to scrutinize, and it was partly broken. Full suite after the fix: `[ FAIL 0 | WARN 0 | SKIP 3 | PASS 725 ]`.

## Notes
- `list_cargo_features()` / `print.miniextendr_cargo_features` kept per the PR's assessment. With the print bug fixed they now render correctly, which strengthens the keep decision.

---

</details>

PR #1239, opened 2026-07-09
fix(macros): gate DataFrameRow parallel methods on api-side rayon feature (#1117)

### LINKED ISSUE #1117: macros: trybuild UI snapshots mismatch locally (unexpected_cfgs warnings from rayon-gated DataFrameRow expansion)

CGMossa:
Closes #1117. `#[derive(DataFrameRow)]` was stamping raw `#[cfg(feature = "rayon")]` onto its parallel `*_par` methods, which is evaluated against the *consumer* crate and so tripped `unexpected_cfgs` in every downstream package that derives `DataFrameRow` without a `rayon` feature of its own. This routes the cfg decision into miniextendr-api instead, and removes the now-obsolete warning blocks from the affected trybuild snapshots.

<details>

## Root cause
A `cfg` inside a derive macro is resolved against the *destination* crate, not the crate that defines the derive. The eight `#[cfg(feature = "rayon")]` sites the derive emitted (`from_rows_par`, `try_from_dataframe_par`, and the `rows_into_dataframe_par` / `rows_from_dataframe_par` trait overrides, in both the struct-align and enum-align paths) therefore referenced the *consumer's* `rayon` feature. A consumer that doesn't declare one (the common case, and the trybuild fixture crate) got an `unexpected_cfgs` warning on every derive, and the parallel path only ever activated by the accident of the consumer happening to name a feature `rayon`.

## Fix path taken (the plan's preferred, api-side gate)
- Added `miniextendr_api::__dataframe_row_when_rayon!` in `miniextendr-api/src/dataframe.rs`: two `#[macro_export] macro_rules!` definitions, one `#[cfg(feature = "rayon")]` (passes the wrapped items through), one `#[cfg(not(feature = "rayon"))]` (expands to nothing). The `#[cfg]` now lives in miniextendr-api, where `rayon` is always a declared feature.
- The derive emits an unconditional `::miniextendr_api::__dataframe_row_when_rayon! { ... }` call around each of the eight sites; miniextendr-api decides, on **its own** `rayon` feature, whether the methods appear. No `#[cfg]` reaches the consumer crate, so no `unexpected_cfgs`, and the parallel path tracks the api feature regardless of what the consumer names its own.
- This is consistent with the existing api-side gating: `DataFrameRowConvert` / `IntoDataFrame` / `FromDataFrame` already declare their `*_par` methods `#[cfg(feature = "rayon")]`, so the wrapped overrides appear/vanish in lockstep with the trait methods they override.

## Snapshots
- Surgically removed only the `unexpected_cfgs` / `rayon` warning paragraph from the 13 affected `tests/ui/*.stderr` snapshots (paragraph-level removal, error diagnostics left byte-for-byte intact). Did **not** run `TRYBUILD=overwrite`.
- Added `tests/ui/pass/derive_dataframe_no_rayon_no_unexpected_cfgs.rs`: a `#![deny(unexpected_cfgs)]` compile-pass fixture (struct + align enum) that fails to compile if the raw cfg is ever reintroduced, since the fixture crate declares no `rayon` feature.

## Test plan
- [x] `cargo test -p miniextendr-macros --test ui`: 0 `unexpected_cfgs` warnings anywhere in the output; the new pass fixture compiles clean under `#![deny(unexpected_cfgs)]`; the 6 snapshots whose only diff was the warning now match.
- [x] clippy_default, clippy_all (`full-codegen`+rayon), clippy_all_s7 (`full-codegen-s7`+rayon): all `-D warnings` clean. The `*_all` legs enable rayon, exercising the macro's body-expansion arm.
- [x] `just check`: rpkg (a rayon-enabled `#[derive(DataFrameRow)]` consumer) compiles, verifying the rayon-on expansion still builds.
- [x] `just fmt` clean; `just test`: 64 test groups pass; the only failure is the `--test ui` target, from the residual rustc-version skew below.

## Notes
- **Residual local-only skew (deliberately left).** 7 `derive_dataframe_*` UI snapshots still mismatch on the local toolchain (rustc 1.95), but the diffs are purely rustc-version rendering differences unrelated to this bug: E0080 note format (`= note: the failure occurred here` on newer stable vs the expanded `panic_fmt`/`const_format_args` lines on 1.95) and E0599 wording (`variant, associated function, or constant` vs `variant or associated item`). Verified each of the 7 diffs contains no `rayon`/`unexpected_cfgs` content. CI stable is authoritative for these snapshots; the committed CI-stable wording was left untouched, so CI is expected green.
- No `#[miniextendr]` exports change: for a rayon-enabled consumer the macro expands to exactly the previous cfg-true body, so `wrappers.R` / `NAMESPACE` / `man` are unaffected (no regeneration needed).

---

</details>


CGMossa:

<details>

## Summary
- Verdict: the fix is correct and complete against #1117. `__dataframe_row_when_rayon!` moves the `rayon` decision into miniextendr-api (the only crate where `rayon` is always a declared feature), the token-tree passthrough matcher handles every wrapped item shape, and exactly one macro variant is exported for any feature state.
- Verified all 8 former emission sites now route through the api-side macro (4 in `dataframe_derive.rs`, 4 in `dataframe_derive/enum_expansion.rs`) and that no `#[cfg(feature = "rayon")]` reaches consumer crates anywhere in the derive output. The only other emitted cfg is `target_arch = "wasm32"`, which check-cfg always accepts.
- Verified the new `#![deny(unexpected_cfgs)]` compile-pass fixture works as a regression guard (it fails if the raw cfg is ever reintroduced).
- Ran trybuild under a CI-equivalent toolchain (rustup `1.97.0`, minimal profile, matching `dtolnay/rust-toolchain@stable`): 140/140 green. This should be the first branch with a green Rust-tests leg since CI stable moved to 1.97.0 on 2026-07-07.
- Changes I pushed: merged `origin/main` (post #1207; agent force-push is blocked, so merge instead of rebase), re-blessed `fn_implicit_dots_name_conflict.stderr` for rustc 1.97's promotion of `varargs_without_pattern` from warn to deny (rust-lang/rust#145544), byte-verified against the output CI itself produced in the last red run, and updated one stale rustdoc line in `dataframe_derive.rs` that still described the raw consumer-crate cfg.
- Re-bless rationale and ordering: the `unexpected_cfgs` warnings were a real end-user regression on rustc >= 1.97 (two warnings per derive in any consumer without its own `rayon` feature, and broken `-D warnings` builds downstream), not mere snapshot noise. Fixing the emission first and only then blessing was the right order; blessing on main would have baked the warnings into the expected outputs and hidden the regression.

## Notes
- The 13 `derive_dataframe_*` snapshots needed no re-bless at all: with the warnings surgically removed by this PR, their committed content already matches CI's rustc 1.97 output byte for byte.
- Root cause of the recurring "local trybuild mismatch" confusion (the original #1117 report and this PR's residual-skew note): it is mostly the rust-src component, not the rustc version. With rust-src installed (typical dev machine, rust-analyzer setups), rustc renders stdlib-span diagnostics as source snippets with carets; CI's minimal profile renders the bare fallback (`= note: the failure occurred here`, and no snippet under `help:` blocks pointing at `$RUST` paths). Same rustc 1.97.0 on both sides. Five fixtures (the four E0080 collision cases plus `enum_struct_field_no_derive`) will therefore mismatch on any rust-src machine while matching on CI. To reproduce CI locally: `rustup toolchain install 1.97.0 --profile minimal`, then `RUSTUP_TOOLCHAIN=1.97.0 cargo test -p miniextendr-macros --test ui`.
- Verification on this branch: `cargo fmt --all` clean; `clippy_default`, `clippy_all`, and `clippy_all_s7` all clean under `-D warnings` on 1.97 (the `*_all` legs enable `rayon` and exercise the macro's passthrough arm); full `cargo test --workspace --locked --no-fail-fast` under the CI-equivalent toolchain: all 66 test targets green, including the trybuild ui suite at 140/140. The pre-fix CI run on this branch (before the re-bless commit) failed with exactly `1 of 140 tests failed` on the varargs fixture, which cross-validates the analysis. The `R CMD check / Linux oldrel-1` red on that same run was an HTTP 504 during a setup download, transient infra, rerun by this push.
- Interaction with the rest of the queue: every other open PR that triggers the Rust leg stays trybuild-red until this merges, and #1205's all-green is stale (its runs predate the 1.97 bump). #1207 already landed, which healed main's clippy; this PR heals the tests leg. Recommend merging it promptly so the fleet can go green on re-runs.

---

</details>

Issue #1104, opened 2026-06-28
Guard against aliasing &mut [T] slices from the same SEXP (debug-build check)


CGMossa:
## Summary

`impl TryFromSexp for &mut [T]` (`miniextendr-api/src/from_r.rs:624`) can hand out **aliasing `&mut` slices** when the same R vector is bound to two `&mut [T]` parameters of a `#[miniextendr]` function — a soundness foot-gun with no compile-time guard. Track a debug-build aliasing check.

## Background

R argument passing can bind the same SEXP to multiple parameters (e.g. `f(x, x)`). The `&mut [T]` conversion produces a `&mut` slice over the SEXP's data pointer; two such parameters over the same SEXP yield two live `&mut` slices aliasing the same memory — UB under Rust's aliasing rules. The impl's own safety note (`from_r.rs:619`) documents this but nothing enforces it.

In practice `#[miniextendr]` functions with multiple `&mut [T]` params are rare, so this is latent rather than urgent — but it's a real unsoundness hole.

## Options (audit's recommendation)

1. **Debug-build runtime check**: when a function has >1 `&mut [T]` parameter, compare the underlying SEXP pointers and `panic!` with a clear message on alias (debug only, zero release cost). Could live in the macro-generated wrapper (which knows the param shapes) rather than the conversion impl.
2. Document the foot-gun more prominently in the `#[miniextendr]` user docs.
3. Accept as inherent and rely on the safety note.

## Source

Audit: `audit/from_r.md` issue #1. Verified: `impl<T> TryFromSexp for &mut [T]` with aliasing safety note at `from_r.rs:619-624`.

<sub>Drafted by Claude (Opus 4.8) from the automated `audit/` review; filed at the maintainer's request.</sub>

CGMossa:
Implemented by PR #1240 (merged). Closing as done — verified during the 2026-07-11 issue-rectification sweep. The unguarded cross-argument aliasing sibling remains tracked separately as #1252.

Issue #202, opened 2026-04-17
Run Miri against try_recover_r_sexp to verify #63's cfg(miri) gate


CGMossa:
## Context

PR #197 (closes #63) gates the speculative sxpinfo read in \`try_recover_r_sexp\` under \`#[cfg(miri)]\`: the whole function becomes \`return None\` so callers fall through to the copy path, and Miri never sees the provenance-invalid load.

## Gap

I never actually ran \`cargo +nightly miri test\` against the change. The PR description noted this openly. The fix is a \`return None\` short-circuit so it's hard to get wrong, but without a CI check the \`#[cfg(miri)]\` branch will bit-rot silently.

## What we want

Either:

1. **Add a Miri job to CI** scoped to \`miniextendr-api/src/r_memory.rs\` (or the whole crate if feasible). A minimal test that exercises a round-trip through \`try_recover_r_sexp\` on a Rust-allocated buffer (so the speculative read would trip Miri if the gate were missing). Nightly-only, so it's fine as an informational job.

2. **Or: document a local reproduction** in \`docs/\` (or a repo-level \`tools/run-miri.sh\`) so reviewers can run it on demand. Acceptable if CI budget for nightly Miri is a concern.

Option 1 is better — silent rot is exactly the risk this issue is trying to surface.

## Notes

- Miri nightly may need feature subsetting: the full workspace pulls in Arrow / DataFusion / linkme dylibs that Miri can't handle. Scoping to \`miniextendr-api\` with minimal features should work.
- The relevant callers (Arrow integration, \`Cow<[T]>\` round-trip) are all feature-gated, so a default-features Miri run should hit the gate deterministically.

CGMossa:
Implemented by PR #1240 (merged). Closing as done — verified during the 2026-07-11 issue-rectification sweep.

PR #1240, opened 2026-07-09
fix(macros)+ci: debug guard for aliasing &mut [T] args (#1104) + informational Miri job (#202)

CGMossa:
Pairs the two mechanical items from the api batch: a debug-build guard that catches aliasing `&mut [T]` slices from one SEXP (#1104), and an informational nightly Miri job that keeps the `try_recover_r_sexp` provenance gate honest (#202).

<details>

## Summary

**#1104 (debug guard for aliasing `&mut [T]` arguments)**
- `impl TryFromSexp for &mut [T]` (and `Option<&mut [T]>`) borrows R's data pointer directly (`r_slice_mut`, no copy). Binding one R vector to two such parameters (`f(x, x)` from R) hands out two aliasing `&mut` slices over one buffer, which is undefined behavior.
- The generated C wrapper now emits a `debug_assert!` at the top of the wrapper body (both the main-thread and worker-thread variants), before any conversion, comparing the raw SEXP identities pairwise. On a shared SEXP it panics (converted to an R error by the surrounding unwind guard) naming both parameters. Zero cost in release (the assert compiles out).
- `match_arg` + `several_ok` `&mut [T]` params are excluded: they get their own owned `Vec<T>` storage and never alias R's buffer.
- Implementation lives in `miniextendr-macros/src/c_wrapper_builder.rs`: a free `is_mut_slice_family` classifier plus `CWrapperContext::build_alias_guard`, wired into both wrapper generators.
- rpkg fixtures (`rpkg/src/rust/conversions.rs`): `alias_probe(a: &mut [i32], b: &mut [i32])` and `debug_assertions_enabled() -> bool`. The probe gates the testthat because rpkg's default and CI cargo profile is `release` (its `[profile.release]` does not set `debug-assertions`), so the assert is compiled out there.
- testthat `rpkg/tests/testthat/test-alias-guard.R`: distinct vectors pass with no false positive (both increment in place); `alias_probe(x, x)` errors under a debug build, `skip()`ped when `debug_assertions_enabled()` is FALSE.
- Unit tests in `c_wrapper_builder/tests.rs` cover the classifier and the pairwise guard emission.
- `docs/TYPE_CONVERSIONS.md`: the page previously claimed `&mut [T]` was rejected at compile time, which is stale (it is supported, with fixtures). Rewrote the section to document the in-place zero-copy semantics and added the aliasing foot-gun plus the debug-guard note.

**#202 (informational Miri job for `try_recover_r_sexp`)**
- Added a Miri-only regression in `miniextendr-api/src/r_memory.rs` gated `#[cfg(all(test, miri))]`. It forces a non-zero `SEXPREC_DATA_OFFSET` so the speculative sxpinfo read path is reached, round-trips `try_recover_r_sexp` on a Rust-allocated buffer, and asserts `None`. With the existing `#[cfg(miri)]` no-op branch in place Miri stays clean; remove the gate and Miri flags the out-of-provenance `wrapping_byte_sub` read and the test fails. The module is gated on `miri` (not just `test`) so a normal `cargo test` neither runs the unsound read nor leaves an unused import.
- New `.github/workflows/miri-nightly.yml`: nightly cron (04:30 UTC, offset from the existing gctorture-nightly and weekly CI crons) plus `workflow_dispatch`. It runs `cargo +nightly miri test -p miniextendr-api --no-default-features --lib` scoped to that one test. It is informational: never runs on PRs, the `cargo miri test` step is `continue-on-error: true` with its outcome surfaced in the job summary, and it must not be added to required checks. A separate workflow file (rather than a job in ci.yml) avoids 7x-ing the weekly CI surface, since `dorny/paths-filter` marks everything changed on non-PR events.

## Test plan
- [x] `just fmt` clean
- [x] `just check` clean across all manifests (including rpkg with the new fixtures)
- [x] clippy 3 legs at `-D warnings`: `clippy_default`, `clippy_all` (full-codegen), `clippy_all_s7` (full-codegen-s7), all green (feature lists read from ci.yml)
- [x] `just test`: all suites pass except the 7 known `derive_dataframe_*` UI trybuild skews (local rustc 1.95.0 vs CI stable, per project memory; snapshots not overwritten). New macros unit tests pass.
- [x] rpkg double install with `CARGO_PROFILE=dev` (so `debug_assertions` is on), then `test-alias-guard.R`: 4 pass, 0 fail, 0 skip. `debug_assertions_enabled()` returns TRUE, `alias_probe(x, x)` errors with an "aliasing" message, distinct vectors pass with no false positive.
- [x] `cargo check -p miniextendr-api --no-default-features` compiles (the compile half of the Miri command)

## Notes
- Local Miri was NOT run: `cargo-miri` is not installed for the local nightly toolchain. Per the issue plan the deliverable is the CI job plus the gate test; the informational nightly job exercises it on CI. Verification recipe for a maintainer: `cargo +nightly miri test -p miniextendr-api --no-default-features --lib -- recover_from_rust_buffer_returns_none` should pass with the gate present, and fail (Miri provenance error) if the `#[cfg(miri)]` branch in `try_recover_r_sexp` is removed.
- In a release build the alias guard is compiled out, so it is a debugging aid, not a correctness boundary; the docs say so and advise passing distinct vectors or using `&[T]` / `Vec<T>` when two arguments might reference the same vector.

---

</details>


CGMossa:
Re-reviewed after rebasing onto current `main`. Both halves hold up: the aliasing guard does the right thing and the Miri job is genuinely informational and green today. I pushed one hardening commit that extends the guard to the mixed `&mut [T]` + `&[T]` case, which is the same soundness hole and was slipping through. Details and my verification below.

## Summary

**Verdict: (A) guard LGTM after the pushed extension; (B) Miri job LGTM.** Rebased onto `origin/main` (was 12 behind, incl. macros-heavy #1219); clean, no conflicts on the touched files.

**(A) Aliasing guard (#1104)**
- Semantics are correct. The guard compares raw **SEXP identity**, not the data pointer. That is the right call: the common aliasing case (`f(x, x)`, or `y <- x; f(x, y)`) is literally the same SEXP passed twice, and identity comparison sidesteps the empty-vector false positive (two distinct empty vectors share R's `0x1` sentinel data pointer but are different SEXPs, so a data-pointer comparison would wrongly fire; identity does not).
- N-arg generalization is correct (pairwise double loop).
- Emitted in the C wrapper (both main-thread and worker-thread bodies), before conversion, inside the unwind guard. It covers free fns, inherent-impl methods, and trait-impl methods: all three builder sites strip the `self` receiver before constructing `CWrapperContext` (`sig_without_env` / `filtered_inputs`), so the `inputs`-to-`sexp_idents` zip stays aligned. `match_arg` + `several_ok` params are correctly excluded (they own a `Vec<T>`).
- Gap found and fixed: the original guard only fired for two-or-more `&mut [T]` params. But `&[T]` is also a zero-copy view over R's buffer (`from_r.rs` `as_slice`), so a `&mut [T]` paired with a `&[T]` bound to the same SEXP is a live mutable+shared borrow over one buffer, which is also UB, and the guard missed it. My commit `a6a8a49c` classifies each slice param as `Mut`/`Shared` (`slice_borrow_kind`) and emits the pairwise `debug_assert` for any pair where at least one borrow is mutable; two shared `&[T]` reads stay unguarded (sound). Docs, the `&mut [T]` safety note, and unit tests updated (mut+shared emission, shared+shared non-emission, 3-arg pairwise = 3 asserts).
- `Box<[T]>` / `Vec<T>` copy and are correctly not classified. The debug-only design is exactly what #1104 requested ("debug only, zero release cost"); the docs are explicit that it is not a release-mode correctness boundary.

**(B) Miri job (#202)**
- Informational and non-gating, verified: separate workflow file, triggers are `schedule` + `workflow_dispatch` only (never `pull_request`/`push`, so it cannot appear as a PR check or block a merge), the `cargo miri test` step is `continue-on-error: true` with the outcome surfaced in the job summary, and the header says do not add it to required checks.
- What runs under Miri: exactly one test, `recover_from_rust_buffer_returns_none`, scoped by `-p miniextendr-api --no-default-features --lib -- <name>`. The test module is `#[cfg(all(test, miri))]`; the R-touching tests are filtered out by name (they cannot run under Miri, which has no R runtime). Toolchain is handled correctly: explicit `cargo +nightly` overrides the repo's `rust-toolchain.toml` (`channel = "stable"`), and a dummy `R_HOME` satisfies `build.rs` without installing R. Action versions match the rest of the repo (`checkout@v7`, `rust-cache@v2`, `rust-toolchain@nightly`).
- Health, verified locally with the exact CI command: **passes green today** (1 passed, 298 filtered out). I also confirmed it has teeth: temporarily neutralizing the `#[cfg(miri)]` gate in `try_recover_r_sexp` makes Miri flag the speculative sxpinfo read (`memory access ... points to before the beginning of the allocation`) and the test fails. Reverted after.

SHAs: PR original (rebased) `b121172e`, hardening `a6a8a49c`.

## Notes
- Runtime-verified the guard end to end with a `CARGO_PROFILE=dev` (debug_assertions on) install: `alias_probe(x, x)` raises the aliasing R error, `alias_probe(a, b)` with distinct vectors passes with no false positive and mutates both in place, `debug_assertions_enabled()` is TRUE. The accidental first (release) install confirmed the guard compiles out in release. `test-alias-guard.R` passes 4/4 under the dev build.
- The `Vec<&'static mut [T]>` / `Vec<Option<&'static mut [T]>>` conversions already guard intra-list aliasing always-on (they collect element data pointers, skip empties, and return a clean `Err` on duplicates), so that sibling surface is not a gap. Worth noting the asymmetry: that path is always-on and error-returning, while the per-parameter guard is debug-only, by #1104's design.
- Residual sibling filed as #1252: cross-argument aliasing between a `Vec<&mut [T]>` list element and a direct slice parameter (`f(list(v), v)`) is covered by neither guard. Exotic and latent; tracked, not fixed here (needs a mechanism spanning conversion sites).
- The Miri workflow does not run on PRs by design, so it will not show in this PR's checks; I validated it locally instead.
- 5 `derive_dataframe_enum_*` UI trybuild snapshots mismatch locally; these are the known toolchain/rust-src skew (CI is authoritative), unrelated to this PR, not overwritten.

</details>

Issue #1133, opened 2026-07-03
post-review cleanups for #1131 (thread-pool control)


CGMossa:
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h4>AI-written details</h4></summary>

An 8-angle review of PR #1131 surfaced cleanup findings that were deliberately not folded into the PR's CI-fix push (correctness holes were fixed in-PR; these are quality follow-ups). One issue to keep them tracked, per the concessions rule:

1. **Test subprocess over-spawning** (`rpkg/tests/testthat/test-rayon.R`): the thread-control tests spawn ~9 callr subprocesses, each paying full R startup + a ~90 MB `.so` load. The "defaults to available parallelism" scenario duplicates the uncapped baseline spawn byte-for-byte, and the "errors once built" scenario could piggyback on any pool-building subprocess by returning a list. ~4 spawns (~5-10 s of check wall-clock) are avoidable with identical coverage.
2. **`run_with_env` duplicates `run_isolated`** (`rpkg/tests/testthat/test-subprocess-isolated.R:20`): near-verbatim copies of the fresh-subprocess-eval pattern in sibling test files (a third variant `run_load_in_locale` lives in `test-encoding.R:143`). Extend `run_isolated` with an `env_vars` parameter, promote it to a `helper-*.R`, and use it from all three files.
3. **Precedence table restated in 4 render targets**: `parallel.rs` module rustdoc, `docs/RAYON.md` (canonical), `thread_control.rs` roxygen, `use_rayon()` roxygen → `use_rayon.Rd`. The next precedence change requires four synchronized edits with no sync check. Keep two canonical copies (rustdoc for docs.rs, RAYON.md for the site) and reduce the roxygen copies to pointers.
4. **Third env-flag truthiness convention**: `parallel.rs::is_truthy` treats `yes` as truthy while `backtrace.rs:43` treats `yes` as false (`eq_ignore_ascii_case("true") || v == "1"`). Also `is_truthy` diverges from R's own `config_val_to_logical` (which accepts `0`/`no` as false). Extract one shared env-flag helper with explicit truthy/falsy sets.
[...]

PR #1241, opened 2026-07-09
refactor: post-review cleanups for thread-pool control (#1133)

CGMossa:
Post-review cleanups for the thread-pool control work (#1131), tracked in #1133. Six independent items; each is a separate commit so they're individually reviewable.

<details>
<summary>AI-generated summary of the six items (landed/declined + verification)</summary>

### Items

**1. Trim callr over-spawning in `test-rayon.R` — LANDED.**
9 subprocess spawns → **5**, with equivalent coverage:
- "defaults to available parallelism" folded into the uncapped-baseline spawn.
- "errors once built" piggybacked onto the `set_threads` spawn (`set(2)` builds the pool, so the follow-up `set(4)` must fail with "already built" in the same process — also asserts set-succeeds → 2).
- The `FALSE`/`false`/`""` falsy matrix collapsed to one representative (`"FALSE"`); the per-value set is already covered by the Rust unit test `resolve_cran_flag_treats_false_and_empty_as_unset`.

Wall-clock (local, whole file): **~3.09s → ~2.10s (~1.0s, ~32% faster)** for the 4 fewer spawns (~0.25s/spawn locally; larger on CI where cold R startup dominates each `callr::r`).

**2. Dedupe the fresh-subprocess helpers — LANDED (partial by design).**
Promoted `run_isolated(expr, env_vars = character(), timeout = 30)` to `rpkg/tests/testthat/helper-subprocess.R` (auto-sourced by testthat), extended with an `env_vars` param. `test-rayon.R` now builds its resolver-var blanking via a trivial `rayon_env()` and calls the shared helper (dropped the duplicate `run_with_env`); `test-subprocess-isolated.R` dropped its local copy.
`run_load_in_locale` (`test-encoding.R`) **intentionally left separate**: it tests whether `library(miniextendr)` itself *fails* under a bad locale (library-load wrapped in `tryCatch`, structured return) — the opposite contract from `run_isolated`'s "load succeeds, then eval". Folding it in would need a second code path and lose clarity; the plan explicitly said to fold only if clean.

**3. Precedence table 4 → 2 copies — LANDED.**
Kept the two canonical copies (`parallel.rs` rustdoc for docs.rs; `docs/RAYON.md` for the site). Reduced the other two to one-line pointers: `thread_control.rs`'s `miniextendr_num_threads` roxygen, and minirextendr's `use_rayon()` roxygen (kept the actionable `ensure_pool()` guidance, dropped the env-var enumeration). Regenerated `rpkg/man/thread_control.Rd` + `minirextendr/man/use_rayon.Rd`; no `NAMESPACE` change.

**4. One env-flag truthiness helper — LANDED.**
`MINIEXTENDR_BACKTRACE` (`backtrace.rs`, `true`/`1` only) and `_R_CHECK_LIMIT_CORES_` (`parallel.rs`, "anything but false") had divergent ad-hoc checks. Added `env_flag::parse_bool`, aligned with R's `config_val_to_logical`: case-insensitive, whitespace-trimmed `yes`/`true`/`1`/`on` vs `no`/`false`/`0`/`off`/empty; unrecognized → `None` so each caller keeps its own default. `backtrace.rs` → default off on unrecognized (now also honors `yes`/`on`); `is_truthy` → **fails safe toward limiting** on unrecognized (documented divergence — under a CRAN check we'd rather cap cores).

**5. De-publicize the pool choke point — LANDED.**
`par_iter_ready` was an overridable provided trait method, so a downstream `impl RParallelIterator` could silently skip `ensure_pool()`. Replaced it with a module-private free fn `par_iter_ready(&source)`; the trait now exposes only `par_iter`, and the CRAN-compliance guarantee can't be bypassed. Confirmed no downstream override (grep of `rpkg` + `tests/cross-package`).

**6. `RAYON_NUM_THREADS` parse parity — LANDED.**
Verified rayon-core 1.13.0's actual behavior. The `RAYON_*` vars are now parsed exactly as rayon-core (`usize::from_str`, **no whitespace trim**), and the deprecated `RAYON_RS_NUM_CPUS` fallback is now honored (after `RAYON_NUM_THREADS`, matching rayon-core). Two deliberate, documented divergences: `MINIEXTENDR_NUM_THREADS` (our own var) stays lenient and trims; `RAYON_NUM_THREADS=0` is treated as *unset* (falls through to the CRAN cap) rather than rayon's "0 = all cores", so a stray `0` can never bypass `_R_CHECK_LIMIT_CORES_`. New unit tests cover each case; precedence docs updated with the new row + a "Divergence from rayon-core" note.

### Verification
- `just fmt` — clean.
- Clippy `-D warnings`, all three CI legs: `clippy_default`, `clippy_all` (`full-codegen` + integration list), `clippy_all_s7` (`full-codegen-s7`) — all pass.
- `just check` — pass (all manifests).
- Rust unit tests: `env_flag::tests` (3) + `optionals::parallel::tests` (12, incl. the new item-6 cases) — 15 passed / 0 failed.
- `just rcmdinstall` + `just force-document` — clean; only `thread_control.Rd` / `use_rayon.Rd` regenerated (intended), `NAMESPACE` unchanged.
- `test-rayon.R` (item 1/2 load-bearing): all thread-control tests green.
- `test-subprocess-isolated.R` (shared helper, `NOT_CRAN=true`): 14 tests green.
- `just devtools-test` (full suite): the affected files pass (see the two targeted lines above). The full-suite run hung on an unrelated worker-thread panic fixture (`rpkg/src/rust/worker_tests.rs`, `"Worker: about to panic"`) — a pre-existing hazard-class test, the very kind `test-subprocess-isolated.R` was created to isolate — and was killed. It touches none of the files in this PR, and the `backtrace.rs` change is behaviorally identical when `MINIEXTENDR_BACKTRACE` is unset (the test case), so it cannot introduce the hang.
- The 7 known `derive_dataframe_*` UI trybuild skews are unrelated to this PR and were not touched.

No items declined; no deferrals.

</details>

### DIFF
--- a/docs/RAYON.md
+++ b/docs/RAYON.md
@@ -597,14 +597,24 @@ rule:
 |---|---|---|
 | 1 | `MINIEXTENDR_NUM_THREADS` env var | Explicit override — wins over everything |
 | 2 | `RAYON_NUM_THREADS` env var | Rayon's own convention, respected as-is |
-| 3 | `_R_CHECK_LIMIT_CORES_` env var (truthy) | Caps at `min(2, available_parallelism())` — CRAN's `--as-cran` policy sets this to `"TRUE"` |
-| 4 | (none of the above) | `std::thread::available_parallelism()` |
+| 3 | `RAYON_RS_NUM_CPUS` env var | Rayon's deprecated fallback, still honored (after `RAYON_NUM_THREADS`) |
+| 4 | `_R_CHECK_LIMIT_CORES_` env var (truthy) | Caps at `min(2, available_parallelism())` — CRAN's `--as-cran` policy sets this to `"TRUE"` |
+| 5 | (none of the above) | `std::thread::available_parallelism()` |
 
-`_R_CHECK_LIMIT_CORES_` truthiness follows R's own convention: unset, empty,
-or `"false"`/`"FALSE"` count as not-limited; anything else (including R's
-`"TRUE"`) caps at 2. Because rayon's global pool cannot be resized once
-built, this resolution happens exactly once per process — the first
+`_R_CHECK_LIMIT_CORES_` truthiness follows R's own `config_val_to_logical`
+convention: unset, empty, or `"no"`/`"false"`/`"0"`/`"off"` count as
+not-limited; `"yes"`/`"true"`/`"1"`/`"on"` (and, failing safe, any
+unrecognized value) cap at 2. Because rayon's global pool cannot be resized
+once built, this resolution happens exactly once per process — the first
 parallel call locks it in.
+
+The two `RAYON_*` vars are parsed exactly as `rayon-core` parses them
+(`usize::from_str`, no whitespace trimming), so a value rayon would reject we
+reject too. Two deliberate exceptions: (1) we treat `RAYON_NUM_THREADS=0` as
+*unset* — falling through to the `_R_CHECK_LIMIT_CORES_` cap — rather than
+rayon-core's "0 means use the default (all cores)", so a stray `0` can never
+bypass CRAN's core limit; (2) `MINIEXTENDR_NUM_THREADS`, being our own var, is
+lenient and trims surrounding whitespace.
 
 If your own code calls `rayon::ThreadPoolBuilder::build_global()` before any
 miniextendr rayon call, that wins outright: `ensure_pool()` sees the pool
--- a/miniextendr-api/src/backtrace.rs
+++ b/miniextendr-api/src/backtrace.rs
@@ -21,10 +21,11 @@ static INSTALLED: AtomicBool = AtomicBool::new(false);
 
 /// Register the miniextendr panic hook.
 ///
-/// If `MINIEXTENDR_BACKTRACE` is set to `true` or `1`, the default Rust
-/// panic hook runs (full traceback printed to stderr); otherwise the hook
-/// swallows the panic output silently so the R error (emitted by
-/// `panic_message_to_r_error`) is what users see.
+/// If `MINIEXTENDR_BACKTRACE` is truthy (`yes`/`true`/`1`/`on`, per
+/// [`crate::env_flag::parse_bool`]), the default Rust panic hook runs (full
+/// traceback printed to stderr); otherwise the hook swallows the panic output
+/// silently so the R error (emitted by `panic_message_to_r_error`) is what
+/// users see. Unrecognized values default to off.
 ///
 /// Idempotent within a DLL instance: the first call installs, subsequent
 /// calls are no-ops. If the DLL is unloaded and loaded again, the new
@@ -41,7 +42,8 @@ pub extern "C-unwind" fn miniextendr_panic_hook() {
     let default_hook = std::panic::take_hook();
     std::panic::set_hook(Box::new(move |x| {
         let show_traceback = std::env::var("MINIEXTENDR_BACKTRACE")
-            .map(|v| v.eq_ignore_ascii_case("true") || v == "1")
+            .ok()
+            .and_then(|v| crate::env_flag::parse_bool(&v))
             .unwrap_or(false);
         if show_traceback {
             default_hook(x)
--- a/miniextendr-api/src/env_flag.rs
+++ b/miniextendr-api/src/env_flag.rs
@@ -1,0 +1,55 @@
+//! Shared truthiness parsing for environment-variable flags.
+//!
+//! Two flags in this crate read boolean-ish env vars — `MINIEXTENDR_BACKTRACE`
+//! (`backtrace.rs`) and `_R_CHECK_LIMIT_CORES_` (`optionals/parallel.rs`) — and
+//! they used to disagree on what counts as "on" (`true`/`1` vs. anything but
+//! `false`). [`parse_bool`] is the single source of truth, deliberately aligned
+//! with R's own `config_val_to_logical` (`tools/R/utils.R`): a case-insensitive,
+//! whitespace-trimmed match against explicit truthy/falsy sets.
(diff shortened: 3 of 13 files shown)


CGMossa:
Rebased onto current `main` and re-verified this end to end; it's clean and complete, and I'd merge it. The two red checks on the previous run were both pre-existing and had nothing to do with this diff (the `state.rs:113` `question_mark` clippy lint on CI's rust-1.97.0, and the `derive_dataframe_*` trybuild UI skew), and `main` already carries the fixes for both, so the rebase I just pushed should come back green. All six review items from #1133 are addressed; the one deviation (item 2's third helper) is deliberate and documented.

## Summary

Verdict: approve after the rebase. No code changes were needed beyond the rebase; the PR faithfully implements the #1133 follow-up review.

Rebased `worktree-agent-a175c938187892e11` from base `e3826c55` onto `0a58ce5a` (clean, no conflicts). New commits: `81738b60`, `f6e3250b`, `9b28fee3`, `13747730`, `666d7b2d`. Force-pushed with lease.

Checklist vs the original #1133 review (the 6 items spun out of the #1131 8-angle review):

1. Over-spawning in `test-rayon.R`: ADDRESSED. 9 callr spawns to 5. "defaults to available" folded into the uncapped baseline; "errors once built" piggybacked on the `set_threads` spawn (set(2) builds the pool so set(4) must fail "already built"); the FALSE/false/"" matrix collapsed to one representative, with the per-value coverage living in the Rust unit test `resolve_cran_flag_treats_false_and_empty_as_unset`. Verified 5 spawns (the "rejects non-positive" case runs in-process).
2. `run_with_env` duplicates `run_isolated`: ADDRESSED (partial by design). `run_isolated(expr, env_vars, timeout)` promoted to `helper-subprocess.R` (auto-sourced) and consumed by both `test-rayon.R` (via a trivial `rayon_env()`) and `test-subprocess-isolated.R`. The env merge is now a proper replace-or-add over `rcmd_safe_env()` rather than the old `c()` concat, which is cleaner. `test-encoding.R`'s `run_load_in_locale` is left separate: it asserts `library(miniextendr)` itself FAILS under a bad locale (load wrapped in tryCatch, structured return), the opposite contract from run_isolated. The original ask said "all three files"; leaving the third out is a reasoned, documented call, not an oversight.
3. Precedence table in 4 render targets: ADDRESSED. Two canonical copies kept (`parallel.rs` rustdoc, `docs/RAYON.md`); `thread_control.rs` roxygen and `use_rayon()` roxygen reduced to pointers, with `thread_control.Rd` and `use_rayon.Rd` regenerated to match. Swept tracked source: the full table survives in exactly those two places, no orphaned 5th copy.
4. Third env-flag truthiness convention: ADDRESSED. New `env_flag::parse_bool`, aligned with R's `config_val_to_logical` (case-insensitive, trimmed; yes/true/1/on vs no/false/0/off/empty; unrecognized to None). Both prior callers adopt it. Grep confirms no other ad-hoc boolean env parsing remains in the api crate, so no missed sibling.
5. `par_iter_ready` public overridable method: ADDRESSED. Replaced by a module-private free fn `par_iter_ready<T: RParallelIterator + ?Sized>(&T)`; the trait now exposes only `par_iter`, so the `ensure_pool()` choke point can't be bypassed by a downstream impl. All ~28 default methods updated; grep confirms the method form is gone (remaining hits are stale sibling worktrees, not this branch).
6. `RAYON_NUM_THREADS` parse parity: ADDRESSED. `RAYON_*` parsed exactly as rayon-core (`usize::from_str`, no trim); deprecated `RAYON_RS_NUM_CPUS` honored after `RAYON_NUM_THREADS`. Two documented divergences: `MINIEXTENDR_NUM_THREADS` stays lenient/trimmed, and `RAYON_NUM_THREADS=0` is treated as unset (falls through to the CRAN cap) so a stray 0 can't bypass `_R_CHECK_LIMIT_CORES_`. Covered by new unit tests.

Thread-pool hazards checked:
- Global-pool-once: unchanged; `ensure_pool()` stays `Once`-guarded and defers to a user-built global pool. The refactor is purely method to free fn.
- Thread-count validation incl. NA (i32::MIN): `miniextendr_set_threads` uses `usize::try_from(n)`, which rejects a negative (NA_integer_ is i32::MIN) before it can become a huge usize; the resolver only ever reads env-var strings, and non-numeric/zero fall through. Sound.
- Rayon-workers-never-touch-R: the closures under `par_iter_ready` operate only on numeric `Item` types (`Into<f64>`/`Into<i32>`/`Ord`); no SEXP crosses into a rayon closure. No new R-on-worker path.
- Env-var vs setter precedence: documented in the precedence table and rustdoc divergence note.

Verification (rebased tree, R 4.6.0):
- `cargo fmt --all --check`: clean.
- `cargo clippy -p miniextendr-api --all-targets --features rayon --locked -- -D warnings`: exit 0.
- Rust unit tests: `env_flag::tests` 3/3; `optionals::parallel::tests` 12/12 (incl. new `resolve_rayon_zero_falls_through_to_cap`, `resolve_rayon_vars_match_rayon_no_trim`, `resolve_honors_deprecated_rayon_rs_num_cpus`).
- R: `test-rayon.R` thread-control tests green (0 fail, 0 warn) exercising the shared helper across the reduced spawns; `test-subprocess-isolated.R` 12/12 green under `NOT_CRAN=true`.

## Notes

- The rayon-off default clippy leg is safe: the trait, its default methods, and the new free fn are all rayon-gated together (the trait signature uses `rayon::iter::ParallelIterator`), so there is no config where the fn is defined but unused.
- No deferrals, no new issues filed; nothing was declined that needs tracking.

</details>

Issue #1271, opened 2026-07-10
CI: monorepo-template wasm path is still not built end-to-end (standalone template covered by #1259 leg)


CGMossa:
The scaffold wasm leg added for #1259 builds a package scaffolded from the **standalone** template tree (`minirextendr/inst/templates/rpkg/`, via `create_miniextendr_package()`). The **monorepo** template tree (`minirextendr/inst/templates/monorepo/rpkg/`, via `create_miniextendr_monorepo()` / `use_miniextendr(template_type = "monorepo")`) carries its own copies of `configure.ac`, `Makevars.in`, and `build.rs` — including the wasm branches — and those copies remain CI-unbuilt.

Today the two trees' wasm-relevant lines are kept in lockstep only by review + `just templates-check` against `patches/templates.patch`; a monorepo-template-only regression in the wasm path would still surface for end users first.

Sketch:
- Extend the webr.yml scaffold leg (or add a matrix dimension) to also run `create_miniextendr_monorepo("/tmp/scaffold-mono/mymono", rpkg_name = ...)`, wire `use_local_miniextendr()` on the rpkg subdir, and repeat the native → roxygenise → native → `CC=emcc` install sequence. Load it in tier 3 with a second `SMOKE_SCAFFOLD_PKG`-style env (the smoke runner would need to accept a list).
- Cost: roughly doubles the scaffold leg (~2 extra compiles of a minimal crate — one native, one wasm+build-std). Consider gating to main-push + workflow_dispatch if PR latency matters.
- Alternative cheaper interim: a tier-1-style `cargo check --target wasm32-unknown-emscripten` of a freshly scaffolded monorepo crate, which at least covers the template `build.rs`/cfg-gating without emcc.

Relates: #1259 (standalone scaffold leg), #1270 (local smoke parity), #495, #1079.

Issue #1270, opened 2026-07-10
tests/webr-smoke.sh: add local parity for the CI scaffold wasm leg (#1259)


CGMossa:
The scaffold leg added for #1259 (webr.yml `webr-install` job) exercises the end-user scaffolded-package wasm path in CI: install minirextendr from the checkout, `create_miniextendr_package()`, `use_local_miniextendr()`, native install ×2 with a `roxygen2::roxygenise()` pass, `CC=emcc` install into `/tmp/wasm-lib`, and a tier-3 load via `SMOKE_SCAFFOLD_PKG=mxsmoke`.

`tests/webr-smoke.sh` — the local docker smoke that otherwise mirrors the tier-2/3 job step-for-step — does **not** run this leg, so local runs can't reproduce a scaffold-leg CI failure without hand-driving the steps inside the container.

Sketch:
- Add a `phase_scaffold()` between `phase_wasm_build` and `phase_webr_session`, mirroring the four CI steps (minirextendr + roxygen tooling install into a temp lib, scaffold to `/tmp/webr-smoke/scaffold/mxsmoke`, native→roxygenise→native→wasm installs, install into `/tmp/wasm-lib`).
- Export `SMOKE_SCAFFOLD_PKG=mxsmoke` for the `smoke.mjs` invocation in `phase_webr_session` (the runner is already env-gated; unset keeps today's behavior).
- Gate it behind a `--scaffold` flag (or `WEBR_SCAFFOLD=1`) so the default local loop stays fast; CI parity is the opt-in.
- Cleanup trap should remove the scaffold dir and its temp libs.

Notes / traps learned from the CI leg:
- The container installs R packages from P3M as *source* unless `HTTPUserAgent` is set to the binary-serving format — see the "Scaffold leg — install minirextendr + roxygen tooling" step in `.github/workflows/webr.yml`.
- `git init` the scaffold before configure so a future cargo-revendor-on-PATH can never flip it into tarball mode (which would silently build the published git sources instead of the checkout).
- The scaffolded NAMESPACE is a stub until `roxygen2::roxygenise()` runs; without it `library()` loads no DLL (`useDynLib` missing) — that's why the leg does install → roxygenise → reinstall.

Issue #1079, opened 2026-06-24
Release reproducibility: rpkg framework git deps unpinned + committed Cargo.lock 27 commits stale


CGMossa:
Surfaced while investigating tagged-release / install mechanics (see `docs/RELEASING.md`). Two coupled findings mean **nothing committed in the tree self-describes a release's framework version**.

### Finding 1 — manifest git deps are unpinned
`rpkg/src/rust/Cargo.toml` declares the framework crates with a bare git source, no `tag`/`rev`:
```toml
miniextendr-api  = { git = "https://github.com/A2-ai/miniextendr" }
miniextendr-lint = { git = "https://github.com/A2-ai/miniextendr" }
```
So whenever the lock is (re)resolved, the framework floats to whatever `main` HEAD is at that moment.

### Finding 2 — committed `Cargo.lock` is stale
The committed `rpkg/src/rust/Cargo.lock` pins the framework to `git+...#ebd64a4f` (PR #1002) — **27 commits behind** current `main` (`d76345a3`). The committed lock isn't a meaningful pin for the exemplar.

### Why it mostly doesn't bite today (but is fragile)
- The **built tarball is fine**: `bootstrap.R` → `cargo-revendor` stamps the *current* sha into the shipped lock (verified — a fresh `just r-cmd-build` stamped `91d90ff9`, not the stale `ebd64a4f`), and vendors that source. Mode B (vendored tarball on a Release) is reproducible.
- `install_github` of this repo resolves the framework from the local **tag-state siblings** via configure's filesystem-probe `[patch]` (also verified), so it matches the tag — *as long as* the lock isn't regenerated to float to HEAD.
- The exposure is the gap between manifest (floats to HEAD) and committed lock (stale): anything that regenerates the lock offline/standalone silently changes the framework version.

### Options (flat priority)
1. Stamp `tag = "vX.Y.Z"` (or `rev`) into the framework deps at release time (e.g. extend `scripts/bump-version.sh`), making the manifest self-describing.
2. Periodically refresh the committed `Cargo.lock` so it tracks a recent sha.
3. Document that only the built vendored tarball is reproducible and direct users there (current de-facto state).

### Related
- #1052 (guard source-mode Cargo.lock dev-drift) — same file, different concern (CI drift guard vs release pinning).
[...]

PR #1272, opened 2026-07-10
ci(webr): exercise the scaffolded-package wasm path end-to-end (#1259)

### LINKED ISSUE #1259: CI: scaffolded-package wasm path is never exercised end-to-end (coverage transfers only via templates.patch)

<details>

## Summary
- The wasm branches in `minirextendr/inst/templates/*/configure.ac`, `Makevars.in`, and `build.rs` are never built by any CI job. `webr.yml` tiers 1-3 exercise `rpkg/` only; `templates-check` verifies the rpkg-to-templates delta matches the locked `patches/templates.patch`, so coverage transfers *indirectly* (the functional wasm lines are identical today, comment-only delta).
- The gap: a template-only regression introduced through a `templates-approve` refresh (or a scaffolder change that breaks the wasm path) would surface for end users, not in CI. The end-user story ("build your miniextendr package for webR") has no direct validator.

## Sketch
- Extend the `webr-install` job in `.github/workflows/webr.yml` (or add a sibling job in the same container) with a scaffold leg:
  1. Install minirextendr into a temp lib (pure R, no compile).
  2. `create_miniextendr_package()` into a temp dir with a trivial `#[miniextendr]` function.
  3. Point the scaffolded crate's framework deps at the checkout (the scaffold's git-URL deps would otherwise fetch from GitHub — pin via `[patch]` to the workspace, mirroring what `rpkg/configure` does in monorepo mode; relates to #1079's unpinned-git-deps concern).
  4. Native `R CMD INSTALL` (wrapper-gen pass writes the scaffolded package's `wasm_registry.rs`), then `CC=emcc` install with `webr-vars.mk` to an empty temp lib — same install-to-temp-lib pattern as tier 2.
  5. Optionally load it in the tier-3 webR session alongside miniextendr.
- Cost: roughly one extra wasm compile of a minimal crate (much smaller than rpkg's full feature surface since the scaffold has few deps); the job already pays for the container and toolchain.
- Cheaper interim alternative: a tier-1-style `cargo check --target wasm32-unknown-emscripten` of a freshly scaffolded crate, which at least catches template `build.rs` / cfg-gating breakage without the emcc link.

## Notes
- Origin: webR/wasm support review, 2026-07-10 (observation noted in PR #1256's notes).
- Related: #495 (cross-crate trait dispatch under wasm_registry codegen), #1079 (framework git deps unpinned).

---

</details>

CGMossa:

### Summary

The wasm branches in `minirextendr/inst/templates/rpkg/{configure.ac,Makevars.in,build.rs}` were never built by any CI job — webr.yml's tiers 1–3 exercise `rpkg/` only, and coverage transferred to the templates only via the locked `patches/templates.patch` delta. A template-only regression (e.g. through a `templates-approve` refresh) would have surfaced for end users, not in CI. This PR adds a **scaffold leg** to the existing `webr-install` job that runs the real end-user story end-to-end and loads the result in the tier-3 webR session.

### Implementation

- **`.github/workflows/webr.yml`** — five new steps in the `webr-install` job, between the rpkg tier-2 verify and tier 3 (same container, so the toolchain/image cost is paid once):
  1. Install minirextendr (from the checkout) + its Imports + roxygen tooling into the shared `R_LIBS_USER`. Imports are derived from `minirextendr/DESCRIPTION` (same single-source-of-truth pattern as the existing rpkg step). Sets the P3M binary-serving `HTTPUserAgent` so the usethis/roxygen2 tree installs as noble binaries instead of compiling stringi et al. from source.
  2. `create_miniextendr_package("/tmp/scaffold/mxsmoke")` — the stock template `lib.rs` already ships trivial `#[miniextendr]` fns `add()`/`hello()`, so the scaffold is used untouched. `use_local_miniextendr($GITHUB_WORKSPACE)` writes the `.miniextendr-local` marker so configure emits the `[patch."https://github.com/A2-ai/miniextendr"]` block — the same mechanism `rpkg/configure` uses in monorepo mode — and the leg validates the PR's code, not the published git sources (#1079 unchanged). The scaffold is then `git init`-ed so configure's auto-vendor branch can never flip it to tarball mode.
  3. Native `R CMD INSTALL` ×2 with a `roxygen2::roxygenise()` pass between them: install #1 runs the Makevars wrapper-gen pass (writes `R/mxsmoke-wrappers.R` + `src/rust/wasm_registry.rs`; asserted non-stub via content-hash), roxygenise turns the scaffolded NAMESPACE stub into the real one (`useDynLib` + exports — without this the installed package loads no DLL; found while validating locally), install #2 is warm-cargo cheap. Ends with a native runtime sanity call of `add()`/`hello()`.
  4. `CC=emcc` wasm install mirroring Phase 2 (`webr-vars.mk`, install into the same `/tmp/wasm-lib` so tier 3's existing NODEFS mount exposes both packages).
  5. Verify the side-module is WebAssembly, not ELF (same check as rpkg's).
- **`tests/webr-node-smoke/smoke.mjs`** — env-gated scaffold check (`SMOKE_SCAFFOLD_PKG`): `library(mxsmoke)` in the same webR session, then assert `add(2, 3) == 5` and `hello("webR") == "Hello, webR!"` — proof the template wasm path not only links but dispatches into Rust in a real webR runtime. Unset (e.g. `tests/webr-smoke.sh`), the runner behaves exactly as before.
- **Path filters** — `minirextendr/**` added to `pull_request`/`push` so template/scaffolder changes trigger the workflow.
- **`docs/WEBR.md`** — CI section documents the leg; the smoke.sh bullet notes the leg is CI-only for now.

The native half of the leg (scaffold → configure → patch override → install → roxygenise → reinstall → runtime calls) was validated locally on macOS before pushing; the emcc/webR half is validated by this PR's own workflow run.

### CI evidence

TODO(after CI): link the green `wasm32 R CMD INSTALL (tier 2)` run of this PR showing the "Scaffold leg — …" steps and the tier-3 `[tier3] OK: mxsmoke loaded` line.

### Cost

TODO(after CI): report measured wall-clock delta. Estimate: the leg adds one native release build of the framework crates plus one wasm `-Z build-std` build of a minimal crate (each in its own target dir — `RUSTFLAGS` link-args differ per package, so sharing rpkg's target dir would not hit cache) — roughly +15–20 min on the ~28 min baseline job, well inside the 90-min timeout. The R tooling install is binary-served (~1–2 min).

### Follow-ups (tracked)

- #1270 — `tests/webr-smoke.sh` local parity for the scaffold leg.
- #1271 — monorepo-template wasm path still CI-unbuilt (this leg covers the standalone `templates/rpkg/` tree).

Fixes #1259

</details>

### DIFF
--- a/.github/workflows/webr.yml
+++ b/.github/workflows/webr.yml
@@ -362,6 +362,14 @@ jobs:
       - name: Scaffold leg — native install + roxygen pass (wrappers, wasm_registry.rs, NAMESPACE)
         run: |
           set -euo pipefail
+          # The container's default locale is C. miniextendr's package init
+          # asserts l10n_info()[["UTF-8"]] and refuses to dyn.load otherwise
+          # (miniextendr-api/src/encoding.rs). The wrapper-gen dyn.load inside
+          # R CMD INSTALL takes the minimal init path (MINIEXTENDR_WRAPPER_GEN=1)
+          # and never hits the check, but roxygenise's pkgload load_dll and the
+          # sanity library() below run the full init — so this step needs a
+          # UTF-8 locale. noble ships C.UTF-8 without locale-gen.
+          export LANG=C.UTF-8 LC_ALL=C.UTF-8
           ( cd /tmp/scaffold/mxsmoke && bash ./configure )
           mkdir -p /tmp/scaffold-native-lib
           # Install #1: the Makevars wrapper-gen pass writes
--- a/.github/workflows/webr.yml
+++ b/.github/workflows/webr.yml
@@ -8,6 +8,10 @@ on:
       - 'miniextendr-engine/**'
       - 'miniextendr-lint/**'
       - 'rpkg/**'
+      # The scaffold leg (#1259) builds a freshly scaffolded end-user package
+      # from minirextendr's templates, so template/scaffolder changes must
+      # trigger this workflow too.
+      - 'minirextendr/**'
       - 'tests/cross-package/**'
       - 'tests/webr-node-smoke/**'
       - 'tests/webr-smoke.sh'
@@ -24,6 +28,7 @@ on:
       - 'miniextendr-engine/**'
       - 'miniextendr-lint/**'
       - 'rpkg/**'
+      - 'minirextendr/**'
       - 'tests/cross-package/**'
       - 'tests/webr-node-smoke/**'
       - 'tests/webr-smoke.sh'
@@ -134,12 +139,15 @@ jobs:
 
   # ── Tier 2 + 3: full wasm32 R CMD INSTALL + Node smoke ──────────────────
   # Closes #491 (tier 2) and #492 (tier 3 — Node + webR runtime smoke).
-  # Runs as a single job because all three phases share the same container,
-  # filesystem state, and `/tmp/wasm-lib/miniextendr` artifact:
+  # Runs as a single job because all phases share the same container,
+  # filesystem state, and `/tmp/wasm-lib/*` artifacts:
   #   - Phase 1: native install → regenerates wasm_registry.rs via cdylib pass.
   #   - Phase 2: CC=emcc wasm install via webr-vars.mk → /tmp/wasm-lib/miniextendr.
+  #   - Scaffold leg (#1259): the same native→wasm two-step on a package
+  #     freshly scaffolded from minirextendr's templates → /tmp/wasm-lib/mxsmoke.
   #   - Phase 3: Node + webR session NODEFS-mounts /tmp/wasm-lib, installs
-  #              Imports from repo.r-wasm.org, drives library(miniextendr).
+  #              Imports from repo.r-wasm.org, drives library(miniextendr)
+  #              and library(mxsmoke).
   # Mirrors tests/webr-smoke.sh end-to-end. CI's amd64 runners have no Rosetta
   # and plenty of disk (the local arm64 dev path hits both limits). Phase 2
   # is also the empirical validator for the side-module RUSTFLAGS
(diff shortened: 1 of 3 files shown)


CGMossa:

<details>

## Summary
- The tier-2 job failed twice on this branch, for two independent reasons; both are now addressed.
- First failure: the container's default locale is C, and miniextendr's full package init asserts `l10n_info()[["UTF-8"]]` before `dyn.load` (`miniextendr-api/src/encoding.rs`). The wrapper-gen pass inside `R CMD INSTALL` uses the minimal init path and never hit it, but `roxygen2::roxygenise()`'s pkgload `load_dll` does. Fixed by exporting `LANG=C.UTF-8 LC_ALL=C.UTF-8` in the scaffold leg's native install step.
- Second failure (surfaced once the leg got to tier 3): `mxsmoke::add(2, 3)` in the webR session errored with `failed to convert parameter 'left' to i32`. The parameter name `left` and the `i32` conversion only exist in rpkg's `add(left, right)` fixture (`rpkg/src/rust/panic_tests.rs:63`), not the template's `add(a: f64, b: f64)`. The generated C wrapper symbol is package-agnostic (`C_add`), and Emscripten side modules resolve exported function addresses through a shared GOT, so the scaffolded package's registration bound `C_add` to first-loaded miniextendr's copy and cross-dispatched. Native `dyn.load` is `RTLD_LOCAL`, which is why the identical call passed in the native sanity check of the same job.
- That second one is a genuine product bug affecting any two miniextendr-backed packages sharing an exported function name in one webR session; filed as #1273 with the full mechanism and a fix sketch (package-unique C symbols). The leg now renames the scaffold's stock functions to `mxsmoke_add`/`mxsmoke_hello` at scaffold time, with comments in `webr.yml` and `smoke.mjs` pointing at #1273 so the rename gets dropped when the product fix lands. The leg still exercises the full template build plumbing and real wasm dispatch.

## Notes
- The pre-existing rpkg tier-2 phases were green in both failing runs; both failures were confined to the new scaffold leg.

---

</details>

Issue #1310, opened 2026-07-12
webR: miniextendr-api's own #[no_mangle] exports are byte-identical across packages (GOT interposition surface)


CGMossa:
Follow-up carved out of #1273 (fixed by crate-prefixing all **macro-emitted** C symbols — see `plans/2026-07-10-issue-1273-c-symbol-crate-prefix.md`, work item 16). This issue tracks the remaining, unfixed half of the same webR mechanism.

## The mechanism (recap from #1273)

Native `dyn.load` uses `RTLD_LOCAL`, so identical exported names across package `.so`s never meet. Under webR, package shared objects are Emscripten SIDE_MODULEs sharing one GOT: the **first-loaded** module's definition of an exported name wins for every later module.

## What's still exposed

miniextendr-api is statically linked into *every* miniextendr-based package, and its own `#[unsafe(no_mangle)]` exports carry byte-identical names in each of them:

- `miniextendr-api/src/mx_abi.rs` (3 exports: `mx_query_dispatch` neighborhood — the trait-ABI entry points)
- `miniextendr-api/src/worker.rs` (2)
- `miniextendr-api/src/encoding.rs` (2)
- `miniextendr-api/src/registry.rs` (3, incl. `miniextendr_write_wrappers` / `miniextendr_write_wasm_registry` — native-only wrapper-gen lookups via `getNativeSymbolInfo(..., lib)`, never called on wasm)
- `miniextendr-api/src/backtrace.rs` (1)
- `miniextendr-api/src/lib.rs` (1)
- `miniextendr-api/src/optionals/arrow_impl.rs` (5, feature-gated)
- **most concerning**: `miniextendr-api/src/altrep_impl/builtins.rs` — the builtin-ALTREP registration fns are `#[no_mangle]` AND extern-declared in every package's generated `wasm_registry.rs`. Under GOT interposition, package B's builtin-ALTREP registration can execute package A's copy of the registration fn against **A's module-local statics** (each side module has its own copy of the api crate's statics), registering A's class objects into B's flow.

(`miniextendr_force_link` is package-independent **by design** — value-irrelevant linker anchor referenced by stub.c without configure substitution; interposition of it is harmless. `R_init_<pkg>`/`R_unload_<pkg>` are already package-unique.)

## Why it wasn't fixed in #1273

[...]

PR #1311, opened 2026-07-12
fix(macros): crate-prefix all macro-emitted C symbols for webR cross-package uniqueness

### LINKED ISSUE #1273: webR: C wrapper symbols C_<fn> collide across miniextendr packages in one session (side-module GOT interposition -> cross-package dispatch)

<details>

## Summary
- The `#[miniextendr]` macro emits the C wrapper entry point as `C_<fn_name>` (`c_wrapper_ident()` in `miniextendr-macros/src/miniextendr_fn.rs`, ~line 989). The symbol is `#[unsafe(no_mangle)]` and package-agnostic, so two miniextendr-based packages that export a Rust function with the same name produce byte-identical exported symbol names in their shared objects.
- On native platforms this is harmless: R's `dyn.load` uses `RTLD_LOCAL`, each package's registration table binds to its own DLL's symbols.
- Under webR, package shared objects are Emscripten wasm SIDE_MODULEs. Exported function addresses resolve through a shared GOT across loaded modules, so the first-loaded module's definition wins for every later module that exports the same symbol name. A second package's `R_registerRoutines` table then carries a function pointer into the *first* package's code.
- Observed empirically in the #1259 scaffold CI leg (PR #1272, run https://github.com/A2-ai/miniextendr/actions/runs/29106684936/job/86408792568): the freshly scaffolded package `mxsmoke` ships the template's stock `pub fn add(a: f64, b: f64)`. rpkg (`library(miniextendr)`, loaded first in the tier-3 webR session) exports `pub fn add(left: i32, right: i32)` from `rpkg/src/rust/panic_tests.rs:63`. Calling `mxsmoke::add(2, 3)` in webR returned:

  ```
  ERROR: failed to convert parameter 'left' to i32: wrong type, length, or contains NA: type mismatch: expected INTSXP, got REALSXP
  ```

  The parameter name `left` and the `i32` conversion exist only in rpkg's `add`, so `mxsmoke`'s `.Call(C_add, ...)` dispatched into miniextendr's compiled wrapper. The same call passed in the native sanity check of the same CI job (single package loaded, `RTLD_LOCAL`).
- Ruled out: mxsmoke building against the wrong sources (its `wasm_registry.rs` content-hash `d2af4c21460519df` differs from rpkg's `4f85fb7ca4df3e13`, and `R_init_mxsmoke` loaded fine).

## Impact
- Any two miniextendr-backed packages loaded into one webR session that share any exported Rust function name silently cross-dispatch. The template's stock `add()`/`hello()` make first-party collisions likely for real end users following the getting-started path.
- Affects webR only. Native Linux/macOS/Windows are safe.

## Fix sketch
- Make the C symbol package-unique at macro expansion, e.g. `C_<crate_name>_<fn_name>` using `std::env::var("CARGO_CRATE_NAME")` inside the proc macro (proc macros run in the user crate's env). The R-side `.Call` references, `MX_CALL_DEFS` registration names, and the `useDynLib(.registration = TRUE)` namespace objects all derive from the same naming layer, so the rename must go through `naming.rs` / `c_wrapper_ident()` and the six class generators plus `externalptr_derive.rs` sidecar accessors, `DotCallBuilder`, and the trait-impl path in lock-step. UI/snapshot tests and `patches/templates.patch` regenerate afterwards.
- Alternative considered and rejected: hiding symbols from the wasm export table (visibility flags in the side-module RUSTFLAGS). The registration table takes function addresses, so the GOT entries survive visibility narrowing on Emscripten, and the flags would fork the template Makevars from the native path.

## Notes
- Until fixed, the #1259 CI scaffold leg renames the scaffolded package's stock functions (`add` to `mxsmoke_add`, `hello` to `mxsmoke_hello`) right after `create_miniextendr_package()`, purely to avoid colliding with rpkg's `panic_tests.rs` fixtures in the shared tier-3 session. Drop that rename when this lands.

---

</details>

CGMossa:
After this PR, the only remaining webR symbol-collision surface is miniextendr-api's own `#[no_mangle]` exports (byte-identical in every package — filed as #1310 with an nm-inventory + repro plan), and the reverted #1259 scaffold-leg rename now doubles as the permanent in-CI regression test for the fix; if #1271 (monorepo wasm leg) lands later with its own copy of the old rename workaround, drop it the same way.

<details>
<summary>🤖 AI-written summary and validation details</summary>

## Summary

Implements `plans/2026-07-10-issue-1273-c-symbol-crate-prefix.md` ("Option A"): every `#[miniextendr]`-emitted `#[unsafe(no_mangle)]` symbol is prefixed with the consuming crate's name, read from `CARGO_CRATE_NAME` at macro expansion time (the `miniextendr_init!()` precedent). The invariant **registration name == linker symbol name** is kept everywhere, so the wasm snapshot writer keeps reconstructing its `extern` declarations from `R_CallMethodDef.name` unchanged (`GENERATOR_VERSION` stays 1), and the rename is invisible to R users (`useDynLib(.registration=TRUE)` objects are namespace-local; rpkg's `NAMESPACE`/`man` are byte-identical after regen, as the plan predicted).

| Emission site | New shape |
|---|---|
| bare fns (`miniextendr_fn.rs`) | `C_<crate>_<fn>` |
| impl methods (`miniextendr_impl.rs`) | `C_<crate>_<Type>__<method>`, `C_<crate>_<Type>_<label>_<method>` |
| trait methods + consts (`miniextendr_impl_trait.rs`) | `C_<crate>_<Type>__<Trait>__<member>` (one shared formatter for the former 4 duplicated format strings) |
| sidecar accessors (`externalptr_derive.rs`) | `C_<crate>__mx_rdata_get/set_<Type>_<field>` |
| ALTREP registration (`altrep.rs`) | `__mx_altrep_reg_<crate>_<Ident>` |
| vtable static + shims (`vtable.rs`) | `__VTABLE_<CRATE>_<TRAIT>_FOR_<TYPE>`, `__vtshim_<crate>_...` |

All shapes live in `miniextendr-macros/src/naming.rs` (single source of truth per macros CLAUDE.md); every emission site routes through the helpers. `CARGO_CRATE_NAME` caveat from plan item 1 resolved empirically: cargo does **not** forward it to test-binary processes, so unit/insta-test runtime falls back to normalized `CARGO_PKG_NAME` (`miniextendr_macros`) — stable locally and on CI; a unit test pins the fallback and every shape.

**Beyond the plan's site table**: the stray-format audit (item 8) found two R-side `.Call()` emission sites in `externalptr_derive.rs` (R6 active bindings + S7 properties for sidecar fields) that *reconstructed* the accessor names inline — they now call the same helpers, otherwise crate-prefixing would have silently broken sidecar accessors under R6/S7.

**Deliberately unchanged** (documented in `docs/WEBR.md` + `docs/VISIBILITY.md`): explicit `extern "C-unwind"` fns, `#[export_name]`, `c_symbol = "..."` overrides (author owns webR uniqueness), `R_init_/R_unload_` (already package-unique), `miniextendr_force_link` (value-irrelevant linker anchor).

**MXL303** verdicts are unchanged: the rule compares symbols within one crate, where the prefix is constant, so it keeps matching on the crate-invariant suffix (build.rs has no `CARGO_CRATE_NAME` for the linted crate); rule docs + `lint_code.rs` + lint CLAUDE.md updated to the real shape.

**Workaround revert inventory** (issue #1273 "Notes" + coordinator-expanded scope — all copies found by whole-repo grep, re-checked against origin/main immediately before this PR; PR #1271 has not merged, no third copy exists):
- `.github/workflows/webr.yml` scaffold leg: `sed` rename of `add`/`hello` → `mxsmoke_add`/`mxsmoke_hello`, its two grep asserts, the explanatory comment, and the native-sanity calls.
- `tests/webr-smoke.sh` `phase_scaffold` (added by #1291): same rename block + asserts + comment + native-sanity calls.
- `tests/webr-node-smoke/smoke.mjs`: tier-3 now drives the stock `add()`/`hello()` and asserts the **intentional** collision in both directions — `mxsmoke::add(2, 3) == 5` via mxsmoke's own f64 path AND `miniextendr::add(2L, 3L) == 5` via rpkg's i32 path in the same shared-GOT session.

**Adjacent pre-existing fix** (`rpkg/tests/testthat/test-rng.R`): plan item 10's literal update exposed that the worker-RNG test had *always* skipped — `package_init` locks symbols (`R_forceSymbols`), so `getNativeSymbolInfo(<sym>, "miniextendr")` (PACKAGE-string form) errors even for registered routines (verified identical on main with the old symbol). The probe now uses the DLLInfo form; the test genuinely runs (rng file: 34 pass, 0 skip) and the stale "worker-default" message says worker-thread.

**Cross-package tracked wrappers** regenerated via `just cross-document`: producer.pkg/consumer.pkg now carry `C_producer_pkg_*` / `C_consumer_pkg_*` — an in-tree proof the two packages cannot share wrapper symbols even in principle. The `man/*.Rd` churn in the same regen is roxygen2 8.0.0 catching up on tracked Rd files untouched since #207 (the generated roxygen comment blocks are byte-identical apart from `.Call` symbols; CI's #1218 gate only diffs wrappers.R).

**Docs sweep**: `docs/WEBR.md` gained the crate-prefix gotcha (+ hand-written-symbol responsibility note); shapes updated in CALL_ATTRIBUTION / CLASS_SYSTEMS / R_COERCE / S3_METHODS / TRAIT_AS_R / VISIBILITY / ARCHITECTURE / ENUMS_AND_FACTORS / ERROR_HANDLING / GAPS / MINIEXTENDR_ATTRIBUTE, macros/lint CLAUDE.md, rustdoc, and 5 repo skills.

## Test evidence

- `just test`: all targets green except the 5 pre-existing `derive_dataframe_enum_*` trybuild mismatches (#1239 rust-src artifact, CI authoritative, untouched per plan).
- Macros insta snapshots: 16 rebaselined; verified **mechanically** that every changed content line differs only by inserting the crate prefix (old lines rewritten with `s/C_/C_miniextendr_macros_/g` compare equal to new lines); no other trybuild `.stderr` changed.
- `just devtools-test` (after `rcmdinstall && force-document && rcmdinstall` double-install): `[ FAIL 0 | WARN 0 | SKIP 28 | PASS 7132 ]`.
- Regenerated `wrappers.R` audit: all 52 remaining unprefixed `.Call` symbols verified to be user-owned `extern "C-unwind"` fns (pass-through by design).
- `just cross-install && just cross-test`: producer `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 52 ]`, consumer `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 335 ]`.
- `just minirextendr-test`: `[ FAIL 0 | WARN 0 | SKIP 3 | PASS 746 ]`; `just templates-check` green, no patch churn (templates carry no `C_` literals, as the plan verified).
- `cargo fmt --all` + all three CI clippy legs locally at `-D warnings`: `clippy_default`, `clippy_all` (`full-codegen` + curated list from ci.yml), `clippy_all_s7` — all exit 0.
- The wasm end-to-end proof (plan item 15's done criterion) is this PR's own webR CI leg: tier 3 loads rpkg + mxsmoke with the intentional `add` collision.

## Residual risks (accepted, from the plan)

- Two R packages whose **crates** share a name (deliberate `[lib] name` ≠ package name) still collide; scaffold derives crate from package name and R forbids two same-named packages per session. Optional later guard: `minirextendr_doctor()` warns on crate/package name mismatch.
- Longer names in `getDLLRegisteredRoutines()` output / backtraces are cosmetic.

Follow-up filed: #1310 (api-crate `#[no_mangle]` exports byte-identical across packages; includes the optional nm-based one-symbol-one-module CI assert from plan item 15).

</details>

Fixes #1273


CGMossa:
TODO (author): review framing

<details><summary>AI-written: rebase + review notes</summary>

## Review of the PR itself

Verdict: the approach is sound and well-executed. Specifics checked:

- All 8 emission sites route through the new `naming.rs` helpers (grep-verified: no stray `format!("C_...")` / `format_ident!("C_...")` reconstruction remains in `miniextendr-macros/src/` outside `naming.rs`). The one inline `C_` format left, `match_arg_keys.rs::choices_helper_c_name`, takes the already-prefixed `c_ident` stem, so match-arg helper symbols inherit the crate prefix transitively — no gap.
- The `CARGO_CRATE_NAME` → normalized `CARGO_PKG_NAME` → `"crate"` fallback chain is correct for the three contexts (real expansion / macros-crate unit tests / bare rustc), and the unit test pins both the fallback and every shape.
- MXL303's suffix-only comparison is genuinely prefix-invariant (the prefix is constant within one crate), so verdicts are unchanged; docs updated to the real shape.
- Registration name == linker symbol invariant is preserved everywhere, so the wasm snapshot writer needs no change (`GENERATOR_VERSION` stays 1) — writer unit tests use illustrative literals and are unaffected.
- No unresolved review comments existed on the PR.
- **One real defect found and fixed during the rescue**: the PR's rewritten comment in `tests/webr-smoke.sh` (`# The template's stock fns ...`) sits inside the `phase_scaffold` body, which is a **single-quoted `docker_run '...'` argument** — the apostrophe in "template's" terminated the quote and broke the whole script (`bash -n` fails on the pre-rebase branch tip aa85336e; main passes). Reworded both comments (my new mxmono one included, which had inherited the same wording) and added an explicit no-apostrophes warning marker; `bash -n` now passes.
- One cosmetic nit (pre-existing, not introduced here): the labeled-impl docstring says `C_{crate}_{Type}_{label}__{method}` in one place while the emitted shape has a single underscore before the method (`C_{crate}_{Type}_{label}_{method}`); the shape itself is consistent — doc-only drift, not worth blocking.

## Conflict resolution (rebase onto a88db764)

`git rebase origin/main` — two branch commits replayed cleanly except **one** conflicted file, `tests/webr-node-smoke/smoke.mjs` (hand-written; resolved by hand, no `-X theirs`). Cause: #1308 turned the single scaffold leg into a `SCAFFOLD_PKGS` loop (standalone `mxsmoke` + monorepo `mxmono`). Resolution keeps main's loop structure and applies this PR's semantics inside it: stock `add()`/`hello()` calls, the `miniextendr::add(2L, 3L)` cross-assert, expected string `"5 | Hello, webR! | 5"`.

Beyond the textual conflict, #1308 landed a **new copy of the old rename workaround** in its monorepo leg (exactly the case the PR body pre-announced: "if #1271 lands later with its own copy of the old rename workaround, drop it the same way"). Dropped in the same commit:

- `.github/workflows/webr.yml` monorepo create step: removed the `mxmono_add`/`mxmono_hello` sed + greps; native sanity asserts now call stock `mxmono::add`/`hello`.
- `tests/webr-smoke.sh` monorepo phase: same removal + stock-name asserts.
- `docs/WEBR.md`: the #1308 paragraph describing the mxmono rename workaround rewritten to the intentional-collision framing.

Nice side effect: the monorepo scaffold's crate is `mxmono-core` (≠ package name `mxmono`), so the tier-3 leg now also proves the prefix is the **crate** name (`C_mxmono_core_add`), with a three-way `add` collision (miniextendr / mxsmoke / mxmono) in one shared-GOT session.

Also verified main's new `#[miniextendr]` code (#1313's `condition_demo.rs`, `altrep_condition_tests.rs`, etc.) carries no hard-coded `C_` symbol references — prefixing is fully covered by macro emission; the only R-side literal remains `test-rng.R`'s probe, already updated by this PR.

## Generated artifacts

- Cross-package `producer.pkg`/`consumer.pkg` wrappers.R + man: regenerated via `just cross-install` after the rebase — **byte-identical** to what the PR already committed (clean `git status` after regen), so no extra commit was needed.
- `rpkg` `NAMESPACE`/`man`: byte-identical after `just force-document`, as the plan predicted.
- `rpkg/src/rust/Cargo.lock` configure/install dirt auto-restored by the recipes; tree clean at push.

## Local verification (worktree, R 4.6.0, rebased tree)

- `cargo test -p miniextendr-macros`: 477 unit tests pass, all insta snapshots green; trybuild `ui` shows only the 5 known pre-existing `derive_dataframe_enum_*` span mismatches (#1239 rust-src artifact — the diff is confined to `$RUST/core/src/panic.rs` span rendering; error text and fixture spans identical; CI authoritative, no snapshots regenerated).
- `just test` equivalent: root workspace + consumer.pkg + producer.pkg + rpkg/src/rust cargo tests all pass (same single trybuild exception).
- `just configure && just rcmdinstall && just force-document && just rcmdinstall` (double-install discipline): all four steps exit 0, `DONE (miniextendr)` twice.
- `just cross-install && just cross-test`: producer `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 52 ]`, consumer `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 335 ]`.
- `just devtools-test`: `[ FAIL 0 | WARN 0 | SKIP 27 | PASS 7143 ]`.
- `just templates-check`: green, no patch churn.
- Clippy, all three CI legs at `-D warnings` (`clippy_default`, `clippy_all` with the ci.yml feature list, `clippy_all_s7`): all pass.
- `cargo fmt --all`: no changes.

</details>

Issue #1349, opened 2026-07-13
investigate: propagate spawned-thread panics safely across exported calls


CGMossa:
## Problem

The subprocess fixtures in `rpkg/tests/testthat/test-errors-more.R` document that a panic produced on an ad-hoc spawned thread cannot currently be resumed through an exported `extern "C-unwind"` boundary without a fatal panic-runtime failure on affected platforms. The safe workaround is to translate `JoinError`/panic payloads to an ordinary `Result` instead of calling `resume_unwind`.

## Scope

- Reproduce the failure in the existing subprocess-isolated lane on supported platforms.
- Determine whether the failure is inherent to the Rust/FFI boundary or can be normalized in generated wrappers.
- If fixable, add a supported join helper that converts or transports the panic without unwinding through an unsafe boundary.
- Unskip the focused fixtures only when the behavior is process-safe.
- Keep the documented workaround (`Result`, no `resume_unwind`) until then.

## Context

Tracked from the live limitation in `docs/GAPS.md`; deferred work must have a GitHub issue rather than remain as roadmap prose.

Issue #1348, opened 2026-07-13
feat(ndarray): support string arrays as R character arrays


CGMossa:
## Problem

The `ndarray` integration supports R-native contiguous element types, but `Array<String, D>` / `Array<Option<String>, D>` cannot currently convert directly to R character arrays. Users must flatten or manually assemble a STRSXP and restore `dim`.

## Proposed direction

- Add explicit (non-`RNativeType`) string-array conversion paths for owned ndarray arrays.
- Map `String` to CHARSXP elements and `Option<String>::None` to `NA_STRING`.
- Preserve ndarray shape as R `dim`, with the same column-major/order contract as the existing numeric ndarray conversions.
- Add round-trip tests for matrices and at least one higher-dimensional shape, including empty dimensions, UTF-8, and NA.
- Document allocation/protection discipline: STRSXP elements must be created and installed on the R thread.

## Context

Discovered while reconciling `docs/GAPS.md` and `docs/TYPE_CONVERSIONS.md` against the generated `rust-llm-docs` API corpus. This is the remaining real limitation behind the old “String Matrix/Array Support” entry.

Issue #1347, opened 2026-07-13
deps: remove proc-macro-error2 future-incompatibility warning


CGMossa:
`rust-llm-docs/generate-miniextendr-docs.sh` completes cleanly for project rustdoc, but Cargo reports a dependency future-incompatibility warning from `proc-macro-error2 v2.0.1` (via `tabled_derive v0.11.0`):

```text
warning[E0365]: extern crate `proc_macro` is private and cannot be re-exported
```

The one-line upstream fix is to make the `extern crate proc_macro` item public, but no fixed crate release or upstream commit currently exists. Upstream tracking: https://github.com/GnomedDev/proc-macro-error-2/issues/13

Acceptance criteria:
- consume an upstream release that fixes E0365, or replace/patch the dependency without carrying an unreviewed full crate copy
- regenerate the lockfiles/vendor artifacts through the standard project recipes
- run the rust-llm generator and all affected `just` checks with no dependency future-incompatibility warning
- close only after the warning is absent on a clean build

CGMossa:
Resolved already on current main by #1341 (commit 734222d7): `tabled` now uses `default-features = false, features = ["std"]`, so `tabled_derive` and `proc-macro-error2` are absent. Verified with both `cargo tree -i proc-macro-error2 --all-features` and the benchmark-manifest equivalent; Cargo reports that the package is not present. The clean `just llm-docs` run also emits no future-incompatibility warning.

Issue #1346, opened 2026-07-13
ci(bench): design noise-tolerant microbenchmark regression gating


CGMossa:
The former `docs/FEATURE_BACKLOG.md` parked microbenchmark gating until CI infrastructure existed. CI now exists, but reliable performance gating still needs an explicit noise and baseline strategy before it can block pull requests.

Design and implement a reproducible benchmark regression check.

Acceptance criteria:
- identify a stable runner and representative benchmark subset
- define baseline storage/update policy and noise handling
- choose thresholds that avoid flaky failures
- publish comparison output as a CI artifact or job summary
- document how maintainers intentionally accept a new baseline
- do not make the check required until repeat runs demonstrate acceptable stability

PR #1350, opened 2026-07-13
docs: reconcile manuals with generated Rust API

### LINKED ISSUE #1318: docs(macros): labeled impl-method docstring shows C_{crate}_{Type}_{label}__{method} but emission uses a single underscore before the method
Found during the PR #1311 rescue review (crate-prefix C symbols, #1273). Doc-only; the emitted shape itself is consistent everywhere.

**The drift**: the docstring on `ParsedMethod::c_wrapper_ident` in `miniextendr-macros/src/miniextendr_impl.rs:2123` reads:

```
/// Format: `C_{crate}_{Type}__{method}` or `C_{crate}_{Type}_{label}__{method}`
/// if labeled — crate-prefixed for webR cross-package symbol uniqueness (#1273).
```

i.e. it shows a **double** underscore between the label and the method in the labeled arm.

**The actual emission** (single underscore) is in `miniextendr-macros/src/naming.rs:75`, `impl_method_c_wrapper_ident`:

```rust
quote::format_ident!("C_{prefix}_{type_ident}_{label}_{method_ident}")
```

so a labeled impl method emits `C_<crate>_<Type>_<label>_<method>` — matching the correct shapes documented at `naming.rs:66` and `miniextendr_impl.rs:780` (`C_<crate>_Type_label_method`), the macros CLAUDE.md table, and the PR #1311 body.

**Pre-existing**: before #1311 the same docstring said `C_{Type}_{label}__{method}` while the code emitted `format_ident!("C_{}_{}_{}",  ...)` — the PR carried the doc drift forward verbatim when routing the call through `naming.rs`.

**Fix**: one-line doc edit at `miniextendr_impl.rs:2123` — `C_{crate}_{Type}_{label}_{method}` (single underscore). No codegen, snapshot, or wrapper changes; grep for any other `{label}__{method}` mentions while touching it (none found in source as of c7158ab2 beyond this line).

CGMossa:
## Summary

- reconcile the canonical manuals with the generated Rust API corpus across DataFrame ownership/rooting, serde, conditions, conversions, ALTREP, RNG/rand 0.10, Rayon, feature selection, and extension APIs
- correct the repository architecture throughout the manuals, crate rustdoc, maintainer notes, and hand-authored site pages: generated wrappers execute on R's main thread inside `R_UnwindProtect` by default, while worker dispatch is opt-in; `miniextendr-engine` is only the standalone embedded-R crate
- correct the threading contract throughout the same corpus: the `worker-thread` feature provides infrastructure but does not select worker execution; `#[miniextendr(worker)]`/`worker-default` do; `with_r_thread` routes only from R's main thread or an active miniextendr worker context, not from arbitrary Rust/Rayon threads
- align ExternalPtr and `RAllocator` guidance with that contract: R-facing construction/allocation is valid on the main thread or through the active worker's main-thread rendezvous, while arbitrary threads must do pure Rust computation and return plain Rust data
- remove documentation that treated disabling R's process-global stack check as making off-main R API calls safe; #1352 tracks removal or relocation of the remaining callable escape hatches and unsafe rpkg fixtures
- document the actual R package build: `configure.ac` writes `.cargo/config.toml`, the package static archive is linked into the package shared library, and that library writes wrappers and the wasm registry; those large generated inputs are gitignored but shipped from disk, while `NAMESPACE` and `man/*.Rd` remain tracked
- align ExternalPtr documentation with the implementation: `Any::downcast` is the authoritative type check, R symbols are display/diagnostic metadata, and the erased-pointer path applies only to miniextendr's own `Box<Box<dyn Any>>` representation
- repair stale public examples and references, including custom dots syntax (`name: ...`), `RHash::hash() -> String`, adapter comparison method names, ALTREP registration, feature tables, and labeled C symbols; this resolves #1318 and #1339
- bring source rustdoc into agreement with those contracts and regenerate the tracked LLM-ready corpus
- archive the completed Markdown feature backlog; keep only current limitations in `GAPS.md` and track remaining work in #1345, #1346, #1348, #1349, and #1352
- make LLM-doc generation checkout-independent, add CI coverage for the canonical docs index, and repair broken site links/anchors
- keep `with_rng`'s main-thread precondition active in release builds
- repair `just lint` so it reads the R package generated Cargo configuration and restores its transient lockfile rewrite
- add `just issues-refresh` and require a refreshed, end-to-end issue-index read plus relevant body search before filing follow-up issues

The remaining obsolete handwritten `miniextendr-cli init` scaffold is tracked separately in #1351 so it can be replaced with the canonical package templates rather than partially patched here.

Dependency audit note: #1329 was the canonical `proc-macro-error2` tracker. It and accidental duplicate #1347 are closed after current `main` confirmed #1341 had already removed `tabled_derive -> proc-macro-error2` from the active graph.

## Root cause and impact

The documentation had accumulated several descriptions of earlier implementations: worker execution as the default, code generation owned by the engine crate, a temporary cdylib/document-binary build, committed wrapper files, symbol-based ExternalPtr compatibility, and stack-check suppression as permission to call R from secondary threads. Those descriptions contradicted the current source, generated API reference, and R's documented main-thread requirement. This refresh makes the user manuals, maintainer guidance, shipped skills, site, source rustdoc, and generated LLM corpus describe the same runtime and build system.

## Verification

- `just fmt`
- `just check`
- `just doc-check`
- `just lint`
- `just clippy`
- all three CI `cargo clippy ... -D warnings` configurations (default, full R6, full S7)
- standalone `cargo-revendor` clippy with `-D warnings`
- `just llm-docs`
- `just llm-docs-check` (six renderer tests plus clean regeneration)
- `just site-check` (84 generated pages, no warnings)
- `just issues-refresh` (83 open bodies, 490 closed summaries)
- `shellcheck scripts/refresh-issues-cache.sh`
- `just agents-md-check`
- docs index coverage check
- `git diff --check`

`just test` passed every non-UI target it reached. Five of 143 macro UI comparisons differ only in the documented local Rust 1.97 const-panic source rendering; no trybuild snapshots were updated.


CGMossa:

<details>

## Summary

Reviewed the PR and verified the load-bearing doc changes against the actual code (not just the prose): the headline `String`/`Vec<String>` NA-handling change (docs now say NA -> `""` lossy) matches `from_r/strings.rs`, which returns `Ok("")` for `NA_string` on both scalar and vector paths. The previous module-level doc in `strings.rs` was stale (it claimed `String` "rejects NA with `SexpNaError`", which the code never did), so this is a correct reconciliation, not a regression. `Vec<bool>` -> error, and the `AS_COERCE.md` -> `R_COERCE.md` relinking both check out too.

Pushed one follow-up commit (`dbaea1d9`) fixing two of the items I flagged:

- **CI docs-index check was matching a bare filename substring.** A doc whose name is a substring of another doc's link counted as "indexed" without a real entry. This masked a genuine gap: `PACKAGES.md` (the Package Map) had no index entry, but passed because it is a substring of the linked `NATIVE_R_PACKAGES.md` (same class as `COERCE.md` inside `R_COERCE.md`). Tightened the check to match `](name)` / `](name#anchor)` link targets, added `set -euo pipefail`, and added the missing `PACKAGES.md` entry to `docs/README.md`. Re-ran the tightened check: all top-level docs now resolve to real links.
- **`refresh-issues-cache.sh` truncated silently at `--limit 1000`.** Closed issues are ~488 today but climbing; hitting the page cap would silently cache a partial ledger. Added a warning on cap-hit for both open and closed fetches and parameterized `LIMIT`. `shellcheck` stays clean.

## Notes

- Two flagged items were reviewed and left as-is on purpose: (1) `with_rng`'s `debug_assert!` -> `assert!` is the correct #1096 hardening (it guards a safe public API against off-main-thread RNG UB); it does assume `R_MAIN_THREAD_ID` is always registered before any `with_rng` call, which holds for normal package init. (2) The `just lint` trap's `git restore` would discard hand edits to the generated `rpkg/src/rust/Cargo.lock`, but that file is generated/maintainer-only, so the blast radius is acceptable.
- The large `rust-llm-docs/generated/` churn is line-number-only drift from regeneration, as expected.

---

</details>

PR #1389, opened 2026-08-12
ci(rpkg): gate full-feature clippy warnings

### LINKED ISSUE #1379: ci: no clippy -D warnings leg covers rpkg/src/rust — fixture warnings accumulate silently
**TODO(user): replace this line with your own framing, in your own voice.**

<details>

## Summary
- No CI job runs `cargo clippy` against `rpkg/src/rust` with `-D warnings`. The crate is a standalone workspace, so the root `--workspace` legs (`clippy_default` / `clippy_all` / `clippy_all_s7`) never reach it; the `clippy_all` step comment even documents that its curated feature list exists to *approximate* what rpkg enables. The only CI compiles of the crate are R-install builds, where warnings scroll by ungated.
- Consequence: fixture-crate warnings accumulate silently. #1292's two warnings sat unnoticed since `5492c1bb` (#1122), and between #1292 being filed (07-11) and fixed (07-17), two more `unused_imports` appeared (top-level `ColumnarFrame` imports orphaned by #1355's migration to the trait verbs, used only under `#[cfg(test)]`).
- Precedent: #785 closed the identical gap for rustfmt on this crate.

## Sketch
- Add a step to the clippy CI job: `just configure` (writes `rpkg/src/rust/.cargo/config.toml` with the monorepo `[patch]` overrides), then run clippy **with cwd inside `rpkg/src/rust`** (cargo resolves `.cargo/config.toml` from cwd; `--manifest-path` from the root misses the patches and produces phantom git-dep resolution errors):

  ```bash
  just configure
  cd rpkg/src/rust && cargo clippy --all-targets --features full -- -D warnings
  ```

## Notes
- Feature choice is the main trade-off: `--features full` includes `arrow` + `datafusion` (heavy compile; sccache should keep it tolerable, and it is exactly the config that reproduces #1292). A curated lighter list would skip the arrow/datafusion-gated fixtures.
- Non-`full` selectors (`nonapi`, `connections`, `satellite`, `r6-default`/`s7-default`, `worker-default`) stay uncovered either way, the same limitation `clippy_all` already accepts.
- `--locked` needs care: the committed `rpkg/src/rust/Cargo.lock` is tarball-shape; `just check` works around this by restoring the lock afterwards. Simplest is to omit `--locked` for this leg or reuse the `just check` pattern.

---

</details>

CGMossa:
## Summary
- add a dedicated full-feature, -D warnings clippy step for the standalone rpkg Rust workspace
- extract a reusable clippy-rpkg maintainer recipe with explicit local patches and lockfile restoration
- remove the duplicate test imports the new gate exposed on main and document that failure in reviews/

Closes #1379

## Verification
- just fmt-check
- just check
- just test
- just clippy -- -D warnings
- just clippy-rpkg --features full -- -D warnings
- CI clippy_default, clippy_all, and clippy_all_s7 commands with -D warnings
- actionlint on ci.yml after suppressing the eight diagnostics already present on origin/main

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -424,6 +424,16 @@ jobs:
           -- -D warnings 2>&1 | tee "$RUNNER_TEMP/clippy_all_s7.log"
         shell: bash
 
+      # rpkg/src/rust is a standalone workspace, so none of the root
+      # `--workspace` invocations above reach its fixture crate. The dedicated
+      # recipe supplies local framework patches without requiring configure or
+      # a real R installation, and restores the tarball-shaped lockfile.
+      - name: Clippy (rpkg, full features)
+        id: clippy_rpkg
+        continue-on-error: true
+        run: just clippy-rpkg --features full -- -D warnings 2>&1 | tee "$RUNNER_TEMP/clippy_rpkg.log"
+        shell: bash
+
       # cargo-revendor is a standalone workspace excluded from the root
       # workspace, so the `--workspace` clippy steps above never reach it
       # (#778). It has no R linkage, so it lints without configure/R setup.
@@ -453,6 +463,7 @@ jobs:
           clippy_default="${{ steps.clippy_default.outcome }}"
           clippy_all="${{ steps.clippy_all.outcome }}"
           clippy_all_s7="${{ steps.clippy_all_s7.outcome }}"
+          clippy_rpkg="${{ steps.clippy_rpkg.outcome }}"
           clippy_revendor="${{ steps.clippy_revendor.outcome }}"
           doc_revendor="${{ steps.doc_revendor.outcome }}"
 
@@ -461,6 +472,7 @@ jobs:
           echo "clippy_default:   $clippy_default"
           echo "clippy_all:       $clippy_all"
           echo "clippy_all_s7:    $clippy_all_s7"
+          echo "clippy_rpkg:      $clippy_rpkg"
           echo "clippy_revendor:  $clippy_revendor"
           echo "doc_revendor:     $doc_revendor"
           echo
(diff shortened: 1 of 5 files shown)


CGMossa:

<details>

## Verdict
LGTM. Reviewed the full diff; no correctness issues found.

## Summary
- The new `clippy-rpkg` recipe correctly supplies the framework `[patch]` overrides via `--config`, so it works both with and without a configure-generated `.cargo/config.toml`, and the EXIT trap restores the tarball-shaped lockfile on failure paths too.
- rpkg's `full` feature excludes all six `*-default` codegen selectors, so no r6/s7 mutex conflict; it also includes `macro-coverage` together with `worker-thread`, which closes the previously known gap that no CI job compiled `macro-coverage` (it fails to compile alone).
- The new step passed on this PR's own Rust lint job, so the gate is proven live.
- The duplicate-import fix plus `reviews/2026-08-12-rpkg-clippy-duplicate-imports.md` post-mortem match repo conventions.
- Minor, not blocking: `--all-targets` is a superset of the old `--benches --tests --examples` (adds lib/bins), which is strictly more coverage.

## Notes
- The Sync Checks failure on this PR is unrelated: every open PR fails it due to roxygen2 8.0.0 vs 8.1.0 NAMESPACE formatting drift, fixed in #1424. Rebase onto main after #1424 lands and this PR should go green.

---

</details>

CGMossa (commented):

<details>

## Summary
- Verified on the merged head (main 3a0746a4 merged in, clean): `just test-rpkg` runs the three `#[cfg(test)]` modules (45 passed, 1 ignored, 26 s warm) and the EXIT trap leaves `rpkg/src/rust/Cargo.lock` untouched.
- The leg belongs in `rust-tests` as the PR says: `miniextendr-api/build.rs` links libR unconditionally, so the rpkg test binary needs the R setup that job already has; `rust-lint` has none.
- The `changes` filter already covers `**/*.rs` and `justfile`, so edits under `rpkg/src/rust` trigger the job.
- Closes #1474 as stated; merging on green CI.

---

</details>

PR #1390, opened 2026-08-12
ci: lint GitHub Actions workflows

CGMossa:
## Summary
- add an always-on Workflow lint job using the official actionlint 1.7.12 image and include it in the CI Success fan-in
- fix all ten diagnostics actionlint reports across ci.yml and pages.yml
- record the missing enforcement and its resolution in findings/

The fixes quote GitHub environment-file paths and the macOS user argument, remove stale needs expressions from shell comments, and preserve backslashes in the Pages permission loop.

## Verification
- actionlint 1.7.12 across all workflows
- docker run with rhysd/actionlint:1.7.12 against the worktree
- just fmt-check
- git diff --check

PR #1394, opened 2026-08-12
fix(api): honor per-string R encodings

CGMossa:
## Summary

- classify each CHARSXP with R public encoding APIs instead of inferring its bytes from the process locale
- retain zero-copy borrowing for UTF-8/ASCII, translate Latin-1 text, own translated Cow values, reject bytes strings, and validate before constructing Rust str
- add real R regressions across String, borrowed str, scalar/vector Cow, bytes, and malformed UTF-8
- correct stale encoding and REncodingInfo documentation; record the audit finding and failure review

## Reproduction

In an R 4.6 UTF-8 session, an explicitly Latin-1 tagged facade string failed conv_string_arg in debug builds. The release path used from_utf8_unchecked on the same non-UTF-8 bytes.

## Verification

- just fmt-check
- exact CI clippy default, full-codegen/R6, and full-codegen/S7 configurations with -D warnings
- just doc-check across all manifests
- just rcmdinstall; installed package loads successfully
- focused testthat encoding suite: 29 passes, 0 failures
- just check, just clippy, and just test pass their root and cross-package legs; their standalone rpkg test leg reaches the pre-existing duplicate ColumnarFrame imports already isolated in PR #1389

This PR does not merge itself.


CGMossa:

<details>

## Verdict
LGTM — this is a real soundness fix and the design is right. Two non-blocking notes below.

## Summary
- Confirmed the bug class: a UTF-8 process locale does not constrain per-CHARSXP tags, so the old `from_utf8_unchecked` path could construct an invalid `str` in release builds from an explicitly Latin-1-tagged string. The fix dispatches on `Rf_charIsUTF8` / `Rf_getCharCE` (bindings already present in sys.rs with checked/unchecked variants).
- The `charsxp_to_str` (translating, valid for the enclosing `.Call`) vs `charsxp_to_borrowed_str` (strict, panics on non-UTF-8) split is the key correct decision: `&str` *arguments* are `.Call`-scoped so the `Rf_translateCharUTF8` R_alloc buffer is fine, while `StrVec` / `ProtectedStrVec` views can outlive that scope and correctly refuse to hand out translated buffers, steering callers to `get_cow`.
- The mandatory `from_utf8` validation on the fast path is justified, not optional: your own test shows R happily carries a `CE_UTF8` tag on invalid bytes (`rawToChar(as.raw(0xff))` + `Encoding<-"UTF-8"`), so the tag alone can never re-enable `from_utf8_unchecked`.
- Verified `str_borrow_len` counts chars (not bytes), so the `6L` expectation for translated "façade" is correct.
- Test coverage at the real R boundary (owned String, borrowed str, scalar/vector Cow borrowed-vs-owned, bytes rejection, lying UTF-8 tag) is exactly the matrix that was missing.

## Non-blocking notes
- `Cow<'static, str>::try_from_sexp_unchecked` now just calls the checked `try_from_sexp`, which re-enables the R-thread debug asserts the `_unchecked` contract exists to skip. Harmless today (string conversion happens on the R thread), but a `charsxp_to_cow_unchecked` would keep the checked/unchecked symmetry the rest of `from_r` maintains.
- The fast path gains an O(n) UTF-8 validation where there was none. It's required for soundness, but if any recorded string-conversion benchmarks (scaffolding perf evidence) are re-run, expect a small regression there and treat it as the new baseline.
- Same `findings/` vs `reviews/` convention question as raised on #1390/#1392 (this PR adds one file to each).

---

</details>

PR #1395, opened 2026-08-12
fix(api): protect Jiff timezone metadata

CGMossa:
## Summary
- protect the dynamically created Jiff timezone CHARSXP before any later R allocation
- strengthen the GC-stress fixture so it asserts the exact tzone attribute, not only numeric ALTREP elements
- dogfood the public zoned-vector constructor from R, including timezone preservation and mixed-zone rejection
- update cached-SEXP documentation and record the audit finding plus failed-oracle review

## Audit evidence
R extension guidance requires objects returned from the R API to be protected before any subsequent allocating call. R keeps cached CHARSXPs weakly, so the previous charsxp-then-vector allocation order violated that contract even though local torture runs did not deterministically corrupt the attribute. The existing GC fixture could not detect this because it never inspected tzone.

## Verification
- just rcmdinstall
- focused test-jiff.R: 127 assertions, 0 failures
- 50 iterations of gc_stress_jiff_zoned_vec under gctorture(TRUE)
- just fmt-check
- just doc-check
- all three CI clippy configurations with -D warnings
- just test: root, ndarray, and both cross-package legs pass; standalone rpkg is blocked by the duplicate ColumnarFrame imports already isolated in #1389

This PR is intentionally left unmerged.


CGMossa:

<details>

## Verdict
LGTM. The protection bug is real and the fix is correctly ordered.

## Summary
- The bug is the textbook WRE violation: `SEXP::charsxp(iana)` returns a CHARSXP whose only root is the weak CHARSXP cache, and the very next call (`Rf_allocVector`) can GC. Interning is not rooting; the finding's framing is exactly right, including the honest note that gctorture didn't deterministically reproduce it — the API contract, not the repro, is the authority.
- Verified the drop-order soundness since `OwnedProtect::drop` is a bare `Rf_unprotect(1)` (position-blind LIFO pop): the two guards are declared charsxp-first, tzone-second, and Rust drops locals in reverse declaration order, so the pops match the protect stack. Correct — but this pattern is fragile under refactoring; if anyone reorders those `let`s the stack unbalances silently. A one-line comment noting the required declaration order (or a `ProtectScope`) would make it tamper-proof.
- The fixture-oracle hardening is the most valuable part: asserting the actual `tzone` attribute instead of only ALTREP element reads converts the fixture from "exercises the path" to "detects the corruption". Good catch that the old oracle covered the surrounding object but not the vulnerable metadata.
- Doc updates check out against source: the cfg gates in `cached_class.rs` already read `any(time, jiff, arrow)` / `any(time, jiff)`, so CACHED_SEXPS.md was stale and this aligns it.
- New testthat coverage (constructor round-trip incl. tzone identity, mixed-zone rejection) fills a genuine gap.

## Notes
- Same `findings/` vs `reviews/` directory-convention question as on #1390/#1392/#1394.
- Sync Checks failure is the repo-wide roxygen2 8.1.0 drift (#1424); rebase after it lands.

---

</details>

CGMossa (commented):

<details>

## Verdict: approve

## Summary
- `set_posixct_tz` created the tzone CHARSXP and then allocated its STRSXP container before installing it; R's CHARSXP cache is weak, so the fresh string was collectible across that allocation. It now gets its own `OwnedProtect` first, and the two guards drop in LIFO order.
- The GC fixture only read ALTREP elements before and could not notice a reused tzone; it now asserts the exact `tzone` attribute. Public dogfood covers timezone preservation and mixed-zone rejection.
- Verified on a combined build of the five rooting PRs: `test-jiff.R` 127 pass, 0 fail.

## Notes
- `docs/CACHED_SEXPS.md` feature gates and names vector corrected. Merged `origin/main` and pushed. Body has no attribution placeholder or details block, and no linked issue. Full notes: `reviews/1395-review.md`.

---

</details>

PR #1396, opened 2026-08-12
test(api): dogfood ExternalPtr worker transfer

CGMossa:
## Summary
- add real worker-thread fixtures that receive an R-owned ExternalPtr, read it on the worker, and return the same handle with R identity preserved
- exercise ExternalPtr::into_inner on the worker and verify the original R external pointer is cleared exactly once
- gate the fixtures with worker-thread so minimal-feature rpkg builds remain valid
- replace the stale fixture comment claiming ExternalPtr is !Send
- record the audit finding and the narrow-check setup retries

## Why
The existing tests named "ExternalPtr from worker context" only computed plain values on the worker and constructed every pointer later on the R main thread. They never tested the safety-critical Send contract documented by ExternalPtr itself.

## Verification
- minimal rpkg library check and clippy with -D warnings
- worker-thread rpkg library check and clippy with -D warnings
- just rcmdinstall
- focused test-externalptr.R: 30 assertions, 0 failures
- 25 repeated read -> identity -> into_inner cycles under gctorture(TRUE)
- just force-document; no tracked generated drift
- just fmt-check
- just doc-check
- all three exact CI clippy configurations with -D warnings
- just check / just test: root, ndarray, and both cross-package legs pass; standalone rpkg is blocked by the duplicate ColumnarFrame imports already isolated in #1389

The local documentation pass also reproduces the shared R6-field roxygen warnings seen by current Sync Checks; this branch does not mix that unrelated correction in.

This PR is intentionally left unmerged.


CGMossa:

<details>

## Verdict
LGTM. The coverage gap was real and these are the right three fixtures.

## Summary
- The finding is accurate: the pre-existing "worker context" tests computed plain values on the worker and built every EXTPTRSXP on the main thread afterwards, so the documented `ExternalPtr<T>: Send` contract (the safety-critical part) had zero coverage. The three new fixtures hit exactly the distinct risk surfaces: read-after-transfer, identity-preserving round-trip, and worker-side `into_inner` (checked R API routing + clear-exactly-once).
- The `into_inner` test's follow-up assertion (`external pointer is null` on reuse) is the right oracle for single-clear semantics.
- The stale `!Send` comment removal in worker_tests.rs matches the actual impl (Send when `T: Send`).
- `#[cfg(feature = "worker-thread")]` on the fixtures plus `#[miniextendr(worker, noexport)]` is correct, and CI's r-tests ran these green.

## Non-blocking note
- The testthat tests call the worker fixtures unconditionally, while the fixtures only exist under `worker-thread`. That's safe today because rpkg/tools/detect-features.R always includes `worker-thread` in the detected base, but a deliberately minimal-feature install would fail the suite with "could not find function" instead of skipping. A `skip_if_not(...exists...)` guard would make the suite feature-honest; fine to defer.
- Same `findings/` vs `reviews/` convention question as on the sibling PRs.

---

</details>

PR #1397, opened 2026-08-12
fix(connections): preserve long formatted writes

CGMossa:
## Summary

- preserve the dynamically growing `dummy_vfprintf` callback installed by R
- remove the unused Rust `va_list` override and fixed 10,000-byte formatter
- add a public R regression for 12,000-byte formatted writes
- record the finding/review and correct connection lifecycle/formatting docs

## Runtime evidence

Before this change, `writeLines()` through `memory_connection()` silently returned 9,999 of 12,000 bytes. Against the rebuilt branch, the same public API roundtrip returns all 12,000 bytes exactly.

## Verification

- `cargo lcheck -p miniextendr-api --lib --features connections`
- affected Rust connection tests: 10 passed, 1 intentionally ignored embedded-R placeholder
- `rpkg/tests/testthat/test-connections.R`: 34 expectations passed
- direct 12,000-byte public R roundtrip
- `just fmt-check`
- `just doc-check`
- all three exact CI root clippy gates, including `full-codegen-s7`, with `-D warnings`

`just check`, `just test`, and `just clippy` pass their root and cross-package legs, then stop in the standalone `rpkg` workspace on the duplicate `ColumnarFrame` imports already isolated in #1389. This PR deliberately does not duplicate that unrelated fix.


CGMossa:

<details>

## Verdict
LGTM — the best kind of fix: strictly less code, strictly more correct.

## Summary
- The old trampoline's real sin wasn't the 10,000-byte cap, it was returning `vsnprintf`'s full length after forwarding only the truncated prefix — R had no way to detect the short write, hence silent data loss. The regression test at the public API level (12,000 bytes through `writeLines` → `seek` → `readLines`) pins exactly that.
- Preserving R's `dummy_vfprintf` is the right call: R owns the platform `va_list`, grows its buffer, handles text-mode encoding, and forwards complete bytes to the installed Rust `write` trampoline — so `write` remains the single implementation surface.
- Removing `RConnectionImpl::vfprintf` outright (rather than keeping a dead default) matches the no-backwards-compat policy; the finding confirms no in-repo implementor overrode it, and it was an unportable ABI surface (`*mut c_void` va_list) that invited unsound overrides.
- Docs updated coherently in both places that described the old fallback (callback table + Limitations), and the lifecycle description now matches the builder's actual open() behavior.

## Notes
- This changes behavior for any hypothetical downstream implementor of `vfprintf` (their override silently stops being called → now a compile error since the method is gone — which is the good failure mode).
- Same `findings/` vs `reviews/` convention question as the sibling PRs; Sync Checks failure is the repo-wide roxygen2 drift (#1424).

---

</details>

PR #1398, opened 2026-08-12
fix(expression): root fluent call builder inputs

CGMossa:
## Summary

- root the `RCall` callable and every positional/named argument for the builder lifetime
- cover pool initialization and growth with temporary stack roots
- make `RCall` non-`Send` through its R-owned root pool
- replace compile-only Vec tests with public R dogfood under `gctorture(TRUE)`
- update the existing expression GC fixture to use fresh inline string arguments
- document the corrected ownership contract and record the finding/review

## Runtime evidence

Before this change, the documented fluent form `RCall::new("c").arg(SEXP::scalar_integer(...))` returned this on its first invocation under `gctorture(TRUE)`:

```text
[1] 2 2 8 8 8 8 8 8
```

The expected value was `1:8`; earlier raw SEXP arguments had been collected and reused. Against this branch, both the eight-integer builder (which crosses the root-pool growth boundary) and the inline-string/named-argument fixture pass 20/20 iterations.

## Verification

- `cargo lcheck -p miniextendr-api --lib`
- affected Rust expression tests
- `test-expression.R` in non-CRAN stress mode
- 20 inline-integer `RCall` iterations under full gctorture
- 20 inline-string/named-argument `RCall` iterations under full gctorture
- downstream `test-connections.R` / progress tests
- downstream `test-s4-helpers.R`
- `just force-document` (no tracked drift; reproduced the shared R6 `@field`/`@backref` warnings)
- `just fmt-check`
- `just doc-check`
- all three exact CI root clippy gates, including `full-codegen-s7`, with `-D warnings`

`just check`, `just test`, and `just clippy` pass their root and cross-package legs, then stop in the standalone `rpkg` workspace on the duplicate `ColumnarFrame` imports already isolated in #1389. This PR deliberately does not duplicate that unrelated fix.


CGMossa:
This is worthwhile to do, although I don't think the pool is necessary. A linkedlist is created, if that gets protected, all of its elements also get protection. That strategy should be explored here, as it may be more performant.

CGMossa:

<details>

## Verdict
LGTM as a soundness fix — the bug is real, the rooting is correct, and the runtime evidence is conclusive. On the linked-list comment: agreed it's the better end-state; details below.

## Review of the fix as written
- The UAF is genuine: `.arg()` stored raw SEXPs in a Rust `Vec` while subsequent inline allocations could GC them, and the docs actively advertised the unsafe ergonomic form. The `[1] 2 2 8 8 8 8 8 8` evidence is classic collected-and-reused output.
- Verified the protect-stack discipline: `ProtectPool` anchors its backing VECSXP via `R_PreserveObject` and its internal `Rf_protect`/`Rf_unprotect` pairs are balanced per call, so the temporary `OwnedProtect` guards in `from_callable`/`arg` pop cleanly. The temp root over pool init/growth (so the incoming value survives the pool's own allocation) is exactly right.
- Making `RCall` `!Send` via the pool is a correct side effect, and rooting the resolved closure in `from_r_function` (instead of assuming namespace reachability) closes the same hole for the `from_sexp`-backed path.
- Deleting the null-pointer compile-only unit tests was necessary (construction now touches the R runtime) and the R-side gctorture dogfood is a strictly stronger replacement. Converting the old fixture to inline args makes it test the new contract rather than work around the old one.

## On the linked-list suggestion (from the PR comment)
Feasible and likely both faster and simpler, with two subtleties to carry into the follow-up:
1. The head of the pairlist still needs a **lasting movable root** — `RCall` is a movable value, so a protect-stack entry can't be tied to its lifetime; the head wants `R_PreserveObject` (or a 1-slot pool). Net cost becomes one preserve per builder vs. one preserve plus VECSXP growth today.
2. The **per-insert temporary root is still required**: `Rf_cons(value, R_NilValue)` can GC during its own allocation before storing CAR, so a fresh inline `value` must be stack-rooted across the cons, same as the pool's insert does now.
The win is real though: args go straight into their final CONSXP cells (with `SET_TAG` for named args), `build()` collapses to `Rf_lcons(fun, head)` instead of a second cons pass, and the VECSXP disappears. Repeated `build()`/`eval()` reuse the same chain, which is standard LANGSXP practice. Suggest landing this PR as the soundness fix and filing the linked-list rework as a tracked issue per the concessions-to-issues convention.

## Notes
- Same `findings/` vs `reviews/` convention question as the sibling PRs; Sync Checks failure is the repo-wide roxygen2 drift (#1424).

---

</details>

CGMossa (commented):

<details>

## Verdict: approve

## Summary
- `RCall` now owns a `ProtectPool` rooting the callable and every positional or named argument for the builder's lifetime, with a temporary `OwnedProtect` around each `insert` because pool growth allocates. The pool anchors its backing VECSXP with `R_PreserveObject`, so `RCall`'s drop order cannot unbalance the PROTECT stack, and it makes `RCall` `!Send`. `from_str` roots the resolved closure too.
- Real oracle: the documented `RCall::new("c").arg(SEXP::scalar_integer(..))` form returned `2 2 8 8 8 8 8 8` under gctorture before; the new fixture crosses the pool-growth boundary with eight inline arguments and the test asserts `1:8` for 20 iterations.
- Verified on a combined build of the five rooting PRs with `NOT_CRAN=true`: `test-expression.R` 15 pass, 0 fail; the updated `gc_stress_expression_call` fixture passes five rounds under gctorture.

## Notes
- The two compile-only unit tests that built `RCall` from null SEXPs had to go (no pool without R); the R dogfood replaces them. `docs/EXPRESSION_EVAL.md` updated.
- Merged `origin/main` and pushed. Body has no attribution placeholder or details block, and no linked issue. Full notes: `reviews/1398-review.md`.

---

</details>

PR #1400, opened 2026-08-12
docs: refresh LLM-ready Rust API corpus

CGMossa:
## Summary

- refresh the tracked Rust API corpus from current `origin/main`
- capture public serde, scaffold, and macro API changes merged after the last refresh
- restore a clean generated baseline for subsequent scoped API changes
- record the corpus-drift finding and review

## Evidence

The previous corpus predated the public surfaces merged in #1371, #1372, and #1376. A clean generation changed seven tracked reports. Two consecutive runs produced identical hashes for the API digest, API impl inventory, and macros digest.

## Verification

- `CARGO_TARGET_DIR=target just llm-docs-check`
- six renderer unit tests passed
- post-generation tracked corpus diff was empty
- generated reports contain no worktree or developer absolute paths

PR #1399, opened 2026-08-12
fix(docs): honor Cargo target directory in LLM corpus generation

CGMossa:
## Summary

- resolve Cargo target directories from metadata in both workspaces
- render from the JSON paths Cargo actually used
- keep committed report source labels repository-relative and machine-independent
- record the reproduced failure and code-backed finding

## Verification

- `bash -n rust-llm-docs/generate-miniextendr-docs.sh`
- `python3 -m unittest discover -s rust-llm-docs -p "test_*.py"`
- `just llm-docs` with global `[build] target-dir = "/Users/elea/.cargo-target"`
- verified generated inventory headers retain `target/doc/...` and `cargo-revendor/target/doc/...` labels

PR #1401, opened 2026-08-12
fix(docs): make Rust API corpus ordering deterministic

CGMossa:
## Summary

- replace non-unique report sort keys with semantic total orderings
- stabilize re-exports, trait sections, same-span impl rows, clusters, and manual-vs-macro shapes
- add reversed-JSON-index regression tests for all three affected renderers
- regenerate the corpus once into canonical order
- record the finding and both failure reviews

## Evidence

A factor-only API removal produced roughly three thousand unrelated generated-line changes in a second worktree. The renderers were preserving rustdoc JSON object order whenever their sort keys tied.

## Verification

- `CARGO_TARGET_DIR=target just llm-docs-check`
- all nine renderer tests pass
- forward and reversed fixture indexes render byte-identically
- two consecutive full generations produced identical hashes

## Stack

Based on #1400 so this PR contains only deterministic ordering and its canonical reorder. Do not merge this before #1400.


CGMossa:
These changes should be upstreamed to the rust-llm-docs/ repository as well.

CGMossa:

<details>

## Verdict
LGTM. This fixes the long-standing "impl-inventory files reorder non-deterministically" problem at the root instead of working around it at rebase time.

## Summary
- The sort-key changes are genuine total orderings: every previously tie-vulnerable key (`-len` counts, spans) gained deterministic alphabetical tiebreaks, and `trait_order` being computed once and shared between the summary and detail loops removes a second, subtler inconsistency.
- Sorting the `fors` list inside multi-impl rows changes rendered order from JSON insertion order to alphabetical — a deliberate, correct normalization for a report.
- The reversed-JSON-index regression tests are the right oracle: byte-identical output from forward and reversed fixture indexes is exactly the property that was missing.
- The canonical one-time reorder of 13 generated files is unavoidable churn and correctly isolated from the #1400 baseline refresh per the stated stack order.

## On the upstreaming comment
Agreed these belong in A2-ai/rust-llm-docs. Note there are already three open upstream PRs from this repo's divergent copy (overhaul, API.md + trait-methods, audit scripts), so the determinism fix should go up as a fourth stacked change once this lands — worth a tracked issue so it doesn't get lost, per the concessions-to-issues convention.

## Merge-order caution
- Stack: #1399 → #1400 → this. After this lands, #1421 (the other session's digest refresh) regenerates the same files without these ordering fixes — it should be rebased and regenerated on top of this, or closed as superseded.

---

</details>

PR #1404, opened 2026-08-13
fix(macros): attach R6 active docs to classes

CGMossa:
## Summary

- emit R6 active-binding `@field` tags on the class documentation block
- leave dynamic `$set("active", ...)` calls without adjacent roxygen blocks, avoiding roxygen2 8.0 method misclassification
- preserve internal binding docs, explicit class-level field overrides, and `prop = "..."` names
- update the dogfood fixture wording and record the code-first finding in `findings/` and `reviews/`

## Evidence

roxygen2 8.0.0 records every dynamic `$set()` target as an R6 method without inspecting whether it is `public` or `active`. The old generated shape therefore produced fourteen `@field` / generated `@backref` warnings for seven active bindings during every forced documentation run.

After this change, `just force-document` emits zero R6 ownership warnings and produces no tracked Rd drift; existing Active bindings sections are unchanged.

## Verification

- `just fmt`
- 6 focused macro/codegen regression tests
- `just rcmdinstall` + `just force-document`: zero R6 matching warnings, no NAMESPACE/man drift
- `just wrappers-sync-check`
- `test-r6-noexport-field.R`: 5 expectations
- `test-class-systems.R`: 124 expectations, 2 intentional empty-test skips
- exact CI `cargo clippy` default, full-codegen/R6, and full-codegen/S7 commands with `-D warnings`
- `just check`: root and both cross-package workspaces pass; final standalone rpkg leg reaches the existing duplicate-`ColumnarFrame` base blocker isolated in #1389
- `just test`: root, ndarray, and both cross-package suites pass; final standalone rpkg leg reaches the same #1389 blocker

No merge requested.


CGMossa:

<details>

## Verdict
LGTM. Right mechanism, right placement, and it clears a warning class the project explicitly treats as bugs.

## Review
- The root-cause analysis matches what roxygen2 actually does: the dynamic `$set()` parser classifies any adjacent block as *method* docs regardless of the `"active"` first argument, so class-block `@field` tags (resolved against the runtime class's active bindings) are the only placement that both documents and silences correctly.
- `field_tag_documents` handling comma-separated `@field a,b` names and explicit class-level overrides prevents double-documentation; `(internal)` for noexport/internal bindings follows the repo's documented convention (never `@field name NULL`); the "Active binding." fallback keeps every binding with a real description rather than a silencer.
- Insert-before-`@export` placement and the splice keep the class block shape stable; snapshot + 6 regression tests cover it.
- I checked the current Sync Checks logs on other PRs: the same fourteen `@field`/`@backref` warnings fire under CI's roxygen2 **8.1.0** too, so this fix is still needed after #1424 (which moves the repo to 8.1.0). Please re-run `just force-document` once rebased onto #1424 to confirm zero warnings and no Rd drift under 8.1.0 specifically — the PR's verification was against 8.0.0.

## Notes
- After this and #1424 both land, the Sync Checks job should be warning-free as well as drift-free — worth confirming on the first green run.

---

</details>

PR #1403, opened 2026-08-13
fix(api): pass dimnames to row-name getters

CGMossa:
## Summary

- pass extracted `dimnames` to `Rf_GetRowNames` in both `RArray` and `List`, matching R internals and the existing column-name getters
- add public numeric-matrix and VECSXP-matrix dogfood fixtures for both dimname axes, including absent-row-name behavior
- regenerate tracked R exports/docs and record the audited failure plus verification hiccups in `findings/` and `reviews/`

## Evidence

Before the fix, both column probes returned the expected column names, `RMatrix::get_rownames` reported no names, and `List::get_rownames` returned the first integer data element. R source defines `GetRowNames(SEXP dimnames)` as operating on the already-extracted dimnames VECSXP.

## Verification

- `just fmt`
- exact CI `cargo clippy` default, full-codegen/R6, and full-codegen/S7 commands with `-D warnings`
- focused `test-rarray.R`: 15 expectations
- 50 gctorture iterations across all four public getters
- `just check`: root and both cross-package workspaces pass; final standalone rpkg leg reaches the existing duplicate-`ColumnarFrame` base blocker isolated in #1389
- `just test`: root, ndarray, and both cross-package suites pass; final standalone rpkg leg reaches the same #1389 blocker

No merge requested.

PR #1407, opened 2026-08-13
fix(api): root RArray during construction

CGMossa:
## Summary

- protect the newly allocated RArray data SEXP before installing dimensions
- retain the GC guard while the caller-provided initializer executes
- dogfood the public RMatrix constructor with an allocating initializer and record the finding/reproduction

## Evidence

Before the fix, the public 2-by-3 constructor probe failed its dimension assertion on the first gctorture iteration. After the fix, the installed R export completed 100/100 gctorture iterations with stable dimensions and values.

## Verification

- `just fmt`
- focused `test-rarray.R`: 11 expectations, 0 failures/warnings/skips
- exported constructor dogfood: 100 gctorture iterations
- `just wrappers-sync-check`
- exact CI `clippy_default`, `clippy_all`, and `clippy_all_s7` commands with `-D warnings`
- `just check`: root and both cross-package workspaces pass; standalone rpkg stops at the existing duplicate `ColumnarFrame` imports fixed by #1389
- `just test`: root and both cross-package workspaces pass; same #1389 standalone rpkg blocker

## Related

- #1389 is the independent aggregate-gate blocker
- #1404 independently removes the 14 pre-existing R6 active-binding documentation warnings observed during regeneration

No merge requested.


CGMossa:

<details>

## Verdict
LGTM. Textbook constructor-window rooting fix with first-iteration gctorture evidence.

## Review
- The window is real: `set_dims` allocates the dim INTSXP and the initializer closure is arbitrary user code — both can GC while the fresh data vector has no root, and the caller *cannot* root it because `new` hasn't returned. `OwnedProtect` across the whole construction is the right shape, and the single PROTECT/UNPROTECT pair is balanced (guard drops after the struct copy, returning to the documented caller-roots contract).
- One detail I checked: `dataptr_mut` taken before the initializer runs stays valid across initializer-triggered GC because R's collector is non-moving for standard vectors — fine, and worth remembering if ALTREP-backed construction is ever added here.
- The 2-by-3 probe with an allocation inside every initializer call is exactly the right regression, and dim-corruption-on-first-iteration is conclusive pre-fix evidence.

## Same nits as #1403 (non-blocking)
- `rarray_construct_matrix` is exported; consider `noexport` + `:::` access to keep test probes out of the exemplar package's public API (the #1396 fixtures set that precedent).
- NAMESPACE/man here were regenerated under roxygen2 8.0.0; after #1424 lands (8.1.0), rebase and re-run the document loop.

---

</details>

CGMossa (commented):

<details>

## Verdict: approve

## Summary
- `RArray::new` holds the data SEXP in an `OwnedProtect` across `set_dims` (which allocates the dim vector) and the caller's initializer; that window is the one callers could never close themselves.
- The committed test called the constructor once with no forced collection, so it also passed on main; the PR body's 100-iteration probe was the real oracle but was not in the tree. Pushed c5d31a9f: the test now loops twenty times under `gctorture(TRUE)`, gated like the other stress blocks, and `test-rarray.R` is added to GC-stress shard 1 so CI executes it.
- Verified on a combined build of the five rooting PRs with `NOT_CRAN=true`: `test-rarray.R` 55 pass, 0 fail (includes #1403's getters).

## Notes
- Merged `origin/main` and pushed. #1407 and #1403 append to `rarray_tests.rs` and `test-rarray.R` at the same point; whichever merges second needs a trivial keep-both resolution.
- Body has no attribution placeholder or details block, and no linked issue. Full notes: `reviews/1407-review.md`.

---

</details>

Issue #1413, opened 2026-08-14
build(llm-docs): honor Cargo target-dir configuration


CGMossa:
# LLM-doc generation assumes Cargo uses the repository-local `target/`

## Summary

`just llm-docs` runs every rustdoc command successfully and then fails while
rendering Markdown when Cargo is configured with a non-default target
directory:

```text
FileNotFoundError: target/doc/miniextendr_api.json
```

In the observed environment, `cargo metadata --no-deps --format-version 1`
reported `target_directory` as `/Users/elea/.cargo-target`. The generator
nevertheless hard-codes `target/doc/*.json` for the root workspace and
`cargo-revendor/target/doc/*.json` for the standalone workspace. The JSON is
created successfully in Cargo's configured target directory, then the renderer
looks in the wrong place.

## Reproduction

Configure Cargo's `build.target-dir` outside the checkout, then run:

```text
just llm-docs
```

Rustdoc completes, but `rustdoc_megadoc.py` fails on the first hard-coded
repository-local JSON path.

## Suggested fix

Make the generator own deterministic target directories for both workspaces
by setting `CARGO_TARGET_DIR` on the rustdoc commands and using those same
paths for rendering, or resolve each workspace's actual target directory from
Cargo metadata. Add a regression invocation with an external Cargo target-dir
configuration.

## Overlap audit

The existing LLM-doc reviews cover a stale sccache worktree and an absolute
cargo-revendor source path in generated inventories. Neither covers Cargo's
configured target directory. No open issue matched this failure.

CGMossa:
Duplicate of the defect PR #1399 fixes (cargo-metadata-resolved target directories + machine-independent source labels, finding dated 2026-08-13). Closing in favor of #1399's implemented approach.

CGMossa:
Duplicate of the defect PR #1399 fixes (cargo-metadata-resolved target directories + machine-independent source labels, finding dated 2026-08-13). Closing in favor of #1399's implemented approach.

CGMossa:
Duplicate of the defect PR #1399 fixes (cargo-metadata-resolved target directories + machine-independent source labels, finding dated 2026-08-13). Closing in favor of #1399's implemented approach.

PR #1414, opened 2026-08-14
fix(engine): make initialization handle unforgeable

### LINKED ISSUE #1412: safety(engine): REngine marker is publicly forgeable
# `REngine` can be forged without initializing R

## Summary

`miniextendr-engine` documents `REngine` as a marker proving that embedded R
has been initialized for the process, but declares it as a public unit struct:

```rust
pub struct REngine;
```

That exports the value constructor. Safe downstream code can write
`let engine = miniextendr_engine::REngine;` while
`r_initialized_sentinel()` is still false. The type therefore does not uphold
the invariant its API and consumers assign to it. The benchmark stores the
handle in a process-wide `OnceLock`, so the codebase already treats successful
construction as initialization evidence.

The user-facing embedding example in `docs/ENTRYPOINT.md` exposes the same
contract drift from the opposite direction: it calls
`REngine::build().unwrap()`, although `build()` only returns a builder, and
then claims runtime initialization happened automatically. The real consumer
paths call unsafe `.init()` and separately call
`miniextendr_runtime_init()` before using `miniextendr-api` FFI.

## Reproduction

This compiles without calling R:

```rust
use miniextendr_engine::{REngine, r_initialized_sentinel};

let _forged = REngine;
assert!(!r_initialized_sentinel());
```

The central `docs/ENTRYPOINT.md` example does not compile because
`REngineBuilder` has no `unwrap()` method.

## Suggested fix

- Give `REngine` a private field and construct it only after
  `setup_Rmainloop()` completes.
- Add a compile-fail doctest for direct construction.
- Turn the engine quick-start snippets from ignored examples into compiled
  `no_run` doctests.
- Correct the embedding guide to call unsafe `.init()` and, when using
  `miniextendr-api`, explicitly register the initialized thread with
  `miniextendr_runtime_init()`.

## Overlap audit

Closed issue #974 fixed an older `REngine::new()` rustdoc typo, but did not
cover public construction or the current `docs/ENTRYPOINT.md` example. Open
issue #1352 concerns off-main-thread R API escape hatches; it does not cover
the handle's construction invariant.

CGMossa:
## Summary

- make `REngine` construction private so the handle can only exist after `REngineBuilder::init()` completes R initialization
- remove the empty `Drop` implementation and the corresponding no-op `mem::forget` calls in live-R test harnesses
- compile the engine quick-starts as `no_run` doctests and add a compile-fail test for direct handle construction
- correct the public embedding guide to call unsafe `.init()` and explicitly register the owning thread with `miniextendr_runtime_init()` before checked FFI
- refresh the engine API digest and impl inventory

Closes #1412.

## Verification

- R version: `rproject.toml` pins 4.6; setup and verification used R 4.6.0
- targeted engine doctests: 4 passed, including the direct-construction compile-fail case
- `just fmt`
- all three CI-equivalent clippy legs with `-D warnings`: default, curated `full-codegen`, and curated `full-codegen-s7`
- root workspace and both cross-package legs of `just check`
- root workspace, ndarray, and both cross-package legs of `just test`
- the standalone `rpkg` check/test leg remains blocked by the pre-existing duplicate `ColumnarFrame` imports tracked in #1389
- `CARGO_TARGET_DIR=target just llm-docs`; retained only the generated engine artifacts because corpus-wide unrelated drift is already handled by #1400
- `just site-docs` and `just site-build`
- `git diff --check`

## Additional audit finding

The first corpus regeneration proved that `just llm-docs` assumes repository-local Cargo target directories. That separate build-system defect is documented and tracked in #1413; this branch used a relative `CARGO_TARGET_DIR=target` so each workspace wrote exactly where the renderer expects.

No merge requested.


CGMossa:

<details>

## Verdict
LGTM. The forgeable handle was a real invariant hole and the fix is the standard, minimal one.

## Review
- `pub struct REngine;` exporting a value constructor meant safe code could fabricate "proof" of R initialization (and the benchmark's OnceLock usage shows the codebase genuinely leans on that proof). Private field + `compile_fail` doctest is exactly right.
- Removing the empty `Drop` and the no-op `mem::forget` calls is correct: the old Drop never shut R down, so the handle's only meaning is "init completed" — the new doc line ("dropping the handle does not shut R down") states the real contract.
- The `docs/ENTRYPOINT.md` example was genuinely broken (`REngine::build().unwrap()` doesn't compile — no `unwrap` on the builder) and claimed automatic runtime registration that doesn't happen; the corrected unsafe `.init()` + explicit `miniextendr_runtime_init()` sequence matches the real consumer paths (e.g. r_test_utils).
- Converting the quick-starts from `ignore` to compiled `no_run` doctests turns future drift of this kind into a compile failure — good hardening.

## Cross-PR coordination
- The `findings/2026-08-14-llm-docs-target-dir.md` finding and issue #1413 describe the same defect PR #1399 already fixes (metadata-resolved target dirs; its finding is dated 08-13). When #1399 merges, #1413 should be closed as resolved by it, and this PR's "suggested fix" paragraph is superseded by #1399's implemented approach.
- The refreshed engine digest/impl-inventory files overlap the #1400/#1401/#1421 corpus churn; whichever lands last should regenerate. #1401's deterministic ordering will reorder these again.

---

</details>

CGMossa (commented):

<details>

## Verdict: approve

## Summary
- `REngine` gets a private field, so only a successful `REngineBuilder::init()` can produce the proof-of-initialization handle; the no-op `Drop` and the `mem::forget` calls it forced on the harnesses are gone. The quick-starts become `no_run` doctests and a `compile_fail` doctest pins direct construction. `docs/ENTRYPOINT.md` no longer shows a non-existent `.unwrap()` on the builder and now calls unsafe `.init()` plus `miniextendr_runtime_init()`.
- Merged `origin/main` (clean) and pushed a5e936a2; `cargo test -p miniextendr-engine` on the merged tree passes 3 + 3 + 1 including the compile-fail doctest, and `just llm-docs` produced no drift.

## Notes
- Body has no attribution placeholder or details block; `Closes #1412` matches. Full notes: `reviews/1414-review.md`.

---

</details>

